PR #626, opened 2015-02-03
Comments appearance

### LINKED ISSUE #574: Make entire comment block clickable on comments list

### LINKED ISSUE #613: Make 'report comment' always visible
Currently it is invisible until hover. There is no way of knowing the comment can be reported unless you hover or tap the comment and notice the appearance of the text.

This is particularly a problem for:
* mobile/tablet users, who can't hover. In my testing on an ipad I couldn't activate the hover by tapping a comment.
* keyboard users

equivalentideas:
A number of mostly subtle changes to the presentation comments on the [comments index](http://www.planningalerts.org.au/comments) and on applications.

The major visual change is to make the address/application of the comment a heading on the comments index. In particular I've highlight to suburb of the comment.

I think that the suburb is the more immediately relatable piece of information in the comment to a user visiting the comments index. They are much more likely to recognise a suburb that a street. Then they can be drawn into that comment “I know that place, what's happening over there?”. I think this brings focus to local communities.

It also makes the report link more accessible by making it always visible.

## Before

Index page:
screen shot 2015-02-03 at 11 31 17 am

Application page:
screen shot 2015-02-03 at 11 31 30 am

With hover:
screen shot 2015-02-03 at 11 31 39 am

## After

Index page
screen shot 2015-02-03 at 11 50 00 am
screen shot 2015-02-03 at 11 29 56 am

Application page
screen shot 2015-02-03 at 11 30 32 am
screen shot 2015-02-03 at 11 30 10 am

## TODO — fixed by https://github.com/openaustralia/planningalerts-app/commit/634868ce9cb409fb3531a3e274a7cc2eb10c93b1

In the rspec tests for comments, it includes a test for the presentation of Time.now (see https://github.com/equivalentideas/planningalerts-app/blob/236a1e28fe90af03c6cfc5d9cd6b13bf2bfb80a8/spec/views/comments/comment_spec.rb#L6 ) . Because this changes the time notice to use the time element and include an attribute with a machine readable version of the time, I wasn't sure how that could be tested—how can it predict what the current time will be? I wasn't sure how to do that, so I've left it as a todo. The tests pass today because the Time.now happens to be the same as in the test markup, but it should fail tomorrow.

One option would be to remove that aspect that the test, as it isn't the specific focus of the test e.g. “should remove links in the comment text”. Direction needed here (ping @henare @mlandauer).

closes #574 
closes #613


mlandauer:
Looks good to me.

As soon as you've added tests for the time this looks good to merge. There are a bunch of gems that help with testing time stuff. See https://www.ruby-toolbox.com/categories/time_warping

It looks like the most popular is timecop which I think i've used and is all pretty easy to use.

equivalentideas:
> It looks like the most popular is timecop which I think i've used and is all pretty easy to use.

That looks perfect, thanks for that.

equivalentideas:
Fixed comments time tests with https://github.com/openaustralia/planningalerts-app/commit/634868ce9cb409fb3531a3e274a7cc2eb10c93b1.

equivalentideas:
Ouch. Travis fail.

```
-<time class='comment-time' datetime='2015-01-26'>
+<time class='comment-time' datetime='2015-01-25'>
```

I think this is a problem with the way I setting the time using `Time.parse('2015-01-26 08:32:34 +1100')`, the time code somehow messed it up. I am a noob at times.

equivalentideas:
fixing now with `Timecop.freeze(Time.local(2015, 1, 26, 10, 5, 0))`

equivalentideas:
passing, victory! Could someone please merge this?

mlandauer:
Just before this gets merged @equivalentideas have you checked that it doesn't break the other theme?

equivalentideas:
yikes. No I haven't, had no idea about that. How do I swap between themes? Also, do the tests run for both themes automatically?

mlandauer:
Point your browser at nsw.127.0.0.1.xip.io:3000

You'll see in `app/models/configuration.rb` there's a `THEME_NSW_HOST` - if the browser sees that domain it will show the other theme.

The theming only affects the views so it doesn't matter for tests unless they're view tests (of which there are very few).

equivalentideas:
Thanks for that. Works well on that them as well :)

 Maybe we should add something to the readme?

mlandauer:
Sounds good!

equivalentideas:
Awesome I'll make a ticket. Assigning to you to merg @mlandauer

PR #668, opened 2015-05-06
Bulk date API

henare:
This adds the ability for bulk API users to get all applications created in PlanningAlerts for a particular day.

Whilst I created this for a specific project I can see it being used in other ways.

It could be used by commercial users that want all applications but not an entire copy of the PlanningAlerts database (which is what our current bulk API provides). Or for commercial users that only want a full copy of more recent applications without the years of history we have. With this API they can just make an API call a day to keep up to date with new applications.

It could also be used by people interested in doing ad hoc research, e.g. if you want a whole month or year of applications this API method would be the best way for you to do that.

Assigning to @mlandauer to review the concept, and a brief code review if he feels like it.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -47,9 +47,13 @@ class ApiController < ApplicationController
   def date_scraped
     begin
       date = Date.parse(params[:date_scraped])
-      api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
     rescue ArgumentError => e
       raise e unless e.message == "invalid date"
+    end
+
+    if date
+      api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
+    else
       respond_to do |format|
         format.js { render json: {error: "invalid date_scraped"}, status: 401 }
       end
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -45,8 +45,15 @@ class ApiController < ApplicationController
   end
 
   def date_scraped
-    date = Date.new(2015, 05, 06) # TODO: Parse date from param
-    api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
+    begin
+      date = Date.parse(params[:date_scraped])
+      api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
+    rescue ArgumentError => e
+      raise e unless e.message == "invalid date"
+      respond_to do |format|
+        format.js { render json: {error: "invalid date_scraped"}, status: 401 }
+      end
+    end
   end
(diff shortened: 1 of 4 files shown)


mlandauer:
Looking at this now

app/controllers/api_controller.rb
```diff
@@ -43,36 +44,40 @@ def area
       "Recent applications in the area (#{lat0},#{lng0}) (#{lat1},#{lng1})")
   end
 
+  def date_scraped
+    begin
+      date = Date.parse(params[:date_scraped])
+      api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
+    rescue ArgumentError => e
```
mlandauer: Could this rescue block just be around the Date.parse so that it doesn't accidentally catch anything from anywhere else?

app/controllers/api_controller.rb
```diff
@@ -123,6 +128,16 @@ def check_api_parameters
     end
   end
 
+  def authenticate_bulk_api
```
mlandauer: Awesome extracting this into a filter!

app/controllers/api_controller.rb
```diff
@@ -43,36 +44,40 @@ def area
       "Recent applications in the area (#{lat0},#{lng0}) (#{lat1},#{lng1})")
   end
 
+  def date_scraped
+    begin
```
mlandauer: Should there be a call to `ApiStatistic.log(request)` here?

henare: It gets called as part of `#api_render`.

app/controllers/api_controller.rb
henare: You're right - I will change this. I just did this because of a problem when I put it around just `Date.parse` at first.

app/controllers/api_controller.rb
henare: Thanks :smile:

app/controllers/api_controller.rb
```diff
@@ -43,36 +44,44 @@ def area
...
+      date = Date.parse(params[:date_scraped])
+    rescue ArgumentError => e
+      raise e unless e.message == "invalid date"
+    end
+
+    if date
+      api_render(Application.where(date_scraped: date.beginning_of_day...date.end_of_day), "All applications collected on #{date}")
+    else
+      respond_to do |format|
+        format.js { render json: {error: "invalid date_scraped"}, status: 401 }
```
henare: Ahem. I think someone copied this line - the status code doesn't look right to me.

henare:
Thanks heaps for the code review @mlandauer. It's always very helpful.

Does your review also mean you conceptually agree with merging this? If so I think it's ready to go.

mlandauer:
Yeah cool. I'll just take one more quick look over things now.

mlandauer:
LGTM. Thanks!

henare:
Thank you :bear:

PR #671, opened 2015-05-08
Upgrade to Rails 4.1

henare:


henare:
TODO:

- [ ] Fix header container off to the left
- [ ] Fix squished maps/streetview container
- [ ] Fix very large more info button
- [ ] Test scraping and alerting

selection_001

henare:
@equivalentideas it would be good to get you to look at these layout bugs if you have some time.

mlandauer:
Why is upgrading rails causing layout changes? That seems mighty weird.

equivalentideas:
That does seem weird. Any ideas @henare ?

henare:
I'm pretty sure it's to do with the required upgrades of `compass`, `compass-blueprint`, and `susy`. Since I don't understand how that all hangs together I've likely stuffed something up.

Related commits: f445ff2c8d3dd7a03765045c8e35b989e7c613ec 4ce8e99b4eff7c586de632d4330ee2338a014743 59e7e20dc1e36e2dc7e170781ce7b1c9fe449b17 6067c39ad78ae3a4fb1e7a09d207a42d6d67fa39 a2a0d810d83d86fe6f2abb387cd5d272827d1cf7 and maybe more.

henare:
@mlandauer the current `master` branch refers to [Blueprint](https://github.com/openaustralia/planningalerts/blob/0f48cf07959dd7afe9bda4027644cb03acb5c5c3/app/assets/stylesheets/screen.scss#L2) but I can't for the life of me work out how this is installed/accessed. It's not in the Gemfile and even grepping through the code I can't find any other instances of it being referenced. Help!

henare:
When deploying, don't forget there are new configuration settings: https://github.com/openaustralia/planningalerts/pull/671/files#diff-645114e68f0da5ba4d02a2928efd7ed7R59

henare:
Yaaay :tada: I've tested everything on my list and it's good to go.

henare:
A couple more issues:

- [ ] User signup confirmation doesn't work
- [ ] API returning all results for me, not filtered by address or lat/long when requested

henare:
I've deployed this branch to staging with `cap deploy -S stage=test -S branch=rails4`

@openaustralia/team-planningalerts anyone care to see if they can spot any problems? https://test.planningalerts.org.au/ Because we force SSL you now need to click past the certificate warning.

equivalentideas:
Might be expected, but:

* There's no 'Number of applications scraped each week' chart on the authority page https://test.planningalerts.org.au/authorities/albury vs https://www.planningalerts.org.au/authorities/albury
* no 'or locate me automatically' link under search form

henare:
:star: :star: :star: :star: :star: bugs, thanks @equivalentideas

henare:
Those two bugs are now fixed, thanks @equivalentideas!

henare:
Assigned to @mlandauer for sign-off :pencil2:

henare:
I normally wouldn't skip versions but Rails 4.0 isn't supported and the upgrade to Rails 4.1 isn't a huge one so I've gone ahead and added it to this pull request.

Gemfile
```diff
@@ -45,7 +50,8 @@ gem 'delayed_job_active_record'
 gem 'daemons'
 gem "validate_url"
 gem "twitter"
-gem "atdis"
+# Can revert to plain gem when this PR is merged https://github.com/openaustralia/atdis/pull/39
+gem "atdis", github: "openaustralia/atdis", branch: "activemodel4"
```
mlandauer: @henare I’ve released a new version of the atdis gem with your PR merged in. Could you fix the hack here?

henare: :rainbow:

app/models/configuration.rb
```diff
@@ -55,4 +55,8 @@ module Configuration
   # for subscriptions. See app/controllers/subscriptions_controller.rb
   STRIPE_PUBLISHABLE_KEY = 'xxxxxxxxxxxxxxxxx'
   STRIPE_SECRET_KEY = 'xxxxxxxxxxxxxxxxx'
+
+  DEVISE_SECRET_KEY = "9c92ee433b851b3db3952e81e0c9d678de778b194986e2ae8b160eca08bea4696152801e83d88b75a17375867b07e46d70d9bfb7ff5a352c2e1de8790448eaa5"
+
+  SECRET_KEY_BASE = "39d782b099a06652f92585de84f568c71885a4a129a307f03d2a803c6298ea129e334442ab042dc7fa664adc6e28b41317c3d53291366d0d7b5761c65102f3d4"
```
mlandauer: Should this be here??

henare: Yes. In the nasty way we have configuration set up in production this file is overwritten completely with the real values: https://github.com/openaustralia/planningalerts/blob/master/config/deploy.rb#L47

Why use what look like real values, you ask? Well we had them committed before so I thought, meh, why not keep the same keys in development (e.g. https://github.com/openaustralia/planningalerts/pull/671/files#diff-afe98573d9aad940bb0f531ea55734f8L7 )

mlandauer: Ah okay. Cool. I think it would be good for those values not to look like real values just so that the next person that comes to look at those lines doesn't react the same way as me ;-)

mlandauer:
I've skimmed through the PR and nothing else jumped out at me so I think we can say that this is good merge. What do you say to doing the merge first thing Monday morning followed by a deploy?

It feels too risky to do it now.

henare:
Thanks so much for checking over it.

I agree. If I'm working tomorrow morning I might do it then as it would be even less risky but otherwise Monday sounds like a plan (get it, plan!).

Issue #712, opened 2015-08-11
Subscription activated page needs some work


henare:
![selection_001](https://cloud.githubusercontent.com/assets/48945/9189696/88cfd724-402f-11e5-9654-c20a526ad4e9.png)

* There's no space after the first sentence.
* The sparse page looks a little out of keeping with the rest of the site. We don't normally use black text for headings, for instance.
* The last paragraph has a hanging word, which I know totally Grinds @equivalentideas' Gears (hey, that rhymes).
* It doesn't tell you the wonderful thing that happens (i.e. that you've unlocked unlimited alerts or that your alerts will start working again now)

henare:
The "contact us" link goes to an email. Maybe we should change it to say "email us".

I say this because often "contact us" links go to a contact page whereas this will open your email client which may be unexpected.

If we change it to email us we probably should also include the email address since so many people use webmail these days and are probably used to copy and pasting text (but not links).

equivalentideas:
> If we change it to email us we probably should also include the email address since so many people use webmail these days and are probably used to copy and pasting text (but not links).

Sounds good :+1:

PR #717, opened 2015-08-11
New subscription page

### LINKED ISSUE #708: When I want to activation my subscription from the trial banner, take me to page where I can subscribe
We need a page at /subscriptions/new where people can subscribe.

equivalentideas:
This implements a first pass on the new subscription page that users see when they click the link in the subscriptions banner when they are on a trial subscription.

These commits:

* replace the page copy with copy appropriate to the new system we're making
* strips back the styles to better match the site generally

screen shot 2015-08-12 at 3 20 34 pm

Any feedback/thoughts at this stage @henare ? I'll assign to you to merge or respond.

closes #708 

^^^ Close that issue and open new ones for continued iteration like #712

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -15,6 +15,7 @@ class SubscriptionsController < ApplicationController
 
     if params[:email]
       @email = params[:email]
+      @subscription = Subscription.find_by(email: @email)
     end
   end
 
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -15,7 +15,6 @@ class SubscriptionsController < ApplicationController
 
     if params[:email]
       @email = params[:email]
-      @subscription = Subscription.find_by(email: @email)
     end
   end
 
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -1,7 +1,6 @@
 .registration-page,
 .registration-success-page {
   padding-bottom: 2em;
-  text-align: center;
 
   @include at-breakpoint(40em) {
     padding-bottom: 4em;
@@ -26,20 +25,6 @@
     @include at-breakpoint(40em) {
       font-size: 1.3em;
     }
-  }
-}
-
-.registration-page {
-  @include at-breakpoint(40em) {
-    @include span-columns(10, 12);
-    margin: 2em auto 0;
-    float: none;
-  }
-
-  @include at-breakpoint(60em) {
-    @include span-columns(8, 12);
-    margin: 2em auto 0;
-    float: none;
   }
 }
 
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -12,9 +12,7 @@
 }
 
 .registration-intro {
-  @include at-breakpoint(53em) {
-    width: 60%;
-  }
+  max-width: 42em;
 
   p {
     @include at-breakpoint(40em) {
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -1,13 +1,13 @@
 .registration-page,
 .registration-success-page {
-  padding-bottom: 2em;
+  padding-bottom: 4em;
 
   @include at-breakpoint(40em) {
-    padding-bottom: 4em;
+    padding-bottom: 8em;
   }
 
   @include at-breakpoint(60em) {
-    padding-bottom: 6em;
+    padding-bottom: 9em;
   }
 }
 
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -9,14 +9,6 @@
   @include at-breakpoint(60em) {
     padding-bottom: 6em;
   }
-
-  p {
-    font-size: 1.1em;
-
-    @include at-breakpoint(40em) {
-      font-size: 1.3em;
-    }
-  }
 }
 
 .registration-success-page {
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -36,8 +36,7 @@
 }
 
 p.form-processing {
-  margin: 0 auto 1em;
-  padding: 1em 0 .7em;
+  margin: 0 0 .2em;
   font-size: 1.25em;
   font-weight: bold;
 }
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -10,15 +10,6 @@
     padding-bottom: 6em;
   }
 
-  h1 {
-    font-size: 2em;
-    line-height: 1.2;
-
-    @include at-breakpoint(40em) {
-      font-size: 2.5em;
-    }
-  }
-
   p {
     font-size: 1.1em;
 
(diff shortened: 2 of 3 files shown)


henare:
The layout is nice and clean @equivalentideas.

The title doesn't describe what this page is about, i.e. it's not about getting planning alerts for all the locations you need, it's about signing up to a subscription. But I don't mean we should put "sign up for a subscription" as the title either :)

The remaining copy also needs work which is evident if you read it out loud. This copy must be super clear about why you're here and what you should do next. It should also be in the voice of the rest of the site - simple, friendly, helpful.

Can you please also change the HTML title?

equivalentideas:
> The title doesn't describe what this page is about, i.e. it's not about getting planning alerts for all the locations you need, it's about signing up to a subscription. But I don't mean we should put "sign up for a subscription" as the title either :)

I think you're right, it’s too focused on the value of what you get, where as it should be more about doing this thing to get your alerts working again. Fixing the block in your alerts, rather than getting something new.

**"Activate your subscription"** might be a better title as you've come from a banner or prompt when trying to view your alerts. You're thinking 'Alright, what have I got to do to get this happening?' and there's then answer. It also reinforces that it's currently inactive or in trial, you have to do something to fix it.

henare:
>"Activate your subscription" might be a better title

Sounds good. It would also lead nicely into the text that follows because you'd then be able to explain why, etc.

equivalentideas:
I think it's flowing better, and is much less passive aggressive with:

> Activate your subscription
> You need an active subscription to receive planning alerts for multiple street addresses or for commercial purposes.

Friendlier, clearer and much more direct I think. Their might be a better phrase for 'receive planning alerts for multiple street addresses' but it's not coming to me right now.

screen shot 2015-08-11 at 4 48 25 pm

equivalentideas:
I've just adjusted the layout to make it a bit nicer and for the shape of the text to put more focus on the button on wide screens. I'm assigning back to you @henare for any more comments or to merge.

screen shot 2015-08-11 at 5 08 05 pm

henare:
@equivalentideas to me it looks like the remaining copy has just been rearranged and it doesn't address the feedback I gave earlier:

>...This copy must be super clear about why you're here and what you should do next. It should also be in the voice of the rest of the site - simple, friendly, helpful.

Please let me know if I need to clarify any of that.

equivalentideas:
Thanks @henare , I think I do need you to clarify :)

> to me it looks like the remaining copy has just been rearranged and it doesn't address the feedback I gave earlier.

So, initially it was:

> While personal use of PlanningAlerts is free tracking multiple locations, or using it for commercial purposes, requires an active account.

Now it's:

> You need an active subscription to receive planning alerts for multiple street addresses or for commercial purposes.

With 'Activate your subscription' as the heading.

As I said above: "Friendlier, clearer and much more direct I think."

Do you mean to say it's not simple, friendly or in the voice of the rest of the site?

henare:
>I think I do need you to clarify :)

Okey dokey :)

>So, initially the paragraph was:...
>Now it's:...

Which is largely just a rearrangement of what was there.

>Do you mean to say it's not simple, friendly or in the voice of the rest of the site?

By itself maybe but when you put it in this context...

>This copy must be super clear about why you're here and what you should do next.

...you start to see that maybe it's not. For example "You need an active subscription" is passive and obscure when in fact you're on a page where you can get an active subscription!

See how you go addressing all of that earlier feedback and let me know if you're still stuck.

equivalentideas:
Putting in the new wording from the google doc:

> Subscribe here
> Subscribe here to keep getting alerts of planning applications near your 3 addresses. Once you’re subscribed you can sign up for alerts from as many locations as you need and use PlanningAlerts in your work.

I realised that we need an email to work out the number of addresses in that first sentence. What if the app doesn't know your email when you hit this page? Any thoughts on this @henare ? What about if we have no email:

> Subscribe here if you need planning applications near several addresses.

and if we do, the:

> Subscribe here to keep getting alerts of planning applications near your 3 addresses.

equivalentideas:
Also, if you hit this page and we have your email but you don't have alerts for 3 or more addresses, we shouldn't say 'Subscribe here to keep getting alerts of planning applications near you 1 address', we should go to the 'Subscribe here if you need alerts for planning applications near several addresses' situation.

henare:
@equivalentideas I don't think the sentence should have a number in it. Also remember it has to make sense for people currently in a trial or maybe not.

equivalentideas:
>  I don't think the sentence should have a number in it.

Yeah I agree, it's less robotic as “Subscribe here to keep getting alerts of planning applications near several addresses.”

> remember it has to make sense for people currently in a trial or maybe not.

I've just written some alternatives for the different situations I can think of in https://docs.google.com/document/d/1HVyGjueO0K_6mtxhADZgqlrNKxpJFzlFy27oQn3hOWo/edit could you have a squiz at that and see if I've missed anything?

henare:
>I've just written some alternatives for the different situations I can think of in https://docs.google.com/document/d/1HVyGjueO0K_6mtxhADZgqlrNKxpJFzlFy27oQn3hOWo/edit @henare could you have a squiz at that and see if I've missed anything?

Do we *really need* 3 different bits of text? I was expecting one bit of text that didn't sound awkward no matter what situation you were in.

equivalentideas:
> Do we really need 3 different bits of text? I was expecting one bit of text that didn't sound awkward no matter what situation you were in.

@henare Yeah it feels like optimisation too early to me. I think I should remove that and give everybody:

> Subscribe here to get alerts of planning applications near several addresses. Once you’re subscribed you can get alerts from as many locations as you need and use PlanningAlerts in your work.

That works whether you have a trial subscription, an expired subscription or no subscription.

I think the extra par if you're in a trial is good cause it reminds you that you may as well subscribe now if the alerts are valuable to you, cause you'll have to do it in a few days anyway.

I've also tweaked the layout.

## Here's what it looks like with a trial subscription
screen shot 2015-08-12 at 3 20 34 pm

equivalentideas:
@henare I've just assigned this back to you for more feedback please :)

app/controllers/subscriptions_controller.rb
```diff
@@ -15,6 +15,7 @@ def new
 
     if params[:email]
       @email = params[:email]
+      @subscription = Subscription.find_by(email: @email)
```
henare: With just someone's email address I can now see if they have a trial and how many days left on that trial. It's not exactly a huge information leak but we still shouldn't do it.

henare:
@equivalentideas once you've fixed up that security issue I think we can merge this. I'll then add issues for some outstanding things we need to address.

app/controllers/subscriptions_controller.rb
equivalentideas: Ah I didn't think of that :+1:

People will most likely be coming from an alert email if they see that, would it be ok to use the alert.confirm_id as a way to hide this, as we do for the unsubscribe and area actions on alerts? Then you need the person's email and an email that's been sent to them.

henare: Yes, good thinking, you could use that token.

However it's worth asking ourselves if we really need this element. As you just said, "People will most likely be coming from an alert email if they see that", so why show it again?

equivalentideas: > However it's worth asking ourselves if we really need this element. As you just said, "People will most likely be coming from an alert email if they see that", so why show it again?

Hmmm. I do think it's helpful to have it there on the page with the information about what the subscription does for you—but, again it feels like over optimising at this point, I'm sure we'll make another changes as we iterate and we add it if it feels necessary.

I’m going to cut it.

equivalentideas:
Thanks for that @henare I've removed the trial days count completely with 3e08657 . I'm happy for you to merge this.

henare:
Merged into master e3771dfd92c2672e0bddd1df9eb0fc5341e92f25

PR #732, opened 2015-08-14
Alert confirmation page for trial subscribers

### LINKED ISSUE #711: Subscription information on alert activation page is confusing
When you subscribe for your 3rd alert you see this page:

selection_001

The two biggest issues I can see are:

* The box says "your alert has been activated" but it also contains a button "Subscribe now $49/month". Which one is it, have I subscribed or not?
* The text explaining stuff about subscriptions doesn't clearly state what's happened, why it's happening, what will happen (i.e. your trial will end), and what you need to do about it

equivalentideas:
This improves the text on the alert confirmation page for people on their trial subscription. It makes it friendlier, much more direct and gives more of an explanation of what they need to do.

It also simplifies the page by removing the radius adjustment link and the link to subscribe to more alerts.

It also removes the 'Subscribe now' button and replaces it with a link to the new subscription page to give more focus to the information on the page, rather than the new action.

## Before
screen shot 2015-08-14 at 3 51 16 pm
screen shot 2015-08-14 at 3 51 05 pm

## Now
screen shot 2015-08-14 at 3 50 42 pm
screen shot 2015-08-14 at 3 50 22 pm

closes #711

### DIFF
--- a/app/controllers/alerts_controller.rb
+++ b/app/controllers/alerts_controller.rb
@@ -22,7 +22,17 @@ class AlertsController < ApplicationController
   def confirmed
     @alert = Alert.find_by!(confirm_id: params[:id])
     @alert.confirm!
-    @alert.create_subscription_if_required
+
+    if Subscription::FEATURE_ENABLED && @alert.email_has_several_other_alerts?
+      if @alert.subscription.nil?
+        @subscription = Subscription.create!(email: @alert.email, trial_started_at: Date.today)
+        @new_subscription = true
+        @alert.reload
+      else
+        @subscription = @alert.subscription
+        @new_subscription = false
+      end
+    end
   end
 
   def unsubscribe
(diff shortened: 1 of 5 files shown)


equivalentideas:
Assigning to you @henare for feedback etc.

henare:
@equivalentideas this is excellent - such a big improvement :balloon: 

Your changes also spurred me on to a bit of a refactor of some nasty things I'd done so I've added them here :point_up:

equivalentideas:
> Your changes also spurred me on to a bit of a refactor of some nasty things I'd done so I've added them here :point_up:

Very nice. I had one question about flipping the order in the logic.

We could assign a better name to the question like `if @not_a_trial_subscriber` and flip it back so the most common case, the user isn't a subscriber, comes first.

henare:
>We could assign a better name to the question like if @not_a_trial_subscriber and flip it back so the most common case, the user isn't a subscriber, comes first.

I've not come across that convention and I don't think it's important here - what improvement would we get?

equivalentideas:
Ah I remembered it wrong anyway, it was the more specific test comes first https://github.com/openaustralia/morph/commit/624eed36e60f58134ca88a5795a177d84105a88a

So we're right here then, it's more specific that someone will be a trial subscriber that not :+1:

PR #734, opened 2015-08-17
Note to paid subscribers in alert email

### LINKED ISSUE #713: Add something to the alerts of subscribers
A thank you and something about how to unsubscribe?

equivalentideas:
This adds a note for subscribers, reminding them that they're paying and that they can sign up for alerts from as many addresses as they want. It also let's them know how they can cancel their subscription.

closes #713 

screen shot 2015-08-17 at 10 32 20 am

The's also a refactor of the alert notifier view spec.

### DIFF
--- a/app/views/alert_notifier/_footer_main.html.haml
+++ b/app/views/alert_notifier/_footer_main.html.haml
@@ -1,3 +1,11 @@
+- if @alert.subscription.try(:paid?)
+  %p
+    You’re a paid subscriber so feel free to
+    #{link_to "sign up for alerts from as many locations as you need", new_alert_path(email: @alert.email)}.
+  %p
+    You can cancel your subscription at any time, just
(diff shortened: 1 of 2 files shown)


henare:
Looks good.

I wonder if we need to clarify the final sentence now? It mentions unsubscribing to "stop receiving these emails" - what emails, this one or all of them?

equivalentideas:
> I wonder if we need to clarify the final sentence now? It mentions unsubscribing to "stop receiving these emails" - what emails, this one or all of them?

Nice, I had a similar thought brewing.

For the majority who are subscribed to one alert, I like the simplicity of the sentence “unsubscribe and stop receiving these emails” (thought maybe it should be “to” instead of “and”). But for the minority receiving two or more alerts, it could be confusing, as you say “what emails, this one or all of them?”

What about:

> unsubscribe to stop receiving alerts for this address.

That makes sense for one or several alerts.

henare:
Bellissimo :ok_hand:

PR #737, opened 2015-08-17
Unsubscribe action wording

equivalentideas:
This changes the wording of the unsubscribe link in alert emails to make it clearer what you are unsubscribing from. The idea for this change came from the discussion here https://github.com/openaustralia/planningalerts/pull/734#issuecomment-131649057

The wording changes from:

> [unsubscribe](#) and stop receiving these emails

To:

> [unsubscribe to stop receiving alerts for this address](#)

screen shot 2015-08-17 at 12 46 42 pm
screen shot 2015-08-17 at 12 46 27 pm

### DIFF
--- a/app/views/alert_notifier/_footer_actions.html.haml
+++ b/app/views/alert_notifier/_footer_actions.html.haml
@@ -2,5 +2,4 @@
   You can
   = link_to 'change the size', area_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id)
   of the area covered by this alert or
-  = link_to 'unsubscribe', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id)
-  to stop receiving alerts for this address.
+  #{ link_to 'unsubscribe to stop receiving alerts for this address', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id) }.
--- a/app/views/alert_notifier/_footer_actions.html.haml
+++ b/app/views/alert_notifier/_footer_actions.html.haml
@@ -1,5 +1,5 @@
 %p
   You can
   = link_to 'change the size', area_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id)
-  of the area covered by this alert or
-  #{ link_to 'unsubscribe to stop receiving alerts for this address', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id) }.
+  of the area you’re interested in or
+  #{ link_to 'stop receiving alerts for this address', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id) }.
--- a/app/views/alert_notifier/_footer_actions.html.haml
+++ b/app/views/alert_notifier/_footer_actions.html.haml
@@ -2,4 +2,4 @@
   You can
   = link_to 'change the size', area_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id)
   of the area you’re interested in or
-  #{ link_to 'stop receiving alerts for this address', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id) }.
+  #{ link_to 'unsubscribe from alerts for this address', unsubscribe_alert_url(protocol: @protocol, host: @host, id: @alert.confirm_id) }.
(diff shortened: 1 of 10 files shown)


henare:
Hmm, now I look at this sentence overall it feels disjointed:

>You can change the size of the area covered by **this alert** or unsubscribe to stop receiving alerts for **this address**.

In the original sentence we already referred to "this alert" earlier in the sentence but the reason it says "these emails" later is to help the person understand that it will stop these emails if you unsubscribe. So the following could work and it reinforces the concept of "an alert":

>You can change the size of the area covered by this alert or unsubscribe and stop receiving this email alert.

What do you think @equivalentideas?

henare:
Maybe "disjointed" is not the right word? Just confusing because it could be taking about two separate things.

equivalentideas:
I think we're picking up a bit of the confusion of the duel meaning of “alert” in the app.

Sometimes we refer to “alerts” as the individual emails notifying you of new development applications:

> Find out for free what is being built and knocked down in your area and receive email alerts when something new is planned.

> Subscribe here to get alerts of planning applications near several addresses. 

> you received 72,791 email alerts

And sometimes we use “alert” to refer to a thing sends you email notifications for DAs around an address:

> You can change the size of the area covered by this alert

> you signed up for 308 new email alerts

I think the former makes more sense and we should try and avoid the later. The former feels more natural to more, which is probably why we use it more and in more prominent places.

So in this case, what about something like:

> You can [change the size](#) of the area you get alerts for or [unsubscribe to stop receiving alerts around this address](#).

Does this make any sense? Wording is confusing my brain right now.

equivalentideas:
We also say 'create alert' on the search results page. Maybe 'Get alerts' or something would be better. It'd be interesting to AB test this, but it feels like something bigger than we should look into right now and possibly a big yack.

Ima make a new issue to discuss this at some point and we should just try and come up with something clear for this text now.

henare:
>Ima make a new issue to discuss this at some point and we should just try and come up with something clear for this text now.

Good call.

This is what I was writing as you made that comment:

>Yes! I think you're heading down the right track with identifying the problem.
>
>However If we adopt the approach suggested above what do we call an alert? In natural speech I say I have 2 alerts signed up.
>
>It would be interesting to do a bit of research to see what other people use.

equivalentideas:
Been having a look through the survey and people keep saying things like:

> tracks development happening in my area
> helps me keep track of all the developments
> It lets me know about the building activity
> Good to know what's happening in the area
> I want to know whats going on

So, why not skip the abstract concept of an 'alert' here and describe what people are actually doing:

> You can [change the size of the area you’d like to keep track of](#) or [stop receiving alerts for development near 48 Gowrie St, Newtown](#) whenever you’d like.

I quite like something like that as a direction, much less computer-speaky.

henare:
Excellent, I like where that's going :+1: I think the proposed sentence needs a little more refinement for clarity and simplicity.

equivalentideas:
I've changed this to a simplified version of above:

screen shot 2015-08-17 at 5 02 04 pm

henare:
I think it should still use the "unsubscribe" keyword - people are used to scanning for that at the end of an email.

equivalentideas:
like “or unsubscribe from alerts for this address”?

henare:
Yeah - have we come full circle yet? ;)

equivalentideas:
> Yeah - have we come full circle yet? ;)

Lol almost, but it's just more human focused now ;)

equivalentideas:
I've made that change to replace 'stop receiving' with 'unsubscribe from'.

henare:
I was expecting you'd add "unsubscribe" rather than replace what was there, similar to your suggestion in an earlier iteration:

>[unsubscribe] to stop receiving alerts for this address.

That way you have the unsubscribe keyword and you also get an explanation of what will happen if you unsubscribe. What do you think?

equivalentideas:
> That way you have the unsubscribe keyword and you also get an explanation of what will happen if you unsubscribe. What do you think?

I think we can drop the 'stop receiving' if we have 'unsubscribe' because people know what'll happen if they unsubscribe, which is why they go looking for the word. I think the important change for the unsubscribe part of the sentence was https://github.com/openaustralia/planningalerts/commit/28aecd0470c3d3abf50e46233e5c9936dee99239 : making it clear what you're unsubscribing from.

henare:
Okay cool. I was just trying to read the intent of the original wording to make sure we didn't introduce wording regressions :)

equivalentideas:
> Okay cool. I was just trying to read the intent of the original wording to make sure we didn't introduce wording regressions :)

:+1: :+1: It's great that we're making all the wording super tight.

PR #739, opened 2015-08-17
Alert trial expiry

### LINKED ISSUE #698: As a commercial user, when my trial period has finished, keep sending my alerts but don't show the applications, to remind me to subscribe

henare:
Doesn't show the applications in an alert email if the subscription has expired.

Fixes #698.


henare:
This functionally works. We just need teh werds: https://github.com/openaustralia/planningalerts/pull/739/files#diff-2a80d0e5a4655f4a15cbd11904edc4b9R1

equivalentideas:
@henare you can assign this over to me if you'd like me to werk teh werds.

equivalentideas:
I'm gonna pick this up and do a pass on the wording.

equivalentideas:
Here's an idea to keep the trial banner, which they should now be familiar with, to tell them that their trial has expired. Then we use a slight variation on the text from the subscription page to explain what they need to do to continue getting alerts for several addresses. I think being consistent through this process is gonna help people understand what's going on.

I think the subject of the email should still be "12 new planning applications near 6-475 Fake Street" to tell people what they're missing.

screen shot 2015-08-21 at 4 27 05 pm

henare:
@equivalentideas I like the idea of reusing the text.

I think we should have a big header above that text though. Something along the lines of "Oh no! Your trial has expired".

I don't think the link needs to include "for several addresses". It's punchier with just "Subscribe now to continue getting alerts".

equivalentideas:
Just merged master into this long running branch.

equivalentideas:
@henare I'm trying to go with something more positive than 'Oh no!', though I do like the casualness of that:

screen shot 2015-08-26 at 2 27 11 pm
 
Could you give me a quick response to that heading?

henare:
>I'm trying to go with something more positive than 'Oh no!'

@equivalentideas I personally think we need to be straight up with the fact that not getting alerts any more is a negative thing. But I'm happy to give your suggestion a try - we can always change it!

equivalentideas:
> @equivalentideas I personally think we need to be straight up with the fact that not getting alerts any more is a negative thing. But I'm happy to give your suggestion a try - we can always change it!

Thanks, that sounds good :+1: I feel like we want them focus on the value and be happy about stepping into a subscription. But as you say, let's give it a go and see what sticks :snowboarder:

equivalentideas:
Hey @henare I've update the tests and the text versions, so it should be easier to keep the text version up to date now with tests.

Assigning back to you.

henare:
>I've update the tests and the text versions, so it should be easier to keep the text version up to date now with tests.

Cool!

>Assigning back to you.

Looking good :shipit:

PR #742, opened 2015-08-17
Make the trial banner less forceful

### LINKED ISSUE #736: Email subscribe banner is missing something
I think the problem is that it's telling me something and I might not understand that but the only available action is to "subscribe now". Maybe it's as simple as putting a link on "trial subscription"?

equivalentideas:
These commits:

1. add a link to 'trial subscription' so it doesn't seem like the only option is to subscribe right now
2. make the button text simply 'subscribe' instead of 'subscribe now' to make it less commanding/rude.

This is an attempt to fix #736

screen shot 2015-08-17 at 5 58 21 pm
screen shot 2015-08-17 at 5 58 15 pm


equivalentideas:
Sloppy me forgot to update the tests.

henare:
Hmm, I'm not sure about the button change. My initial reaction was that it now seems to be part of the content, i.e. here's your trial alert, click here to subscribe permanently.

I'll merge and deploy this and we can see how we feel.

equivalentideas:
> Hmm, I'm not sure about the button change. My initial reaction was that it now seems to be part of the content, i.e. here's your trial alert, click here to subscribe permanently.

I see what you mean.

> I'll merge and deploy this and we can see how we feel.

Sounds good :+1:

PR #743, opened 2015-08-18
Add growl gem for better guard notifications

equivalentideas:
See https://github.com/guard/guard/wiki/System-notifications#growl this is something I found useful working on morph.io.


henare:
Bah. Just Works on Linux ;)

PR #744, opened 2015-08-18
Provide more and clearer information on the new subscription page

### LINKED ISSUE #721: New subscription page doesn't say you only need a subscription for commercial, high-volume use

### LINKED ISSUE #722: New subscription page doesn't tell you what to do if you're a non-commerical user
Also, it should emphasise that PlanningAlerts is free for low-volume, non-commercial use.

equivalentideas:
These commits change the text on the new subscription page to make it clearer that you need a subscription for commercial use or to get alerts for several addresses, and also let you know what to do if you're a non-commercial user. This also makes it consistent with the text on the 3rd alert confirmation/trial start page.

Before:
screen shot 2015-08-18 at 11 21 36 am

After:
screen shot 2015-08-18 at 11 18 08 am
screen shot 2015-08-18 at 11 17 52 am

closes #721 
closes #722

### DIFF
--- a/app/assets/stylesheets/partials/_payment_form.scss
+++ b/app/assets/stylesheets/partials/_payment_form.scss
@@ -1,6 +1,6 @@
 .registration-page,
 .registration-success-page {
-  max-width: 42em;
+  max-width: 43em;
   padding-bottom: 4em;
 
   @include at-breakpoint(40em) {
--- a/app/views/subscriptions/new.html.haml
+++ b/app/views/subscriptions/new.html.haml
@@ -5,12 +5,14 @@
     %h1.page-title Subscribe here
 
     %p
-      Subscribe here to get alerts of planning applications near several addresses,
-      or to use PlanningAlerts for your work.
+      %strong
+        Subscribe here to get alerts of planning applications near several addresses,
+        or to use PlanningAlerts for your work.
       Once you’re subscribed you can sign up for alerts from as many locations as you need.
 
     %p
-      Non-commercial use of PlanningAlerts is free for any number of addresses.
+      %strong
+        Non-commercial use of PlanningAlerts is free for any number of addresses.
       If this sounds like you, please #{mail_to "contact@planningalerts.org.au", "email us"}—
       we’ll be happy to help with a free, non-commercial subscription.
 
--- a/app/views/subscriptions/new.html.haml
+++ b/app/views/subscriptions/new.html.haml
@@ -6,7 +6,7 @@
 
     %p
       %strong
-        Subscribe here to get alerts of planning applications near several addresses,
+        Subscribe here to get alerts for several addresses,
         or to use PlanningAlerts for your work.
       Once you’re subscribed you can sign up for alerts from as many locations as you need.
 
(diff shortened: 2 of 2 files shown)


henare:
Way better :+1: 

Something is not quite right about the first sentence:

>Subscribe here to get alerts for several addresses, or to use PlanningAlerts for your work.

I think it's the "or", which isn't strictly accurate either. How about:

>Commercial users of PlanningAlerts must subscribe here to get alerts for several addresses.

equivalentideas:
> Commercial users of PlanningAlerts must subscribe here to get alerts for several addresses.

Hmmm, I think that takes us back to being indirect and describing a situation, rather than saying what they can/should do now.

Do you think the concept of 'using PlanningAlerts in your work' wont be understood as 'commercial use'?

henare:
>Do you think the concept of 'using PlanningAlerts in your work' wont be understood as 'commercial use'?

Just:

>I think it's the "or", which isn't strictly accurate either.

equivalentideas:
> which isn't strictly accurate either.

Why isn't that accurate? If you're getting several alerts, you need to subscribe, or if you're using it commercially you need to subscribe. Non-commercial users with several alerts still need a subscription, but they can get it for free if they email us.

henare:
>If you're getting several alerts, you need to subscribe, or *if you're using it commercially you need to subscribe*.

Are you sure?

equivalentideas:
I assume you're hinting that it's free for commercial use for one or two alerts. That's true. Hmmm. We should be honest and explain the actual situation rather than leave in ambiguous. Is this what you're talking about?

henare:
It's not really about honesty it's just that in the real world someone using it commercially without several alerts will never even see the subscription stuff so it's not necessary to say "or". It just makes the sentence more complex to understand and is effectively inaccurate, and therefore confusing.

equivalentideas:
The 'or' suggests that if you're using it commercially you can't just unsubscribe from an alert and keep using it for free—which isn't the case.

> someone using it commercially without several alerts will never even see the subscription stuff

I think that's the important thing, we should focus the copy on people in that specific situation, you're using PlanningAlerts commercially and getting several alerts.

What if we drop the mention of work from the opening sentence and put it into the second sentence. We're assuming they're using it in their work, and the non-commercial statement makes it very clear who this does and doesn't apply to.

> Subscribe here to get alerts for several address. Once you're subscribed you can sign up for as many alerts are you need for your work.

Second sentence could be nicer somehow but you get the idea.

henare:
>I think that's the important thing, we should focus the copy on people in that specific situation

Good call.

What about just dropping the "or"?

>Subscribe here to get alerts for several addresses to use PlanningAlerts for your work.

Or even adding something more pointed and instructive?

>You need to subscribe here to get alerts for several addresses to use PlanningAlerts for your work.

equivalentideas:
What about kinda combining those two:

> Subscribe here to get alerts for several addresses if you use PlanningAlerts for your work.

henare:
Nice. That's friendly too.

equivalentideas:
> Nice. That's friendly too.

I'll change it to that.

equivalentideas:
Whoops, sorry @DylanFM didn't mean to assign you ;-)

DylanFM:
Hehe no probs :)

PR #749, opened 2015-08-21
Track how trial subscribers respond so we can improve subscription flow

### LINKED ISSUE #747: Track metrics to judge the success of the subscription process
This is broken out from discussion on #745 

What are the key metrics we need to track so we can improve the subscription process so that when we roll it out we can see what's working and what isn't?

We can look to our flow chart in #692 to see the different points people will step through.

The steps in the key initial flow for people already with lots of alerts is:

* they click on the link in the trial banner in their email alert
* they click subscribe on the subscription page
* they submit their payment details and then hit the subscription confirmation page which means their payment has been successful

There are other flows like _signing up via the alert confirmation page_ and _signing up via the trial expired alerts_ , but I think let's get this initial one set up first.

* [x] we've already go a ga event registered on the subscribe button
* [ ] we can track people hitting the confirmation page with page views, but it could be good to put an event their as well so we can see the actual journeys through google analytics.
* [x] I think we'd need to put a ga email campaign param on the link from the trial banner to track that step, to track what happens to people coming to the site through that trial banner.

@henare can you think of anything else we might want to track at the moment? Or can you think of better ways to do this?

equivalentideas:
This adds Google Analytics tracking params to the trial banner links so we can specifically see how trial subscribers move through the site.

This doesn't feel very DRY, but I couldn't work out how to move that up into the mailer, so I did the dumb thing.

Closes #747.

### DIFF
--- a/app/mailers/alert_notifier.rb
+++ b/app/mailers/alert_notifier.rb
@@ -4,7 +4,7 @@ class AlertNotifier < ActionMailer::Base
 
   def alert(theme, alert, applications, comments = [])
     @alert, @applications, @comments = alert, applications, comments
-
+    @analytics_params = {utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"}
 
     themed_mail(theme: theme, from: email_from(theme), to: alert.email,
       subject: render_to_string(partial: "subject",
--- a/app/mailers/alert_notifier.rb
+++ b/app/mailers/alert_notifier.rb
@@ -4,7 +4,7 @@ class AlertNotifier < ActionMailer::Base
 
   def alert(theme, alert, applications, comments = [])
     @alert, @applications, @comments = alert, applications, comments
-    @analytics_params = {utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"}
+    @trial_subscriber_analytics_params = {utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"}
 
     themed_mail(theme: theme, from: email_from(theme), to: alert.email,
       subject: render_to_string(partial: "subject",
--- a/app/views/alert_notifier/_trial_banner.html.haml
+++ b/app/views/alert_notifier/_trial_banner.html.haml
@@ -3,7 +3,7 @@
     %tr
       %td
         .trial-banner-button-container
-          = link_to "Subscribe now", new_subscription_url(email: @alert.email, utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"), class: 'trial-banner-button'
+          = link_to "Subscribe now", new_subscription_url(@analytics_params.merge(email: @alert.email)), class: 'trial-banner-button'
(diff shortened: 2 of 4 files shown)


henare:
>This doesn't feel very DRY, but I couldn't work out how to move that up into the mailer, so I did the dumb thing.

You could extract this into a helper.

equivalentideas:
> You could extract this into a helper.

Nice, thanks. I'll do that now.

henare:
Or just put them in the mailer - morph.io does this.

equivalentideas:
> Or just put them in the mailer - morph.io does this.

Hey @henare , I've just been trying this and I'm blocked. I don't understand why the instance variable I'm setting in the mailer is nil in the partial. I feel like I'm missing something easy but I just can't see it. Could you have a quick look at this for me when you get a chance please.

henare:
@equivalentideas this is working it's just the tests that are broken.

In view tests you have to [assign](http://www.rubydoc.info/gems/rspec-rails/file/README.md#assign_key__val_) the instance variable in the test because they're isolated from the controller. So because `@analytics_params` is not set in the test this is failing.

I've added a commit to fix this.

app/mailers/alert_notifier.rb
```diff
@@ -4,7 +4,7 @@ class AlertNotifier < ActionMailer::Base
 
   def alert(theme, alert, applications, comments = [])
     @alert, @applications, @comments = alert, applications, comments
-
+    @analytics_params = {utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"}
```
henare: We use analytics parameters in other links so I'm not sure it's a good idea to call these *the* analytics parameters when they're only used on the trial subscriber links.

app/views/alert_notifier/_trial_banner.html.haml
```diff
@@ -3,7 +3,7 @@
     %tr
       %td
         .trial-banner-button-container
-          = link_to "Subscribe now", new_subscription_url(email: @alert.email), class: 'trial-banner-button'
+          = link_to "Subscribe now", new_subscription_url(@analytics_params.merge(email: @alert.email)), class: 'trial-banner-button'
         %p
-          Your #{link_to "trial subscription", new_subscription_url(email: @alert.email)}
+          Your #{link_to "trial subscription", new_subscription_url(@analytics_params.merge(email: @alert.email))}
```
henare: It's probably a good idea to differentiate these links with `utm_content`.

henare:
@equivalentideas I've added a few comments, reassigning to you to review them.

equivalentideas:
> In view tests you have to assign the instance variable in the test because they're isolated from the controller. So because @analytics_params is not set in the test this is failing.

Thanks @henare that's a really helpful explanation!

I'll make those suggested comments.

equivalentideas:
I've made those changes, assigning back to @henare

henare:
:snowflake:

PR #751, opened 2015-08-21
Ensure that the inline link is clearly a link in all email clients

### LINKED ISSUE #748: "Your subscription" link not visible on Android
screenshot_2015-08-20-10-29-02

equivalentideas:
This commit declares the underline specifically on the inline link, and adds !important to the button style to make sure it isn't underlined.

fixes #748


henare:
If you say so :smiley:

PR #758, opened 2015-08-25
Email css

equivalentideas:
This fixes bugs in the rendering of the trial banner in alert emails, with the colouring of the links and background in the banner and it's padding.

I'm gonna run an inbox inspection in Mailchimp to test this.


equivalentideas:
Ok these changes fix the major bugs, that the white text was unreadable when the client didn't render the background colour of the banner. Far from ideal, but it's usable. 

Here's outlook 2013, the problem client:
screen shot 2015-08-25 at 11 42 20 am

The email template should be revisited and probably rebuilt from scratch soon to make it more stable across email clients and generally nicer.

Assigning to @henare for review/merge.

PR #759, opened 2015-08-27
Allow free subscriptions

### LINKED ISSUE #700: Provide non-commercial users free access to as many alerts as they like
We need a way to flag people as non-commercial. I think it would be good to also attach a note with the explanation of why they're non-commercial to for records/avoid confusion.

henare:
Fixes #700.


henare:
@equivalentideas assigning to you for review.

equivalentideas:
Nice, I this all looks good to me :+1:

henare:
@equivalentideas thanks for the review! :eyeglasses:

PR #762, opened 2015-08-31
Fix campaign params and add helpers to simplify views

equivalentideas:
Why:

Trying to make sure all the tracking params were set properly, I discovered that their was a lot of repetition in setting them, which was leading to errors. The utm_campaign param turned out to be left out in more places that I originally thought, which I only learnt after removing the repetition.

This PR:

* Removes some logic and a lot of repetition from the alert_notifier views by adding new url helpers
* fixes the places where utm_campaign is not being set
* changes the utm_source on subscriber links to be consistent with other links see https://github.com/openaustralia/planningalerts/commit/d8e68d1b5fae9f9fbfe3f6e30a052959eda2abf0

- [x] TODO: There is still a lot of repetition in the way we are passing the protocol and host for the current theme through the controller into the view. I feel like there must be a way to work out what they should be in the helpers so this stuff can be removed from the views. Haven't worked that out yet.


equivalentideas:
Picking this back up now. I'm going to deal with the TODO above.

equivalentideas:
Hey @henare , could you review this please?

app/helpers/alert_notifier_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+    { host: host(theme), protocol: protocol(theme) }
+  end
+
+  def base_tracking_params
+    { utm_source: "alerts", utm_medium: "email" }
+  end
+
+  def application_url_with_tracking(theme: nil, id: nil)
+    base_params = host_and_protocol_for_theme(theme)
+                       .merge base_tracking_params
```
henare: Having this hanging and using implicit parentheses makes it look awkward to me. Is there any reason not just to have it on the same line?

This comment applies to the other instances of this too.

app/helpers/alert_notifier_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+    application_url(
+      base_params.merge(
+        id: id,
+        anchor: 'add-comment',
+        utm_campaign: 'add-comment'
+      )
+    )
+  end
+
+  def new_subscription_url_with_tracking(alert: nil, utm_content: '')
```
henare: How about defaulting `utm_content` to `nil` for consistency?

app/mailers/alert_notifier.rb
```diff
@@ -3,9 +3,10 @@ class AlertNotifier < ActionMailer::Base
   helper :application, :applications
 
   def alert(theme, alert, applications, comments = [])
-    @alert, @applications, @comments = alert, applications, comments
-    # TODO: Differentiate people with expired subscriptions
-    @trial_subscriber_analytics_params = {utm_source: "alert", utm_medium: "email", utm_campaign: "trial_subscriber"}
+    @theme = theme
+    @alert = alert
+    @applications = applications
+    @comments = comments
```
henare: You could set all these on one line like before so it's more concise.

app/views/alert_notifier/_comment.html.haml
```diff
@@ -5,4 +5,5 @@
       by
       = comment.name
       on
-    = link_to comment.application.address, application_url(comment.application, protocol: @protocol, host: @host, anchor: "comment#{comment.id}", utm_medium: "email", utm_source: "alerts")
+    = link_to comment.application.address,
+        comment_url_with_tracking(comment: comment, theme: @theme)
```
henare: Does this really need to be wrapped onto 2 lines?

app/views/alert_notifier/_comment.text.erb
```diff
@@ -2,4 +2,7 @@
 <%= Sanitize.clean(truncate(comment.text, :length => 700), Sanitize::Config::BASIC).html_safe %>
 by <%= comment.name %> on <%= comment.application.address %>
 
-<%= application_url(comment.application, protocol: @protocol, host: @host, anchor: "comment#{comment.id}", utm_medium: "email", utm_source: "alerts").html_safe %>
+<%= comment_url_with_tracking(
+  comment: comment,
+  theme: @theme).html_safe
+%>
```
henare: Could you make this just be on one line so it's a bit more readable?

spec/helpers/alert_notifier_helper_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+    context "when the theme is Default" do
+      theme = "default"
+
+      it {
+        expect(helper.host_and_protocol_for_theme(theme))
+          .to eq(
+            host: host(theme),
+            protocol: protocol(theme)
+        )
+      }
```
henare: I found all the above kind of code wrapping in this file quite jarring but then again I'm not working with the code. If you found it more readable then keep it as is.

henare:
Great to centralise all this and fix the bugs of missing campaign parameters :+1: 

I've add a few minor style comments but apart from that, based on the strength of the tests all passing, I'm happy to merge this.

equivalentideas:
Thanks for that @henare :hand: I'll address these comments now.

app/helpers/alert_notifier_helper.rb
equivalentideas: Yep I agree, it does seem a bit weird. I'll add the parentheses and move onto one line as:

    base_params = host_and_protocol_for_theme(theme).merge(base_tracking_params)

app/mailers/alert_notifier.rb
equivalentideas: I split these apart in https://github.com/openaustralia/planningalerts/commit/be39ee3eb561a8b065b89da2abe1f8412a96d08d 

It's less concise, but I think more readable to have the assignment right next to the term, rather than look back and forth along the line. Ruby style guide: https://github.com/bbatsov/ruby-style-guide#parallel-assignment

Happy to leave this as is?

henare: :+1:

spec/helpers/alert_notifier_helper_spec.rb
equivalentideas: I'm gonna leave this as is because I find it more readable when inputs and outputs in these tests.

equivalentideas:
Thanks @henare I've made most of those style changes and a few others for consistency.

Assigning back to you for merge and deploy when you’re ready

Issue #687, opened 2015-07-29
Set up google analytics events/goals for commenting


equivalentideas:
So we can better understand and test how people are using the form. For example, do most people who comment come from the comment emails or the alert emails?

equivalentideas:
So this is broken at the moment because GA goal expects people to hit a page /comments/[:comment_id]/confirmed but no one ever hits that page because we redirect to the application show view, https://github.com/openaustralia/planningalerts/blob/master/app/controllers/comments_controller.rb#L34 .

We can track the comment confirmation with an ga event fired when the confirmation message is on the application show page and see how that works. Unfortunately that means we can't define a funnel in GA goals (as far as I can tell) because you can't define funnels or event based goals. We can track the event flow info though, but it's more annoying/complex. That's the best option I can think of.

henare:
Can we do it server-side as part of the controller action?

equivalentideas:
hmmm that's a good question. It appears it is, using the [GA Measurement Protocol](https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide). It involves getting data from a cookie on the client.

It seems like it'd be simpler to do it client-side to me—but I've never done anything with cookies before, I'm not sure what the considerations and work involved for that would be.

equivalentideas:
I'm going to make a PR to do with on the client, seems simplest to me.

equivalentideas:
We also should be tracking clicks on the comment submit button to see if people aren't confirming their comment. I'm going to add that to #763 .

equivalentideas:
One more thing to track in this process before we close this I think:

* [ ] track when people click the link at the top of a list of comments to make a comment

equivalentideas:
I've just set the goal up.

PR #763, opened 2015-09-11
Comment confirmation tracking

equivalentideas:
## Why

At the moment our attempts to track comment confirmation in Google Analytics aren't working (see [comment](https://github.com/openaustralia/planningalerts/issues/687#issuecomment-132423667) on #687 ). We want to be capture these actions in ga so we can better understand what people do before and after they comment.

## These commits

* add a class to target and some js to fire the event when GA is loaded and the confirmation message is on screen.


equivalentideas:

equivalentideas:
I'm going to add tracking of clicks on the comment submit button to this—assigning back to me.

equivalentideas:
Assigning back to @henare for review/merge

Issue #773, opened 2015-09-14
[Broken into parts and closed] As someone interested in a local development application, help me write to my local councillor about it, so that I can get their help or feedback and find out where they stand on this development I care about


equivalentideas:
From [our 3 month plan blog post](https://www.openaustraliafoundation.org.au/2015/06/25/the-next-3-months-are-going-to-be-really-busy-here-is-our-plan/):

> Then, Luke and Henare will start working on our next major project. The idea is simple. In PlanningAlerts allow people to write to their councillor about a development application on top of what it currently does which allows you to write a comment on a development application which directly feeds back into the official process.
>
> The aim is to strengthen the connection between citizens and local councillors around one of the most important things that local government does which is planning.
>
> This will actually be a surprisingly technically involved project which will involve quite a few different pieces. We will need to find out every councillor in the country, including their contact details (and hopefully their picture). We already starting collecting this data by writing scrapers. Then, information will need to get crowdsourced. We plan to store this information in PopIt (so it can be reused by others) and we will integrate PlanningAlerts with WriteIt to actually send the emails to the councillors.
>
> Then, of course, there’s the vital bit of making this work beautifully and simply within the existing PlanningAlerts application. This is the most fun, satisfying and crucial bit.

We also did this flow sketch at our kickoff meeting (not the best image quality):

![img_4184](https://cloud.githubusercontent.com/assets/1239550/9844848/ef2a69ae-5b09-11e5-8376-788874e29464.jpg)

This shows our idea for how messages to councillors would sit next to comments to council.

equivalentideas:
I'm going to start this off by digging into a static prototype of the feature. During that process I'll bump up against the current [UX commenting issues](https://github.com/openaustralia/planningalerts/issues?q=is%3Aopen+is%3Aissue+label%3Acommenting+label%3AUX) and fix them as I go.

henare:
Nice :ok_hand:

equivalentideas:
Picking up from #755:

> and I had a think about how we can reduce the number of decisions a user has to make upfront. People just want to have their say and have it make an impact—I'm going to iterate this into a single form that presents different options depending on your choices.

I'm going to try a version of this where you get straight into writing your message, then get to choose who it goes to. During this step the implications of these decisions are explained. Then you finish off the final details (name and address), and the get a clear warning that your name and comment will be public before you post your comment.

[...]

PR #775, opened 2015-09-15
Add section for messaging councillors and make comment intro copy more direct or human focused

equivalentideas:
I've started on #773 by breaking the comments section into tabs and adding a note about the upcoming feature. **Currently you only see this if you're an admin and logged in**.

In the course of making the split I made a few refactors by extracting some partials from the comments section.

I also made a slight reworking of the comment intro text to make it more direct and focused on the person making the comment. **This change is visible to all users.**

It was:

> All the information you provide in your comment below will be sent to City of Marion where it will be considered when they decide whether to approve this application or not. The text of your comment and your name will be public.

It's now:

> Your comment will be sent to Rockdale City Council. They’ll consider your submission when they decide whether to approve this application. Your name and comment will be posted publicly here as well.

Behind the feature flag for admins you get:
screen shot 2015-09-15 at 5 31 14 pm
screen shot 2015-09-15 at 5 31 10 pm

Everyone get:
screen shot 2015-09-15 at 5 27 48 pm


app/views/applications/show.html.haml
```diff
@@ -34,61 +34,10 @@
...
-              = text_field_tag :little_sweety, "", placeholder: "Please leave this blank"
-          = f.actions do
-            = f.action :submit, button_html: {class: "button button-action"}
-      - elsif !@application.comment_url.blank?
-        %p
-          = link_to "How to comment", @application.comment_url
-          on this application (from #{@application.authority.full_name}).
-          - if @application.on_notice_to
-            = on_notice_text(@application)
+    - if current_user && current_user.admin?
```
equivalentideas: This is the feature flag

henare:
Looking good! :hatching_chick:

There's a slight meaning change here however:

>All the information you provide in your comment below will be sent to...

vs

>Your comment will be sent to...

The first one accurately indicates to the person that everything in the form will be sent to the authority (name, address, etc.) whereas the new wording could be understood to only mean the comment text gets sent to the authority. I think this needs a tweak but other than that this looks good to go :shipit:

equivalentideas:
>  There's a slight meaning change here however:

Nice, thanks for that :+1:  I'll have a go.

equivalentideas:
@henare I've changed that line to:

> Your comment and details will be sent to Rockdale City Council

I think that’s nice cause it puts the focus on the comment (the interesting bit) and then makes it clear that the you details will also be sent.

app/models/application.rb
```diff
@@ -161,6 +161,14 @@ def find_all_nearest_or_recent
     end
   end
 
+  def official_submission_period_expired?
+    if on_notice_to && Date.today > on_notice_to
+      true
+    else
+      false
+    end
```
henare: This entire method can become:

```ruby
on_notice_to && Date.today > on_notice_to
```

equivalentideas: Nice!

equivalentideas: That produces a different result in fact, because if on_notice_to isn't set it returns nil. For our use of this, we want to return false—we don't know if it's expired so we cannot say it's true.

It can be compressed to:

  on_notice_to && Date.today > on_notice_to ? true : false

equivalentideas:
These changes add a really bare bones form for messaging your councillor behind the feature flag. Putting this in front of a user gave me the feedback that having two separate sections for comments is quite confusing and damages the users confidence in what they need to do to.

@henare and I had a think about how we can reduce the number of decisions a user has to make upfront. People just want to have their say and have it make an impact—I'm going to iterate this into a single form that presents different options depending on your choices.

@henare I'd like to get this merged and deployed this now if you’re happy with that before this PR gets any more massive.

app/models/application.rb
henare: nil will evaluate as false so does that really matter for what you're using it for?

henare:
>@henare I'd like to get this merged and deployed this now if you’re happy with that before this PR gets any more massive.

Good call. Just assign it to me when it's ready to rock :guitar:

app/models/application.rb
equivalentideas: :+1: fixed with 7eb5ce5

spec/features/write_to_councillors_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      within("#new_user") do
+        fill_in "Email", with: admin.email
+        fill_in "Password", with: admin.password
+      end
+      click_button "Sign in"
+      expect(page).to have_content "Signed in successfully"
+
+      VCR.use_cassette('planningalerts') do
+        application = create(:application, id: "1", authority_id: authority.id, comment_url: 'mailto:foo@bar.com')
+        visit application_path(application)
```
henare: This action should be in the scenario because it's part of that process I think.

spec/features/write_to_councillors_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      authority = create(:authority, full_name: "Foo")
+      VCR.use_cassette('planningalerts') do
+        application = create(:application, id: "1", authority_id: authority.id, comment_url: 'mailto:foo@bar.com')
+        visit application_path(application)
+      end
+    end
+
+    scenario "can’t see councilor messages sections" do
+      page.should_not have_content("Write to your elected councillors about this application")
+      # and you should not be able to write and submit a message.
```
henare: I think these should be TODOs.

henare:
Heaps of great code @equivalentideas :sparkles: 

I'll make those minor adjustments then merge and deploy this :shipit:

equivalentideas:
> Heaps of great code @equivalentideas :sparkles:
> I'll make those minor adjustments then merge and deploy this :shipit:

Thanks!

PR #779, opened 2015-09-18
Fix regression where partial can't be found, fixes #778

### LINKED ISSUE #778: Submitting the comment form with a missing field causes a server error
It should show a nice message to the user instead.

    [PlanningAlerts/production] ActionView::Template::Error: Missing partial comments/_comments_area, application/_comments_area with {:locale=>[:en], :formats=>[:html, :text, :js, :css, :ics, :csv, :vcf, :png, :jpeg, :gif, :bmp, :tiff, :mpeg, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :coffee, :arb, :haml, :rabl]}. Searched in:  * "/srv/www/www.planningalerts.org.au/app/releases/20150918053200/app/views"  * "/srv/www/www.planningalerts.org.au/app/shared/bundle/ruby/2.0.0/gems/devise-3.4.1/app/views"  * "/srv/www/www.planningalerts.org.au/app/shared/bundle/ruby/2.0.0/bundler/gems/activeadmin-7e72b7881eb6/app/views"  * "/srv/www/www.planningalerts.org.au/app/shared/bundle/ruby/2.0.0/gems/kaminari-0.16.3/app/views"

## Backtrace

    line 40 of [PROJECT_ROOT]/app/views/applications/show.html.haml: _app_views_applications_show_html_haml__962219298_105754090
    line 25 of [PROJECT_ROOT]/app/controllers/comments_controller.rb: create

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37075/faults/15376779)

equivalentideas:
When the applications/show is called from the comment controlled,
it searches for partials in the views/comments directory.

This raises an error because partials were extracted from this view using just
the partial name.

Using an absolute path (from the view directory) fixes this as
rails can always find the partial no matter which controller the
applications/show view is called from.

This also adds an integration test for the situation that
this bug was discovered in.

### DIFF
--- a/app/views/applications/_comments_area.html.haml
+++ b/app/views/applications/_comments_area.html.haml
@@ -7,4 +7,4 @@
       - @comments.each do |comment|
         %li{:id => "comment#{comment.id}"}
           = render :partial => 'comments/comment', :locals => {:comment => comment}
-  = render "comment_form"
+  = render "applications/comment_form"
--- a/app/views/applications/show.html.haml
+++ b/app/views/applications/show.html.haml
@@ -35,9 +35,9 @@
         %p.source (Source: #{link_to @application.authority.full_name, authority_path(@application.authority.short_name_encoded)}, reference #{@application.council_reference})
 
     - if current_user && current_user.admin?
-      = render "comments_options_tabs"
+      = render "applications/comments_options_tabs"
     - else
-      = render "comments_area"
+      = render "applications/comments_area"
 
 #sidebar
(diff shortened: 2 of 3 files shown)


henare:
Sweet tests! It's :green_heart: so I can just merge and :shipit:

PR #781, opened 2015-09-21
Fix for #778 when your logged in as admin

equivalentideas:
This adds absolute path references for partials in the applications/show view, which fixes a bug where you get a template not found error if your comment gets rejected and the show view loads from the comments controllers. See #778 for more details on the bug.

e79f79f was a fix for this, but it didn't work for admins because their are other partials that only they see that were not included.

This also moves the comment form related partials to a more logical home in the comments view folder.


equivalentideas:
Assigning to @henare for review/merge

henare:
Thanks for remembering this :alien: 

There are now two `councillor_message_form` partials, is this correct?

equivalentideas:
> There are now two councillor_message_form partials, is this correct?

Nope :S Sorry I missed that.

equivalentideas:
> There are now two councillor_message_form partials

I've removed the stay file in defa00f. Thanks for picking that up @henare , assigning back to you.

PR #785, opened 2015-09-23
Comment flow performance

### LINKED ISSUE #772: As an developer of PlanningAlerts, show me some stats about the use of the site, so I can better understand how it is being used
I'm always surprised to hear how many people there are signed up for alerts, I thought there was around 25000, but Henare tells me there's much more.

A simple chart of new signups per week would be helpful on a stats page, like https://morph.io/users/stats or https://www.gov.uk/performance/prison-visits .

This would help us track the performance of our changes.

It would be also helpful to get a better understanding of how comments are being used and how our changes impact use. Metrics could include like:

* number of people who commented per week
* number of comments per person per week

equivalentideas:
## Why

Focused data about the use of the commenting section will help us judge it's success and see how people respond to our changes. Seeing how many new commenters and returning commenters we're getting each day will help us judge our current work on the comments flow.

## These commits:

Add a bare bones /performance page with a table containing our comments data. A chart would be quicker to view and see patterns in, but for a first pass I think a table will begin to give us useful information. We can see how this works without investing time developing a chart, and then enhance this if it's heading in the right direction.

Closes #772 (open new issues for specific additions or enhancements to this performance page)

@henare could you please review this? I've moved the methods to scopes as suggested and made some other refactors since yesterday.

screen shot 2015-09-23 at 10 18 18 am


app/views/static/performance.html.haml
```diff
@@ -0,0 +1,23 @@
+- content_for :page_title, "Performance"
```
henare: This page is not static so it seems like it's on the wrong controller to me.

app/views/static/performance.html.haml
```diff
@@ -0,0 +1,23 @@
...
+  %table
+    %caption
+      New and returning people commenting on applications over time
+    %thead
+      %tr
+        %th Date
+        %th First time commenters
+        %th Returning commenters
+    %tbody
+      - (3.months.ago.to_date..Date.today).to_a.reverse.each do |day|
```
henare: I think all this logic belongs in the controller, not the view.

spec/factories.rb
```diff
@@ -19,6 +19,14 @@
     text "a comment"
     address "12 Foo Street"
     association :application
+
+    factory :unconfirmed_comment do
```
henare: :ok_hand:

app/views/static/performance.html.haml
equivalentideas: Yep I agree. It might be quite clear to have a performance controller for this page.

app/models/comment.rb
```diff
@@ -7,6 +7,20 @@ class Comment < ActiveRecord::Base
   scope :visible, -> { where(confirmed: true, hidden: false) }
   scope :in_past_week, -> { where("created_at > ?", 7.days.ago) }
 
+  scope :visible_with_unique_emails_for_date, ->(date) {
+    visible.where("date(created_at) = ?", date).group(:email)
+  }
+
+  scope :by_first_time_commenters_for_date, ->(date) {
+    visible_with_unique_emails_for_date(date)
+    .select {|c| where("email = ? AND created_at < ?", c.email, c.created_at.to_date).empty? }
```
henare: These will result in dozens of queries for each page view which is obviously not good. Unfortunately I can't think of a better way to do this right now :-/

app/views/static/performance.html.haml
henare: :+1:

app/models/comment.rb
equivalentideas: Lets add a TODO to work out a more efficient way to do this.

henare: Fair call. Maybe it won't be a problem anyway?

equivalentideas: > Fair call. Maybe it won't be a problem anyway?

As discussed in chat, we could add some caching to help here.

app/views/performance/index.html.haml
```diff
@@ -0,0 +1,23 @@
...
+  %table
+    %caption
+      New and returning people commenting on applications over time
+    %thead
+      %tr
+        %th Date
+        %th First time commenters
+        %th Returning commenters
+    %tbody
+      - @comments_metrics_timespan.each do |day|
```
equivalentideas: I've moved some of this logic into the controller, are their any other refactors you can see here @henare ?

app/controllers/performance_controller.rb
```diff
@@ -0,0 +1,6 @@
+class PerformanceController < ApplicationController
+  def index
+    @comments = Comment.all
```
henare: This is going to be huge! You should be getting all the values you need to use in your view here, not just getting all the comments then still calling methods against them all in the view.

app/views/performance/index.html.haml
henare: See my comment on the controller.

app/views/performance/index.html.haml
```diff
@@ -0,0 +1,20 @@
...
+%table
+  %caption
+    New and returning people commenting on applications over time
+  %thead
+    %tr
+      %th Date
+      %th First time commenters
+      %th Returning commenters
+  %tbody
+    - @daily_comment_metrics.each do |day|
```
equivalentideas: While this collection could be extracted into a partial, I think it's actually more helpful to have all the table markup inline, since it's such a simple table.

app/controllers/performance_controller.rb
equivalentideas: As discussed I've now extracted all the queries into the controller and removed this behemoth Comment.all call :elephant:

equivalentideas:
@henare I've had a go at implementing your suggestions, assigned back to you :baseball:

Issue #767, opened 2015-09-14
Make commenting button text more human and engaging


equivalentideas:
Currently links to add a comment use the text “add your own comment” and then to submit a comment “create comment”:

Application page:
![screen shot 2015-09-14 at 9 52 16 am](https://cloud.githubusercontent.com/assets/1239550/9839768/69ab1bc2-5ac6-11e5-9387-2db9f4bfb902.png)

Alert email:
![screen shot 2015-09-14 at 9 43 01 am](https://cloud.githubusercontent.com/assets/1239550/9839769/69b1d700-5ac6-11e5-8b67-55dd87aee7db.png)

Comment form:
![screen shot 2015-09-14 at 9 52 24 am](https://cloud.githubusercontent.com/assets/1239550/9839864/6dff8ed6-5ac8-11e5-8a97-fc8ec3409e3b.png)

These terms sound like database functions, where as they should sound human and encourage people to contribute their thoughts.

Relates to #766

equivalentideas:
Looking in the tests I notice we refer to the comment feature as 'giving feedback' on an application. This is much more direct terminology than 'comment' here:

> Give feedback on this application

Instead of:

> Add your own comment

There's something I like about the word feedback that implies a back and forth.

Another nice term is 'have your say'. Sounds less bureaucratic than 'feedback', but is less specific to this process. Hmmm.

henare:
A nice thing about "comment" is that it's a well understood website concept (leaving a comment on a blog or news article).

So while "feedback" might be more accurate (I think that's what you mean by "direct terminology"), it could also be more foreign and daunting for people. It's really important that commenting is not foreign or daunting - that's what makes it awesome right now.

mlandauer:
Not sure about this one. I guess "Add comment" instead of "Create comment" makes sense to me. It doesn't feel like a big problem to me right now. I think I'll close this one

PR #787, opened 2015-09-25
Use a more intuitive verb and make comment button personal

equivalentideas:
## Why

Wording of the 'submit comment button' is not relevant to people making a comment, it is currently "Create Comment".

The term 'create' here refers to the record being created in the database. This is an opportunity to encourage the person and make them feel more comfortable with publicly discussing local planning.

## The PR

The verb 'post' would be clearer here to describe the comment being
posted to the page and also to the council. Post is the term
used by Facebook to describe making a status update as well, so
people may be comfortable with this concept.

Also refer to 'your comment' rather than just 'comment' to
re-enforce the person’s sense of ownership of their action.

## Before
screen shot 2015-09-25 at 11 51 08 am

## After
screen shot 2015-09-25 at 11 50 47 am

Relates to #767

### DIFF
--- a/app/views/comments/_comment_form.html.haml
+++ b/app/views/comments/_comment_form.html.haml
@@ -29,7 +29,7 @@
         = label_tag :little_sweety, 'Little Sweety'
         = text_field_tag :little_sweety, "", placeholder: "Please leave this blank"
     = f.actions do
-      = f.action :submit, button_html: {class: "button button-action"}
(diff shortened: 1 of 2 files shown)


equivalentideas:
@henare I’m pretty confident this is a positive change, I don't think we need to AB test it. What do you think?

henare:
SGTM!

Issue #774, opened 2015-09-15
Donations warning is both easy to miss and is clogging up space above comment input


equivalentideas:
The warning about disclosing your donation in your comment is really easy to miss because it looks like just part of the intro text to comment form—people are likely to skip it as you skip over a par as soon as your brain thinks you get the gist of the section.

At the same time, we want people to feel like making a comment isn't a big deal—having such a long bit of text to introduce the comment input makes it seem like a big deal.

![screen shot 2015-09-15 at 5 04 57 pm](https://cloud.githubusercontent.com/assets/1239550/9870063/2752086a-5bcc-11e5-8c8f-2dc065f2f826.png)

One solution could be to move this to a kind of footer notice at the bottom of the input, reduce it's size and give it a notice colour (yellow):

![img_4190](https://cloud.githubusercontent.com/assets/1239550/9870126/b80ed6c6-5bcc-11e5-8b95-5fe0761723a8.jpg)

We could front load the filtering question “Have you made a donation to ...?" so that people can check this once they’ve started writing and quickly understand if it applies to them.

equivalentideas:
Closed by 682f6d3138f01fa0bd80bd10dc45e7571428c561

PR #788, opened 2015-09-25
Write your comment then decide who it should go to

equivalentideas:
This picks up from https://github.com/openaustralia/planningalerts/issues/773#issuecomment-142468778 to change the flow of writing to councillors.

First you write your comment, then you get options about who to send it to.

This also merges the councillor stuff into the current form so that as this develops, improvements will be felt in the current comment experience, even if there are no councillors.

TODOs:

1. [x] help people understand why they might write to their councillors instead of the council
2. [x] swap the fake data out for a real council (will help with testing)

screen shot 2015-09-28 at 11 43 14 am
screen shot 2015-09-28 at 11 43 23 am


equivalentideas:
Assigned to @henare for review/merge :rocket:

app/controllers/applications_controller.rb
```diff
@@ -107,6 +107,23 @@ def show
     # Required for new email alert signup form
     @alert = Alert.new(address: @application.address)
 
+    if current_user && current_user.admin?
+      if params[:with_councillors] == "true"
+        @application_authority_councillors = [{ id: "1", name: "Mark Gardiner", party: "Independent" },
```
henare: Instead of manually putting an ID on each of these you could use `#each_with_index`. The order of an Array is predictable.

spec/features/write_to_councillors_spec.rb
```diff
@@ -16,13 +16,16 @@
...
     scenario "can’t see councillor messages sections" do
-      page.should_not have_content("Write to your elected councillors about this application")
+      page.should_not have_content("Who should this go to?")
       # TODO: and you should not be able to write and submit a message.
     end
   end
 
   context "when logged in as admin" do
     background do
+      @council_or_councillor_explanation = "Write to the council
```
henare: * This isn't really part of setting up the test so it should be in-lined.
* I find it confusing splitting up this line of text ("why is this line split like this? is it supposed to be telling me something?"). Furthermore we don't really need to read the text to understand the code - it's just a string we're testing for on the page.
* Instead of setting an instance variable you should [use `let`](http://betterspecs.org/#let). It's aliased to `given` in Capybara too.

henare:
From what I can tell, this change should not affect people using the site. The change only shows up if you're logged in as an admin and a parameter, `with_councillors=true`, is present.

@equivalentideas it would be handy if you included something like that :point_up: with the pull request, like your excellently helpful screenshots. It also helps document what we expect this to do.

henare:
* If the commenting period is over I'm warned in bold above the form, "...**not be officially considered**...". For some reason this felt out of place. Now I think I know what it is - the form intro talks about official submissions but then the form allows me to do something else. I think we need to fix this clashing.

* "Your name and comment will be posted publicly here." has been moved from before the form to the bottom of the form but only on the councillors version. If that's the best place, then why not move it on both the forms? If it hasn't been moved because it's a better spot, why has it been moved?

* Are bold titles really needed in the middle of the form? They're out of place in the context of the rest of the site. Are we trying to draw lots of attention to them?

* There's a horizontal line after the recipient selection. I'm not sure what it's supposed to be separating but it feels out of place.

* The list of elected officials is inline with the rest of the radio buttons. Which, when selected, makes them look like part of the same list.

Back to you @equivalentideas, I think there are a few things in there to address before we merge this into production.

equivalentideas:
> Back to you @equivalentideas, I think there are a few things in there to address before we merge this into production.

Thanks @henare lots of good comments. I'll reply to them on the commits/individually.

spec/features/write_to_councillors_spec.rb
equivalentideas: I originally I did it this way cause I was writing it in the test before the view and I found i easier to read/write that way.

But looking at it now I agree its confusing @henare . I think inlining the string is the simplest thing to do.

equivalentideas:
> "Your name and comment will be posted publicly here." has been moved from before the form to the bottom of the form but only on the councillors version. If that's the best place, then why not move it on both the forms? If it hasn't been moved because it's a better spot, why has it been moved?

[In that commit there's the explanation:](https://github.com/openaustralia/planningalerts/commit/ae832d52bc621b57292eb938cdde6abdb64d0e89)

> If the commenter has options as to where their comment will be sent
it doesn't make sense to inform them that the comment will be posted
on this page at the top. It’s more relevant at the point that they’re
going to post their comment.

> This only effects the version of the form with councillors.

> If they can only write to the council, then it still makes sense to
have this notice when the user is told that their comment will go
to the council, because they could assume it just goes to council.

So that's why I made it different in the different versions, and I think it can be merged as is—but I think this raises another question: Why not move the notice that it will be sent to the council to the bottom so people can get straight into writing?

I think this depends on whether people see that notice and then want to write a comment because they know it goes to the council. I'd like to AB test the form with the explanation text below the inputs vs above the inputs.

app/controllers/applications_controller.rb
equivalentideas: Addressed in https://github.com/openaustralia/planningalerts/commit/3deae0c749e1243e7b78cc9cae867cb44d257450

equivalentideas:
> I think there are a few things in there to address before we merge this into production.

I'm pulling a TODO list out the the feedback to work through:

* [ ] "If the commenting period is over I'm warned in bold above the form, "...not be officially considered...". For some reason this felt out of place. Now I think I know what it is - the form intro talks about official submissions but then the form allows me to do something else. I think we need to fix this clashing."
* [ ] "Are bold titles really needed in the middle of the form? They're out of place in the context of the rest of the site. Are we trying to draw lots of attention to them?"
* [ ] "There's a horizontal line after the recipient selection. I'm not sure what it's supposed to be separating but it feels out of place."
* [ ] "The list of elected officials is inline with the rest of the radio buttons. Which, when selected, makes them look like part of the same list."

equivalentideas:
> "Are bold titles really needed in the middle of the form? They're out of place in the context of the rest of the site. Are we trying to draw lots of attention to them?"

@henare in the comment form you're right, these labels stand out against the other inputs too much.

We do actually use this style for form labels through the site though. This black, bold text style is our current label style from the search and signup form e.g. https://www.planningalerts.org.au/alerts/signup

I think it's a good simple style for adding some visual hierarchy to text and works well.

I think that we should address the current imbalance in the comment form by adding labels to the other inputs, as it's a bug not to have labels ( see #795 ) .

equivalentideas:
Other TODOs:

* [ ] Make it so non-admins can see this form with the special params, so it can be tested with people who aren't admins.
* [ ] Prevent people from submitting this form—we don't want emails going to councils from the testing at this point— and give them an alert that it wont be sent and thanks for testing the prototype.

equivalentideas:
>  "If the commenting period is over I'm warned in bold above the form, "...not be officially considered...". For some reason this felt out of place. Now I think I know what it is - the form intro talks about official submissions but then the form allows me to do something else. I think we need to fix this clashing."

When the form allows you to do something in addition to official comments, I've moved this warning into the label of the option to make official submissions. When you're making this choice is when the information is actually useful to you so you get it there. This is in d774094

This could be refined further but I think that can be raised in new issues after this is merged. e.g. @henare suggested reducing the focus on the specific council you are writing to here and making it more about the thing the person cares about, impacting the process/being listened to.

equivalentideas:
I've dealt with all the TODO's here except for the one about the out of place bold form labels. This will be dealt with by merging #797 into this, once that's merged into master.

@henare Can you look at #797 next for me when you're ready? I want to deal with that before I ask you to review this again.

equivalentideas:
This process so far has cleaned up a lot of the information that was previously at the top of the form, and moved it to places where it's move useful.

This has left the 'donations disclosure' message all alone at the top of the form:
screen shot 2015-10-15 at 12 02 20 pm

This is an important consideration, but only relates to a very small set of people who have donated. It also reminds people about the boring bureaucracy of council submission and I think will be off putting in this current primary position in the form.

#774 raises this, and suggests moving this warning to the input hint position below the input, using a similar pattern to the address input hint:

screen shot 2015-10-15 at 12 08 00 pm

I think that’s a good first step to keep this information present, but to push it back to reflect it's lower importance.

#797 makes changes to the comment inputs hint text, and I want to merge that in here, so I'm not going to implement this till then.

equivalentideas:
> #797 makes changes to the comment inputs hint text, and I want to merge that in here, so I'm not going to implement this till then.

Now that #797 is merged into master, I'm going to merge master back into this branch to fix the conflicts.

equivalentideas:
> #774 raises this, and suggests moving this warning to the input hint position below the input, using a similar pattern to the address input hint:
> I think that’s a good first step to keep this information present, but to push it back to reflect it's lower importance.

I've implemented this to work the same way as the address faq link:
screen shot 2015-10-15 at 6 42 38 pm

equivalentideas:
Now that the disclosure warning has been moved out of the intro, their's nothing between the comment section heading and the comment text input for the form with councillors:
screen shot 2015-10-16 at 4 07 21 pm

This creates an odd, double heading or double label effect. I think we should make the heading the form label and use the label text: 'Have your say on this application'. I think that's actually a nice simplification, we don't need a heading anymore, because the label tells you what this is about.

screen shot 2015-10-16 at 4 16 57 pm
or we could use the heading style still:
screen shot 2015-10-16 at 4 14 58 pm

I think that's a good change for the current (no councillors) form too:
screen shot 2015-10-16 at 4 17 59 pm

I think this should be broken off into an issue to fix after this is merged. The situation for the current form (no councillors) is not as bad without the disclosure notice moving into the hint text, because there's other text between the heading and the form:
screen shot 2015-10-16 at 4 19 18 pm

equivalentideas:
@henare this is ready for another review please.

This PR:

* merges the councillor options out of a separate tab and makes it part of the normal comment form (if you have the magic 'with_councillors=true' params set)
* let's non-admins see this form with the 'with_councillors=true' query string so we can test this with non-admins
* prevents you creating a real comment when you submit a comment with the councillor options so testers don't accidentally submit real comments
* moves the donations disclosure text into the hint for the comment text input (like the address field). I made this part of this PR because the form with councillors just had text about donations for it's introduction which made no sense. _This effects both versions of the form because it was a shared win_. 682f6d3138f01fa0bd80bd10dc45e7571428c561 

There’s an issues that this iteration has brought up that I want to deal with separately from this, that I don't think is a blocker to merging it as it just affects the form with councillors:

* The form appears to have two headings (The 'add your own comment heading' and the first label), [see discussion above](https://github.com/openaustralia/planningalerts/pull/788#issuecomment-148612800)

## Councillor options before:
screen shot 2015-10-16 at 5 31 57 pm

## Councillor options after:
screen shot 2015-10-16 at 5 31 33 pm

## Normal form before:
screen shot 2015-10-16 at 5 32 32 pm

## Normal form after:
screen shot 2015-10-16 at 5 32 53 pm

app/helpers/comments_helper.rb
```diff
@@ -2,4 +2,12 @@ module CommentsHelper
   def comment_as_html(text)
     Sanitize.clean(simple_format(text), Sanitize::Config::BASIC).html_safe
   end
+
+  def donation_declaration_notice
```
henare: I'd inline these. They're not reused and they don't contain much logic so it seems unnecessary to have them as helper methods.

henare:
Nice - it's really starting to take shape! :crocodile: 

I just made a tiny code suggestion but apart from that it looks ready to merge to me :rocket:

equivalentideas:
@henare I've inlined those methods and done a little clean up of now unused stuff I'd added along the way.

Assigning back to you.

henare:
Nice. I'll leave it for you to deploy :balloon:

Issue #791, opened 2015-09-28
Put more focus on the person who has commented, to increase the connection between people


equivalentideas:
Currently the design of a comment item plays down the person who has commented. Their name is in low contrast text at the bottom of the comment.

![screen shot 2015-09-28 at 12 12 11 pm](https://cloud.githubusercontent.com/assets/1239550/10126678/2bbff2dc-65da-11e5-8310-012994f98f5f.png)

It's one of the aims of this site to increase connections between people around local planning and local democracy. In the new feature to enable comments to councillors #773 , we're also trying to emphasise the connection between citizens and councillors:

> The aim is to strengthen the connection between citizens and local councillors around one of the most important things that local government does which is planning.

Here are some examples of comment elements from other projects that emphasise connections between people:

![screen shot 2015-09-28 at 12 07 47 pm](https://cloud.githubusercontent.com/assets/1239550/10126713/a050e4ee-65da-11e5-9b91-b9eb32180341.png)
![screen shot 2015-09-28 at 12 03 48 pm](https://cloud.githubusercontent.com/assets/1239550/10126716/aa9ffa70-65da-11e5-88d3-7a83c8cd3849.png)
![screen shot 2015-09-28 at 12 01 43 pm](https://cloud.githubusercontent.com/assets/1239550/10126717/aaa47320-65da-11e5-9018-6f4bfaa1b13e.png)
![screen shot 2015-09-28 at 12 00 35 pm](https://cloud.githubusercontent.com/assets/1239550/10126718/aaa8a7a6-65da-11e5-9f56-8804aaeb9b2c.png)
![screen shot 2015-09-28 at 11 59 39 am](https://cloud.githubusercontent.com/assets/1239550/10126733/f6167074-65da-11e5-81c0-ad6d2329676c.png)

All these examples place the persons name in the primary position. That's one step we can make to put more focus on the people using PlanningAlerts.

equivalentideas:
This was fixed by #792 closing

PR #792, opened 2015-09-28
Emphasise the person behind the comment

equivalentideas:
Why:

We want to emphasise humans: humans who will be impacted by planning decisions, humans who you are writing with when you comment, your neighbours. Currently the person who commented is deemphasised in the design. We should change this to give a more human feel to the comment section.

This PR:

* Moves the name and time to the top of the comment.
* Reduces repetition in the comment views
* Fixes the description of one of the comment view tests https://github.com/openaustralia/planningalerts/commit/228acf019461c13860b0d4dcb8a313ef4b0d9fb8
* Makes the comment view tests more focused https://github.com/openaustralia/planningalerts/commit/a5cb579fc6d6f4f1300ed9c8fbe97453c89fc69e

Before:
screen shot 2015-09-28 at 4 06 40 pm
screen shot 2015-09-28 at 4 06 21 pm

After:
screen shot 2015-09-28 at 4 05 42 pm
screen shot 2015-09-28 at 4 04 59 pm

See #791 

This begins cleaning up this view ahead of adding councillor comments into the mix.

### DIFF
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -72,10 +72,6 @@ ol#comments {
   font-size: 1em;
 }
 
-.comment-meta {
-  color: $font-color;
-}
-
 .comment-address,
 .comment-meta {
   padding: .75em 1em;
@@ -84,6 +80,11 @@ ol#comments {
 .comment-text,
 .comment-actions {
   padding: 0 1em .75em;
+}
+
+.comment-author,
+.comment-time {
+  color: $font-color;
 }
 
 .comment-author {
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -81,7 +81,6 @@ ol#comments {
   padding: .75em 1em;
 }
 
-
 .comment-text,
 .comment-actions {
   padding: 0 1em .75em;
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -89,10 +89,6 @@ ol#comments {
   font-size: 1em;
 }
 
-.comment-meta {
-  color: $font-color;
-}
-
 .comment-author {
   margin-right: .5em;
   display: block;
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -87,7 +87,7 @@ ol#comments {
 }
 
 .comment-author {
-  margin-right: .75em;
+  margin-right: .5em;
   display: block;
   font-weight: bold;
 
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -101,7 +101,7 @@ ol#comments {
   }
 }
 
-.comment-time {
+.comment-time-block {
   font-size: 0.9em;
 }
 
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -8,7 +8,6 @@
   }
 }
 
-
 ol#comments {
   @include no-bullets;
   padding-left: 0;
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -72,6 +72,10 @@ ol#comments {
   font-size: 1em;
 }
 
+.comment-meta {
+  color: $font-color;
+}
+
 .comment-address,
 .comment-meta {
   padding: .75em 1em;
@@ -80,11 +84,6 @@ ol#comments {
 .comment-text,
 .comment-actions {
   padding: 0 1em .75em;
-}
-
-.comment-author,
-.comment-time {
-  color: $font-color;
 }
 
 .comment-author {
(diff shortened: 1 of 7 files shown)


equivalentideas:
Hey @henare could you review this for me please?

henare:
@equivalentideas sure thang.

henare:
I like the reasoning behind this but I think the implementation needs work. As it stands only the positioning has changed and we've even left off the "by" part of the attribution that helped personalise the old design.

Now that the positioning has been changed to emphasise humans we should do the same for the text, "Luke Bacon said about 1 hour ago..." or something (better) along those lines. Make it human - at the moment it's robotic.

The examples in #791 don't emphasise connections between people just because of the common factor of where they place the person's name. There's more to it and it's missing here.

equivalentideas:
> Now that the positioning has been changed to emphasise humans we should do the same for the text, "Luke Bacon said about 1 hour ago..." or something (better) along those lines. Make it human - at the moment it's robotic.

Thanks @henare I'll play with that.

> The examples in #791 don't emphasise connections between people just because of the common factor of where they place the person's name. There's more to it and it's missing here.

I think a big part of those examples are having avatars. I think that's too big a step for us right now—but i like the idea of balancing our lack of images with less contracted text. Are their other elements of those designs that you think we could draw on?

equivalentideas:
I'm quite liking 'commented' @henare what do you think?

screen shot 2015-09-29 at 10 58 38 am
screen shot 2015-09-29 at 10 43 21 am

I also think the way we're displaying time “20 hours ago” is robotic.



Something like: 'half an hour ago', '1 hours ago', 'earlier today', or 'yesterday', then '2 days ago', about '3 months ago' would be nicer.

henare:
Currently the text of the comment is most emphasised bit of content in each comment. I'm concerned that the person is not the most important part for someone viewing comments and that this change will detract from the usefulness of the comments display.

The display of comments is an important social validation for new commenters too and I'm worried that emphasising the commenter over what they're saying will not help this.

After a discussion with @equivalentideas along these lines he'd like to split test this change to ensure it doesn't impact the rate of new comment creation. Assigning back to you.

equivalentideas:
After discussing with @henare we've got different ideas about the impact this could have on the experience.

@henare thinks putting the name above the comment will push back the content that people are interested in: the comment text:

> Currently the text of the comment is most emphasised bit of content in each comment. I'm concerned that the person is not the most important part for someone viewing comments and that this change will detract from the usefulness of the comments display.
>
>The display of comments is an important social validation for new commenters too and I'm worried that emphasising the commenter over what they're saying will not help this.

I think that using the name as a kind of heading to the comment will in-fact draw people into it, and make it more interesting, even if you don't recognise the person who's made the comment. People are interested in people. As above: I think this helps humanise the comment section.

The real questions is, will this help encourage people to comment. I think it will. I'm going to set this up as an AB test. The question we're testing is: will more people comment on applications with comments when the person’s name is over the comment text and is bold, as opposed to it's current style/position

equivalentideas:
For the A/B Test:

## Hypothesis

Laying out the comment item with the commenter's name at the top will:

* improve conversation rate of commenting – current conversation rate 0.13%
* not decrease the conversation rate of alert subscriptions – current conversation rate 2.16%
* increase visitors time on page/site

### Original design
screen shot 2015-09-28 at 4 06 40 pm

### Variation
screen shot 2015-09-29 at 10 43 21 am

## Experiment

According to this [sample size calculator](http://www.evanmiller.org/ab-testing/sample-size.html#!0.13;80;5;0.065;0) we'll need 200,000 visitors per variation to get a significance rate of 5% (the basic we've made other changes on).

Application pages get about 80,000 page views a month. This means it would take 5 months to run this experiment to see its effect on comments.

Given that the comment page is expected to change a bit in that time anyway, running this experiment is not going to be very helpful. I'm calling this experiment off.

equivalentideas:
@henare just revisiting our thoughts above–you wrote:

> Currently the text of the comment is most emphasised bit of content in each comment. I'm concerned that the person is not the most important part for someone viewing comments and that this change will detract from the usefulness of the comments display.

> The display of comments is an important social validation for new commenters too and I'm worried that emphasising the commenter over what they're saying will not help this.

and I wrote:

> I think that using the name as a kind of heading to the comment will in-fact draw people into it, and make it more interesting, even if you don't recognise the person who's made the comment. People are interested in people. As above: I think this helps humanise the comment section.

I'm not sure that we're actually disagreeing that much. I'm saying that the name is a point to draw people into the comment text, which is still the most important bit that people are their for: they aren't there to read names, they're there to read comments ;-) My suggestion is that we can use the layout of the comment to draw more people in to read them, so the layout becomes a little story that they follow, rather than going straight into the comment text, which I think will be more effective.

Anyway—@henare if you still don't like the idea, I'm happy to revert the bits of this we don't want so we can keep the refactors and bug fix.

henare:
>I'm not sure that we're actually disagreeing that much.

I think you're right. The problem I had was that the emphasis had been pushed too far - so let's start by going for a middle ground.

Let's adopt your suggested variation but retain the existing styling:

selection_001

Assigning to you to make the change. Once you assign it back to me I'll merge and deploy (the code already looks good) :+1:

equivalentideas:
> Let's adopt your suggested variation but retain the existing styling

I've made that change. Assigning back to you @henare :rocket: 

screen shot 2015-10-05 at 5 03 10 pm
screen shot 2015-10-05 at 5 06 51 pm
screen shot 2015-10-05 at 5 05 59 pm
screen shot 2015-10-05 at 5 05 43 pm

henare:
I think this looks really nice and subtly achieves what you set out to do :clap: Nice one.

I'm just going to add back the colour style that's dropped off then I'll merge and deploy.

Issue #795, opened 2015-09-29
Form inputs don't have visible labels


equivalentideas:
Currently the comment form has no visible labels:

![screen shot 2015-09-29 at 12 23 02 pm](https://cloud.githubusercontent.com/assets/1239550/10153958/f548a1a6-66a4-11e5-84ab-8194dcc8f1d7.png)

The placeholder text is used as the visual label, and when you start inputing text it disappears:

![screen shot 2015-09-29 at 12 27 32 pm](https://cloud.githubusercontent.com/assets/1239550/10154005/8e17594a-66a5-11e5-8c14-a9721fb3016c.png)

This is causes a number of usability problems as detailed in [this article](https://www.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/):

> * For keyboard users this can be problematic as they must read ahead of current focus when filling in forms.
> * Users, especially those with memory impairments will not have the text label available for reference at the same time as filling in a field.
> * Users may think a text field is already filled in and skip over it. Data suggests that this can be an issue.
> * If the placeholder hint text is useful information it should be available while the user is filling in the field.

On top of this, the low contrast text (which is basically required for this patter because otherwise it will appear as pre-inputted text) is hard to see for people with limited sight or monitors with poor contrast. [The contrast of the text colour against the background doesn't meet the WCAG 2.0 standard for readable text contrast](https://leaverou.github.io/contrast-ratio/#rgb%28169%2C169%2C169%29-on-white).

henare:
:+1: Nice. Alaveteli recently did a lot of form redesigning that incorporated things like this I think.

Issue #766, opened 2015-09-11
Put more focus on getting involved in the comment section


equivalentideas:
Currently the biggest element in the comment section is a heading telling me the number of comments. Then in tiny text is a call to action telling me I can get involved. This seems backwards.

![screen shot 2015-09-11 at 6 10 20 pm](https://cloud.githubusercontent.com/assets/1239550/9810019/85da9dd0-58b0-11e5-96ac-6e7d93b19207.png)

When there are no comments, things are a bit better because the main message is 'Add your own comment', but then their is a big paragraph of text telling me about the comment process. This could be much shorter without loosing information I think.
![screen shot 2015-09-11 at 6 11 08 pm](https://cloud.githubusercontent.com/assets/1239550/9810031/9f360f3a-58b0-11e5-8368-e3715e132c91.png)

Then the placeholder text in the comment box just says 'comment'. This is an opportunity to give a nice example to show people how easy it is.

I think this section should be friendlier and simpler.

henare:
>Then the placeholder text in the comment box just says 'comment'. This is an opportunity to give a nice example to show people how easy it is.

And to encourage people to make helpful comments, as we've been discussing recently. Like Meetup.com:

![selection_001](https://cloud.githubusercontent.com/assets/48945/9839741/cef3fdc4-5ac5-11e5-80c9-e12bae054310.png)

equivalentideas:
> And to encourage people to make helpful comments, as we've been discussing recently.

:star2:

equivalentideas:
I'm gonna break this down into some tasks:

* [ ] Simplify comment intro text
* [ ] make the placeholder text more encouraging.
* [ ] Make the call to action more prominent.

I think we could also move the comment count into an comment icon on the righthand side of the comment area, so it's still clear and catchy but not taking up the prime position:

![img_4182](https://cloud.githubusercontent.com/assets/1239550/9844118/09e96a8e-5b04-11e5-8835-f112858ad436.jpg)

PR #797, opened 2015-09-30
Add labels to comment and report form inputs

equivalentideas:
Why:
  #795 explains why not having visible labels causes a number of usability problems, especially for people with limited eye sight, mobility or memory issues.

These commits:

* add labels for each input on the comment and comment report forms
* add placeholder text that gives an example of the expected input, like the signup and search forms.
* cleans up the forms scss partial a little bit e8c6a7a , 8db6c42 , f5cc1f1 &  a2d3cdd

I haven't added placeholder text on the comment or report text input because I don't think there's an obvious choise. I think we have to be particularly careful with it cause we'll be shaping the kind of comment people make. It's something we should do experiments with. I've just tried to get this to a basic win to iterate from.

This effects all users and is separate from the councillor options PR #788 . This can be merged into master before that stuff and then merged back into the `write-to-councillors` branch (which won't be automatic). I thought that was a more isolated approach as that other PR is too big. This is a general commenting improvement that relates to one of the [TODOs on the councillor options PR](https://github.com/openaustralia/planningalerts/pull/788#issuecomment-143950833) and will help step it forward.

## Before

screen shot 2015-09-30 at 3 43 35 pm
screen shot 2015-09-30 at 3 43 08 pm

## After

screen shot 2015-09-30 at 3 51 33 pm
screen shot 2015-09-30 at 3 50 44 pm

### DIFF
--- a/app/views/comments/_comment_form.html.haml
+++ b/app/views/comments/_comment_form.html.haml
@@ -13,8 +13,8 @@
     #{link_to("need to disclose this", faq_path(anchor: "disclosure"))}.
   = semantic_form_for [@application, @comment] do |f|
     = f.inputs do
-      = f.input :text, label: "What do you think about this application?",
-        hint: "Explain why you think this application should be approved or not. Please be polite, concise and to the point."
+      = f.input :text, label: "Explain why you think this application should be approved or not",
+        placeholder: "Please be polite, concise and to the point."
       = f.input :name, label: "Your name", placeholder: "e.g. Jane Citizen", hint: "Please use your real full name if possible."
       = f.input :email, label: "Your email", placeholder: "e.g. your@email.com", hint: "We never display your email address."
       - hint = capture do
--- a/app/views/comments/_comment_form.html.haml
+++ b/app/views/comments/_comment_form.html.haml
@@ -16,11 +16,11 @@
       = f.input :text, :label => "Write your comment",
         :hint => "Explain why you think this application should be approved or not. Please be polite, concise and to the point."
       = f.input :name, :label => "Your name", :hint => "Please use your real full name if possible."
-      = f.input :email, :label => "Your email", :hint => "We never display your email address"
+      = f.input :email, :label => "Your email", :placeholder => "e.g. your@email.com", :hint => "We never display your email address"
       - hint = capture do
         e.g. 1 Sowerby St, Goulburn (we never display this).
         = link_to("Why do you need my address?", faq_path(:anchor => "address"))
-      = f.input :address, :label => "Your street address", :hint => hint
+      = f.input :address, :label => "Your street address", :placeholder => "e.g. 1 Sowerby St, Goulburn, NSW 2580", :hint => hint
       %div#faq_commenting_address.inline-hints.hidden
         = render :partial => 'static/faq_commenting_address'
       -# This is a spam bot honeytrap. A human being should not be filling this in
--- a/app/views/comments/_comment_form.html.haml
+++ b/app/views/comments/_comment_form.html.haml
@@ -15,7 +15,7 @@
     = f.inputs do
       = f.input :text, :label => "Write your comment",
         :hint => "Explain why you think this application should be approved or not. Please be polite, concise and to the point."
-      = f.input :name, :label => "Your name", :hint => "Please use your real full name if possible."
+      = f.input :name, :label => "Your name", :placeholder => "e.g. Jane Citizen", :hint => "Please use your real full name if possible."
       = f.input :email, :label => "Your email", :placeholder => "e.g. your@email.com", :hint => "We never display your email address."
       - hint = capture do
         We never display your street address.
--- a/app/views/comments/_comment_form.html.haml
+++ b/app/views/comments/_comment_form.html.haml
@@ -13,14 +13,14 @@
     #{link_to("need to disclose this", faq_path(:anchor => "disclosure"))}.
   = semantic_form_for [@application, @comment] do |f|
     = f.inputs do
-      = f.input :text, :label => "Write your comment", :placeholder => "Comment",
+      = f.input :text, :label => "Write your comment",
         :hint => "Explain why you think this application should be approved or not. Please be polite, concise and to the point."
-      = f.input :name, :label => "Your name", :placeholder => "Name", :hint => "Please use your real full name if possible."
-      = f.input :email, :label => "Your email", :placeholder => "Email", :hint => "We never display your email address"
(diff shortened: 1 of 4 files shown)


henare:
Excellent! :gem: 

I really like the question on the comment report form, "*Why do you think this comment should be removed?*" because it makes it very clear what you're expected to write in there. Could we do the same for the comment label? "*What do you think about this application?*" or similar.

app/views/comments/_comment_form.html.haml
```diff
@@ -13,14 +13,14 @@
     #{link_to("need to disclose this", faq_path(:anchor => "disclosure"))}.
   = semantic_form_for [@application, @comment] do |f|
     = f.inputs do
-      = f.input :text, :label => 'Comment', :placeholder => "Comment",
+      = f.input :text, :label => "Write your comment",
```
henare: How about converting these to 1.9-style hashes as you go?

app/views/reports/new.html.haml
```diff
@@ -8,13 +8,13 @@
   you think the comment should be removed.
 = semantic_form_for [@comment, @report] do |f|
   = f.inputs do
-    = f.input :name, :label => "Your name", placeholder: "Name"
-    = f.input :email, :label => "Your email", placeholder: "Email"
-    = f.input :details, :as => :text, placeholder: "Why you think the comment should be removed"
+    = f.input :name, label: "Your name", :placeholder => "e.g. Jane Citizen"
```
henare: Use a consistent hash style.

app/views/reports/new.html.haml
```diff
@@ -8,13 +8,13 @@
...
+    = f.input :email, label: "Your email", :placeholder => "e.g. your@email.com"
+    = f.input :details, label: "Why should this comment be removed?", as: :text
     -# This is a spam bot honeytrap. A human being should not be filling this in
     -# It's not awesome for screenreaders though
     %li{style: "display: none;"}
       = label_tag :little_sweety, 'Little Sweety'
       = text_field_tag :little_sweety, "", placeholder: "Please leave this blank"
   = f.actions do
-    = f.action :submit, :label => "Send report"
+    = f.action :submit, label: "Send report"
```
henare: :ok_hand:

equivalentideas:
I've made those hash style updates.

> I really like the question on the comment report form, "Why do you think this comment should be removed?" because it makes it very clear what you're expected to write in there. Could we do the same for the comment label? "What do you think about this application?" or similar.

I like the idea. I'll play with some wording. Like the placeholder text this will have a big impact on what people write and whether they feel they even should write anything.

equivalentideas:
I quite like: "What do you think about this application?" as a first pass. It's neutral and makes the task sound approachable.

henare:
Sounds cool, let's see what it looks like :sunglasses:

equivalentideas:
> Sounds cool, let's see what it looks like :sunglasses:

screen shot 2015-10-01 at 4 57 01 pm

henare:
I'm wondering if it's too neutral now. The text below the comment box is so direct and clear, whereas this is quite understated.

Maybe we need to think more about the emotional state people are in when they're at this step and how we positively interact with that.

equivalentideas:
> I'm wondering if it's too neutral now. The text below the comment box is so direct and clear, whereas this is quite understated.

I think that text underneath is a good balance of neutral and direct too. What if we made "Explain why you think this application should be approved or not." label and "Please be polite, concise and to the point." placeholder. That gives people something clear to do, that will help them make an impactful comment, and also prompts them to be nice:

screen shot 2015-10-01 at 5 25 13 pm

equivalentideas:
> Maybe we need to think more about the emotional state people are in when they're at this step and how we positively interact with that.

Something I like about ”Explain why you think this application should be approved or not”, is that it gives you something effective with all the thoughts you're having about the DA you're looking at—explain them. You could be angry or happy and it helps you either way. Any thoughts @henare ?

henare:
I think the new placeholder text is good :+1: It does this https://github.com/openaustralia/planningalerts/issues/766#issuecomment-139929095

I thought of changing the label to a question, "Why do you think this application should be approved or not?" but I can just imaging people starting their comments with, "Because I..." which would be an odd comment by itself :)

The label feels a bit wordy and not quite friendly in this context so I think it need more refinement.

equivalentideas:
> The label feels a bit wordy and not quite friendly in this context so I think it need more refinement.

Agreed. I'm gonna pick this up now.

equivalentideas:
@henare I've done some refining on the comment body label and the placeholder too. I've done quite detailed commit messages on the reasoning 5b97f3e , ee5dcb2 and  67cafdb . I'm assigning this to you for review again. Thanks for all the great help on this so far too :+1:

screen shot 2015-10-07 at 3 52 07 pm
screen shot 2015-10-07 at 3 51 53 pm

equivalentideas:
>  I think we should focus on the benefit to the person making the comment if they follow those suggestions.

We think the if people are *polite, clear and to the point* then their comment will have a greater influence on the decision making process. This is because its points will be more clearly and quickly understood and _be taken more seriously_.

In the commenter's survey responses there's a clear pattern that people are concerned about whether their comment makes an impact:

> I have no real way of knowing whether my concerns are given any attention by local council.

> I have no idea if the comments will be listened to or what impact they will have if any

> I believe that the [council] are going to go ahead and develop, come what may. However, if I and others don't comment/object we will be seen as providing tacit approval to Council's actions 

> I don't feel I have any influence.  I was just sharing my observations, or thoughts with like minded people who may. (have influence)

> I do get the 'Form Letter' from Council but I'm not in any way convinced they listen.

> The process of being alerted and expressing an opinion works well but whether it has any effect is doubtful.

People want their comments to “be listened to”. We think this is more likely to happen if their comments are polite, clear and succinct than if they're rude, confused and rabbling.

So @henare what do you think about:

> Be polite, clear, and to the point so your comment gets listened to

I think that's actually helpful to people and focuses on the benefit they're looking for.

henare:
@equivalentideas bang on. Let's give it a go!

equivalentideas:
> @equivalentideas bang on. Let's give it a go!

Cool. I think that's a good place to start from. I'll implement this and then I'll assign back to you to review the code.

equivalentideas:
I've implemented that change to the placeholder text:
screen shot 2015-10-15 at 12 34 02 pm

equivalentideas:
Assigning to @henare for code review

henare:
LGTM :leaves: I'll leave it (boomtish) for you to deploy.

Issue #798, opened 2015-10-07
As someone interested in a local development, show me the most recent comments on an application first, so I get the most recent information from people and see that other locals have commented recently


equivalentideas:
At the moment, if you land on a DA with a number of comments, you see that oldest comment first. The older the DA, the more this gives the impressions that comments are infrequent—I think “huh, no one has commented in 2 weeks”:

![screen shot 2015-10-07 at 12 04 55 pm](https://cloud.githubusercontent.com/assets/1239550/10326903/79696bd0-6cec-11e5-9974-967ba5cccc87.png)

Then you go all the way to the bottom and see, actually another local has commented 5 minutes ago. The sense that people are acting around you now encourages you to take action.

The order of the comments is now much more noticeable now that the time is a more prominent part of the comment item (moved in #792).

Sometimes there are more recent comments that provide helpful, more up-to-date information about the application as well. For example, a comment explaining that an application has been withdrawn. These are easy to miss when they are way down the page.

I think we should reverse the comment order on the application page to show more recent comments first.

henare:
>I think we should reverse the comment order on the application page to show more recent comments first.

As per https://github.com/openaustralia/planningalerts/pull/800#issuecomment-146403706, this does not take into account important ramifications of making this change. Because we've discussed these ramifications before and they're not included in this issue I'm closing it as incomplete.

Please reopen a new issue that either only reports the specific problem or, if it proposes a solution, includes complete details about the design ramifications of making this change and their mitigation. Specifically how is the conceptual confusion, of adding your comment to the bottom of the comment list but having it appear that the top, addressed?

---

Some other observations about design priorities:

>At the moment, if you land on a DA with a number of comments...

It's worth noting that this is a rare occurrence - the vast majority of DAs have no comments or very few comments.

>...The older the DA, the more this gives the impressions that comments are infrequent—I think “huh, no one has commented in 2 weeks”:

If no one has commented for 2 weeks then there's a good chance the commenting period is over. We don't want to design for a situation that we don't want to encourage, i.e. people commenting on expired DAs meaning their comment is more likely to be ignored.

>Sometimes there are more recent comments that provide helpful, more up-to-date information about the application as well. For example, a comment explaining that an application has been withdrawn. These are easy to miss when they are way down the page.

This is an exceedingly rare situation and one that our commenting system is not designed around. Changing the layout to suit a situation that almost never happens and one that we've not given much design thought to does not seem like a good idea to me.

PR #800, opened 2015-10-07
Order comments to show most recent one first

equivalentideas:
See #798 "As someone interested in a local development, show me the
most recent comments on an application first, so I get the most recent
information from people and see that other locals have commented
recently"

### DIFF
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -101,7 +101,7 @@ class ApplicationsController < ApplicationController
     end
 
     @application = Application.find(params[:id])
-    @comments = @application.comments.visible.order(:updated_at)
+    @comments = @application.comments.visible.order("updated_at DESC")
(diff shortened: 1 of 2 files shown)


henare:
Luuuke, you know better! :broken_heart:

Just flipping the comment order means that the comment submission form is at the wrong end of the comments list. It's a conceptually logical design to add comments to the end of the list with the most recent comment and it's almost exclusively the design pattern for comments all over the web. 

The most perplexing thing is that we've discussed this exact change a number of times - most recently just a few days ago. We discussed how we were starting to get hints that reversing the comment order might be a good idea **but** that change would necessitate moving the comment form too. And by moving the comment form we'd face another problem - because you wouldn't see the comments first, the far more critical social validation to make a comment wouldn't be there.

Every time we've had the discussion we've come to the conclusion that this simple change will require a smart design solution. Nowhere in this or #798 have you made a case that this has changed so I'm closing this.

Issue #804, opened 2015-10-09
Show the report comment link on the comments index


equivalentideas:
Currently we don't show the comment report link on the comments index.

I think that, if you see a comment that needs to be reported, you should be able to see that reporting it is an option and take action, no matter where you are in the site. So, we shouldn't hide the 'report comment' link on the comments index.

henare:
:+1:

PR #805, opened 2015-10-09
Keeping the right comment actions available in the right places

### LINKED ISSUE #803: Don't show report comment link on comment report page
selection_001

equivalentideas:
See #804 and #803 issues about which comment actions should be shown on which pages.

fcbab1a hides the 'report comment' link if you're already on the report page. Fixes #803.

258c577 makes the report and admin links available on the comments index so you can immediately take action if you see a comment that needs it.


equivalentideas:

henare:
:snowflake: thanks @equivalentideas

PR #816, opened 2015-10-19
Halve the hight of the comment text input textarea

### LINKED ISSUE #815: Comment input field is too tall
The size of an input field should default to the intended length of the text to be inputed. By my quick look, most comments are either roughly half the length of the text fields default size, or are much longer. The textarea can be stretched to be taller by the user.

screen shot 2015-10-19 at 11 59 31 am

By making it bigger we're making the form longer, and seem more involved than it needs to be. We're also suggesting a length that most people don't seem to think is necessary. This is in conflict with our direction that people be 'to the point' in writing their comments.

Making it shorter will make the form simpler and appear a bit more approachable and help people scan the inputs—making the form a little more usable.

equivalentideas:
From #815:

> The size of an input field should default to the intended length of the
text to be inputed. By my quick look, most comments are either roughly
half the length of the text fields default size, or are much longer. The
textarea can be stretched to be taller by the user.

> By making it bigger we're making the form longer, and seem more involved
than it needs to be. We're also suggesting a length that most people
don't seem to think is necessary. This is in conflict with our direction
that people be 'to the point' in writing their comments.

> Making it shorter will make the form simpler and appear a bit more
approachable and help people scan the inputs—making the form a little
more usable.

## Before
screen shot 2015-10-19 at 1 44 01 pm

## After
screen shot 2015-10-19 at 1 43 29 pm

Fixes #815

### DIFF
--- a/app/views/comments/_comment_form_inputs.html.haml
+++ b/app/views/comments/_comment_form_inputs.html.haml
@@ -7,7 +7,8 @@
   = f.inputs id: "comment-base-inputgroup" do
     = f.input :text, label: "Have your say on this application",
                      placeholder: "Be polite, clear, and to the point so your comment gets listened to.",
-                     hint: "Have you made a donation or gift to a Councillor or Council employee? #{link_to("You may need to disclose this", faq_path(anchor: "disclosure"))}.".html_safe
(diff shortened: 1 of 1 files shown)


equivalentideas:
@henare I just whipped this up before lunch, could you have a quick look/merge if you think this works.

henare:
:apple: Over to you to deploy.

Issue #782, opened 2015-09-21
Improve conversion on the subscriptions page


equivalentideas:
Looking at Google Analytics, In the last month we've had about 70 people hit the subscription page via an email alert (mostly from people in their trial period). We've only recorded 2 clicks on the subscribe button on that page—one of those clicks resulted in a subscription. 19 people have requested free non-commercial subscriptions.

So people are clicking the subscribe links in their email, probably to get more information, and then not clicking the subscribe button on the page. What changes could we experiment with on the subscription page to convince more people to subscribe?

henare:
Very interesting data! Your idea to de-emphasise the non-commercial subscription could be a simple iteration to test.

henare:
We're going to remove this page now we're not doing the paid feature.

PR #822, opened 2015-10-21
De-emphasise non-commercial subscription section so the subscription page is simpler

henare:
Make the subscription page simpler by pushing the information about non-commercial subscription down the page and de-emphasising it.

Mentioned as something to try on #782.

## Before
selection_002
selection_003

## After
selection_001
selection_004

### DIFF
--- a/app/assets/stylesheets/partials/special_forms/_payment_form.scss
+++ b/app/assets/stylesheets/partials/special_forms/_payment_form.scss
@@ -16,6 +16,9 @@
       font-size: 1.175em;
     }
   }
+  .small {
+    font-size: 0.8em;
+  }
 }
 
 p.form-processing {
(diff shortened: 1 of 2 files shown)


henare:
Assigned to @equivalentideas for review/deploy.

equivalentideas:
@henare looks pretty straight forward :+1:

Does the text really need to be that small to be deemphasised? I think it's bordering on “fine print”—a.k.a 'not intended to be seen'.

henare:
>Does the text really need to be that small to be deemphasised? I think it's bordering on “fine print”—a.k.a 'not intended to be seen'.

That's a good call. I just did the dumb thing and set it to whatever `.small` was elsewhere. I think I'll bump it up before merging and deploying.

PR #825, opened 2015-10-23
[WIP] Change subject line of alerts for people with an expired subscription

henare:
@equivalentideas I mentioned this change on Trello. What do you think about the change generally? Any ideas for the wording?

I also refactored the logic for the subject out of the view an into a helper. It's not my finest work but I think it's a little clearer.


henare:
I guess it might have been helpful to say that all this does is change the subject from:

>1 new planning application near 123 Sample St

To this if you have an expired subscription:

>You're missing out on: 1 new planning application near 123 Sample St

app/helpers/alert_notifier_helper.rb
```diff
@@ -60,4 +60,16 @@ def new_subscription_url_with_tracking(alert: nil, utm_content: nil)
 
     new_subscription_url(params)
   end
+
+  def subject(alert, applications, comments)
+    items = if !applications.empty? && comments.empty?
```
equivalentideas: I think this might be a bit clearer if you use the Enumerable method .any? instead of ![stuff].empty . I found this quite slow to parse what each line was asking.

Also, it could help to put the conditionals indented under the variable assignment to show that it's all inside that, [like this example](https://github.com/bbatsov/ruby-style-guide#use-if-case-returns).

app/helpers/alert_notifier_helper.rb
```diff
@@ -60,4 +60,16 @@ def new_subscription_url_with_tracking(alert: nil, utm_content: nil)
...
+      pluralize(applications.size, "new planning application")
+    elsif applications.empty? && !comments.empty?
+      pluralize(comments.size, "new comment") + " on planning applications"
+    elsif !applications.empty? && !comments.empty?
+      pluralize(comments.size, "new comment") + " and " + pluralize(applications.size, "new planning application")
+    end
+
+    "#{"You're missing out on: " if alert.expired_subscription?}#{items} near #{alert.address}"
```
equivalentideas: While this is succinct on one line, the most common case that people will see when there's no subscription is squashed into the far conner. Even though there's repetition, would something like this be simpler:

```
if alert.expired_subscription?
  "You’re missing out on #{items} near #{alert.address}"
else
  "#{items} near #{alert.address}"
end
```

You could even take out the repetition with a `sentence_base` variable or something. I think that's a bit easier to read, but if you think this is clearer :+1:

spec/helpers/alert_notifier_helper_spec.rb
```diff
@@ -147,4 +147,52 @@
...
+    end
+    let(:comment) { create(:comment, application: application) }
+
+    context "with an application" do
+      subject { helper.subject(alert, [application], []) }
+      it { should eql "1 new planning application near 123 Sample St" }
+    end
+
+    context "with a comment" do
+      subject { helper.subject(alert, [], [comment]) }
```
equivalentideas: I think you could consider using [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments) on this method, at the moment you need to look around at other tests to understand what could have been in the empty array.

spec/helpers/alert_notifier_helper_spec.rb
```diff
@@ -147,4 +147,52 @@
       end
     end
   end
+
+  describe "#subject" do
+    let(:alert) { create(:alert, address: "123 Sample St") }
+    let(:application) do
+      mock_model(Application, address: "Bar Street",
+      description: "Alterations & additions", council_reference: "007",
```
equivalentideas: I think this multi line method call would be clearer with the parameters indented. [From the Ruby style guide](https://github.com/bbatsov/ruby-style-guide#no-double-indent):

>  Align the parameters of a method call if they span more than one line. When aligning parameters is not appropriate due to line-length constraints, single indent for the lines after the first is also acceptable.

equivalentideas:
I think this is a good test and will be interesting to see the results :telescope: . I made a comment in trello as well about this being a different test to the one on that card.

I think we should drop the colon because it breaks up what is otherwise a quite natural sentence:

> You're missing out on 1 new planning application near 123 Sample St

I think it's really good to get the subject out of the view and get some nice tests for it :+1: I made a few style comments.

One thing: there's no test for the case of there being no applications or comment. I assume this is because it would error, but shouldn't there be a test to verify that?

app/helpers/alert_notifier_helper.rb
henare: Both great suggestions :+1:

spec/helpers/alert_notifier_helper_spec.rb
henare: >Ruby 2.0 introduced first-class support for keyword arguments...

Wow, I had no idea! That's super cool. I've still been using options hashes.

I don't think I'm going to do this as part of this PR though. Mainly because:

>Unfortunately, Ruby 2.0 doesn’t have built-in support for _required_ keyword arguments.

It can be a cool improvement for someone to make down the line :)

spec/helpers/alert_notifier_helper_spec.rb
henare: :hurtrealbad: Nice spotting all my laziness :)

henare:
>I think this is a good test and will be interesting to see the results :telescope: . I made a comment in trello as well about this being a different test to the one on that card.

I've added a new card in Trello.

>I think we should drop the colon because it breaks up what is otherwise a quite natural sentence:
>
>    You're missing out on 1 new planning application near 123 Sample St

Absolutely!

>I think it's really good to get the subject out of the view and get some nice tests for it :+1: I made a few style comments.

Great comments, you're a good tough reviewer :smile: 

>One thing: there's no test for the case of there being no applications or comments. I assume this is because it would error, but shouldn't there be a test to verify that?

I think it would just be a blank subject. This PR doesn't change that and it wasn't tested before so I'll leave it for a future improvement.

Issue #827, opened 2015-10-27
The send to coucillor option is broken in Safari


equivalentideas:
The second radio option “One of your elected local councillors
Your elected councillors can answer your questions and represent you in council meetings.” does not display:
![screen shot 2015-10-27 at 12 00 35 pm](https://cloud.githubusercontent.com/assets/1239550/10746797/999d56c6-7ca2-11e5-817c-2f1ee7c52912.png)

should display like:
![screen shot 2015-10-27 at 12 03 56 pm](https://cloud.githubusercontent.com/assets/1239550/10746850/cd5bc7f4-7ca2-11e5-9709-7086590d3fde.png)

equivalentideas:
This is now deployed and fixed in Safari, still works in Chrome and Firefox.

PR #828, opened 2015-10-27
Fix bug in councillor form and improve js performance

equivalentideas:
Why:

See bug #827 a really bad bug in Safari where the label for the councillors option wouldn't show.

These commits:

1. e8194be Includes `applications.js` into the compliation for the site wide `application.js` to reduce http requests. Also if the site wide file is already downloaded, but some network error stops the applications.js being downloaded we wouldn't get all the useful stuff in that file. In other words this reduces number of things that could go wrong.
2.  78211a0 Fixes the bug by adding a missing "]"

Safari before:
screen shot 2015-10-27 at 12 00 35 pm

Safari after:
screen shot 2015-10-27 at 2 20 18 pm

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -3,6 +3,7 @@
 //= require jquery.ui.autocomplete.html.js
 //= require address_autocomplete.js
 //= require geolocation
+//= require applications
 
 $("#menu .toggle").click(function(){
   $("#menu ul").slideToggle("fast", function(){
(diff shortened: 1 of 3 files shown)


equivalentideas:
Assigned to @henare for review

henare:
@equivalentideas this looks pretty straightforward. If you're confident it's been tested please go ahead and merge and deploy.

equivalentideas:
Thanks @henare I'll :shipit:

Issue #817, opened 2015-10-19
Comment display in email is inconsistent with the website


henare:
On the website comments have the name of the commenter first now but that's different to the email alert. Shouldn't they be consistent?

![selection_001](https://cloud.githubusercontent.com/assets/48945/10571709/6646e9a6-768d-11e5-9e71-50133d037ec7.png)

vs.

![selection_002](https://cloud.githubusercontent.com/assets/48945/10571715/73fe48be-768d-11e5-8901-ea8242e45c8b.png)

equivalentideas:
> Shouldn't they be consistent?

Yep I think so :+1:

Issue #799, opened 2015-10-07
As someone interested in a local development, when someone writes to one of the relevant local councillors about it, show me their comment and the councillor’s response, so I feel more comfortable about writing to my councillors too


equivalentideas:
When someone writes to their councillor about an application, show what they've written in the comment feed with the other comments. When the councillor replies, show their response there as well.

This is to show people how easy and beneficial it can be to start a dialogue with your local councillors.

As a secondary benefit, this will also allow people to hold their elected representatives in the council more responsible for the ways they handle feedback and requests from the people they represent.

Also, councillors might have something helpful to say about a development.

Putting these in the current comment feed is a simple extension of the current system as a first pass.

* [ ] Get this working well in the NSW Theme

equivalentideas:
Here's a mock up of one way to show the initial comment to a councillor:
![screen shot 2015-10-20 at 1 30 15 pm](https://cloud.githubusercontent.com/assets/1239550/10596315/be2a1048-772e-11e5-87e1-ea430d1f4861.png)

Using 'wrote to' to describe the action people take, a human and common way to talk about contacting representatives.

I don't think people will recognise their councillors by name, so I think we do need to provide them a label/title. I don't think people need to know which council the person is in, just that they are a councillor local to the area that this planning application is in. I think 'local councillor' is a place to start that achieves this.

It would be great to show an image of the councillor so people can get a better sense of the person writing, and better empathise with them. I did a sketch with the councillor's image with their name in the initial comment to them, but glance at this and think this comment is by the councillor not to them:

![img_4230](https://cloud.githubusercontent.com/assets/1239550/10596447/1155b4ba-7730-11e5-8943-c1ca457a547e.jpg)

I think the image should be saved for the councillor’s reply so you still get to see them, but their image is simple connected to their response.

I've got some ideas so I'm gonna jump in and make something in the browser to look at.

equivalentideas:
Henare and I have been chatting and reworked the name/story of this issue and the description above.

We want people to understand that they can write to their councillors through PlanningAlerts and that it's really easy. They need to be able to see messages to councillors as distinct from normal comments, to understand that they have this option. They should also understand that that both options can benefit them.

While we want to make the comments distinct we also don't want to make one clearly more approachable than the other or make one seem more special at this stage.

The difference between the types of comments is:

* You make a formal submission on the application to the planning authority when you want to impact the decision on the application.
[...]

PR #831, opened 2015-10-28
Make comments in emails consistent with comments in the site

equivalentideas:
This addresses #817 to make the display of comments in emails consistent with their look in the site. This also includes some refactoring ahead of coming changes to the way comments are displayed everywhere ( see #799 ).

On the site we say how long ago this comment was made, but that's not useful in the email alert because it's always going to be about a day ago. Instead I opted to keep the info of the application that this comment was posted on, and changed this to display like on the /comments page with an excerpt of the description of the application description. The other nice thing about this is that you get some helpful context about what the comment is referring to before reading it.

I kept the link to view the comment because we can see in our analytics that people click this link and create comments.

## Before:

selection_001

## After:

screen shot 2015-10-28 at 5 08 21 pm


henare:
This all sounds good to me :cherries:

PR #832, opened 2015-10-30
Increase textarea input line-height for readability

equivalentideas:
The line-height of textarea input is too narrow, making it harder
to check what you've written.

This increases it to match the line-height of text in the rest of
the site.

I noticed this problem on a user's computer during testing.

## Before:
screen shot 2015-10-30 at 11 35 06 am

## After:
screen shot 2015-10-30 at 11 36 34 am


henare:
Really nice subtle improvement :+1:

PR #833, opened 2015-10-30
Increase the contrast of input text

equivalentideas:
The contrast of input text is reduced, making it slightly softer
and less clear.

There’s no obvious reason for this, and it incrementally makes using
the forms for of a task for some users.

I noticed this in usability testing on where a person was squinting
as they checked the comment they had inputed.

## Before

screen shot 2015-10-30 at 11 43 33 am
screen shot 2015-10-30 at 11 43 25 am

## After

screen shot 2015-10-30 at 11 38 39 am
screen shot 2015-10-30 at 11 38 13 am


henare:
Looking at your screenshots I agree with your comments and can't see a reason not to merge this :notes:

PR #835, opened 2015-11-01
Use generic label for planning authority option because many people wont recognise authority by name

### LINKED ISSUE #829: I shouldn't have to know who the planning authority is to make a comment
The form for writing to councillors currently asks you to pick between 'name of planning authority' and 'one of your elected local councillors'.

screen shot 2015-10-27 at 12 03 56 pm

What if you don't recognise the name of the planning authority?

Also, it says this the council, but what if it isn't?
screen shot 2015-10-27 at 2 37 21 pm

This could potentially be very confusing.

Maybe we could use a more generic but accurate term, like “The planning authority”?

equivalentideas:
People shouldn’t have to know the name of the planning authority to
decide whether or not they want to send a comment to them.

Throughout PlanningAlerts we use the generic term 'planning authority'
to refer to these institutions.

Instead of using the authority name for the option label,
name the authority in the label description text so that
people are still informed of precisely who their message will go to.
This should be done in addition to repeating the
term 'planning authority' to make sure people make that connection:

> Write to the planning authority (Rockdale City Council) ...

## Before
screen shot 2015-11-01 at 1 58 24 pm

## After
screen shot 2015-11-01 at 1 58 07 pm

Fixes #829

### DIFF
--- a/app/views/comments/_comment_receiver_options.html.haml
+++ b/app/views/comments/_comment_receiver_options.html.haml
@@ -2,11 +2,11 @@
   %legend Who should this go to?
   %input.receiver-select-radio.receiver-type-option{id: "receiver-to-authority-option", type: "radio", name: "receiver"}
   %label.receiver-select-label{for: "receiver-to-authority-option"}
-    %strong= @application.authority.full_name
+    %strong The planning authority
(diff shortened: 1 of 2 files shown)


henare:
Nice.

Issue #818, opened 2015-10-20
How would someone decided whether to write to a planning authority or write to an elected councillor?


equivalentideas:
You can make an official submission on a development application, or you can write to your local councillor about it. We think that someone’s motivation to take either of these steps is to impact the decision on a development application. But sometimes people do one, and sometimes they do the other.

Presented with the two options, how do you decide which way to go? I don't think we’ve worked out yet what people are thinking when they reach for one option over the other.

I think if we can understand this and give people a really clear explanation of why they would do one or the other then that will really help them act intentionally and more effectively.

I'm going to ask users about this in #770 

Currently we say:

* “Write to the council if you want your comment considered when they decide whether to approve this application.”
* ”Your elected councillors can answer your questions and represent you in council meetings.“

![screen shot 2015-10-20 at 4 28 21 pm](https://cloud.githubusercontent.com/assets/1239550/10598770/9f3ae1d0-7747-11e5-97eb-c0971ccdcd04.png)

These explanations should be more from the users perspective. What's the benefit to them of going one way or the other?

Personally, I would use the official intended path of making a submission to the planning authority, unless I thought that would have no impact. For example, if I was objecting to a big business development that appeared to comply with regulations, but I thought it would impact my area negatively. I'd want a councillor to represent me because I assume the developers have more resources than me and are more familiar with the process. Or if I wanted to highlight an application I supported and say 'we could support developments like this as an area'.

I think I would make an official submission if I thought there was a violation of the planning rules, or if I thought I could impact the decision without having a councillor represent me. So for smaller scale developments.

Current PlanningAlerts commenters are pretty smart and I'm sure they'll work this out when their presented with the options. But to make this process approachable to everyone I think we need to understand and communicate this better.

@kat Have you got any thoughts on this?

equivalentideas:
I just tested the current form text with a someone.

They initially chose to send their message to the planning authority:

> I want them to consider ... I don’t really have a question, and I want them to consider... I don’t understand the difference between giving it to the planning authority and giving it to one of your local councillors who represent you in the council meetings. What different thing happens?

Then they decided to send it to their councillor instead.

> “The councillor definitely need to take [your comment] into account and represent you in a Council meeting, where as the planning authority probably just [throw it way].”

[...]

PR #836, opened 2015-11-02
Cut duplicate comment heading & add notice about comment options before first input

### LINKED ISSUE #830: Tell me about my commenting options before I write my comment
Currently, unless you see the options about who you'll send you comment to, you’re likely to make at least small changes to your comment if you decide to write to a councillor. This is because the options of who to send it to come as a surprise.

We originally moved these options down in the form to let people get straight into commenting, because in user testing the person got stuck deciding who to write to before they had thought about what they wanted to say. I think that’s still a good choice, but we could could some kind of simple heads up/signpost, at the top of the form. I think this would be particularly helpful for new commenters.

equivalentideas:
The label or the comment text input is saying the same thing as the heading to the comment section. This adds unnecessary text for people to read and takes up space. It’s simpler if there is only one heading here. This makes the comment heading and into into the label and replaces the heading text with the text from the original label 'Have your say on this application'.

This also:
* e708fb3 & 76425a9 adds an introduction par to the comment for with councillors, to give some notice that you can choose who you send your comment to; and
* 87f5540 Removes the 'public comment' notice 'Your name and comment will be posted publicly here as well' from the intro text and makes the button read 'post your *public* comment' for the form without councillors. This makes the way both versions of the form handle this consistent.

There’s lots of explanation on the commit messages also.

TODO’s:
* [x] improve the text intro to the comment form with councillors 'Write to the planning authority or one of your elected local councillors'. I think it could be a bit more specific to re-enforce the concepts behind the different kinds of comments. Something like 'Post a comment for the planning authority or write to one of your elected local councillors'

## Before

screen shot 2015-11-02 at 4 53 31 pm
screen shot 2015-11-02 at 4 53 01 pm
screen shot 2015-11-02 at 4 52 04 pm
screen shot 2015-11-02 at 4 53 48 pm

## After
screen shot 2015-11-02 at 4 51 09 pm
screen shot 2015-11-02 at 4 54 40 pm
screen shot 2015-11-02 at 4 54 27 pm

Fixes #830

### DIFF
--- a/app/assets/stylesheets/partials/_form.scss
+++ b/app/assets/stylesheets/partials/_form.scss
@@ -71,14 +71,13 @@ a.btn:hover {
     color: $error-color;
   }
 
-  li.string,
-  li.stringish,
-  li.email,
-  li.text {
-    label {
-      font-weight: 500;
-      font-size: 1.1em;
-    }
+  label {
+    font-weight: 500;
+    font-size: 1.1em;
+  }
+
+  label p {
+    font-weight: normal;
   }
 
   li.string, li.stringish, li.email {
--- a/app/assets/stylesheets/partials/special_forms/_comment_form.scss
+++ b/app/assets/stylesheets/partials/special_forms/_comment_form.scss
@@ -43,6 +43,16 @@
   }
 }
 
+#comment-base-inputgroup {
+  h4 {
+    margin: 0;
+  }
+
+  label p {
+    margin: .5em 0 !important;
+  }
+}
+
 #comment-receiver-inputgroup {
   margin-top: 1em !important;
 
@@ -111,8 +121,8 @@
     display: block;
   }
 
-  .p-name {
-    font-weight: bold;
+  .p-org {
+    font-weight: normal;
   }
 
   :hover + &,
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -37,12 +37,12 @@ module ApplicationsHelper
 
   def on_notice_text(application)
     if application.on_notice_from && (Date.today < application.on_notice_from)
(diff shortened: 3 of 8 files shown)


equivalentideas:
I've updated that text now, I think it’s a bit more helpful ( see 76425a9 ). I think it could still be more focuses on the different benefit to the user from each choice of who to the send the comment to (the same goes for the label text for each option), but we’re not sure how best to describe this yet (see #818 ). I think this definitely a step forward though :walking: .

screen shot 2015-11-04 at 9 28 48 am

equivalentideas:

equivalentideas:
@henare and I just had a quick chat about the wording 'write to your local councillor'. @henare is concerned that people will assume this is a private correspondence like a letter. We are leaning on that analogy pretty hard here.

What about being really explicit with 'post a public message' instead of 'write to':

> 'Post a comment for the planning authority or a public message to one of your local councillors.'

That’s quite straight forward. It could be annoyingly a bit long to fit on a single line on wide screens. I'll see what it looks like.

henare:
Why not just keep it as comment?

>'Post a comment for the planning authority or one of your local councillors.'

henare:
Cool, this looks like a pretty straightforward and sensible change for the existing comment system.

equivalentideas:
Why not just keep it as comment?

> 'Post a comment for the planning authority or one of your local councillors.'

That looks good to me @henare , I'll see how it works implemented.

equivalentideas:
I think this works @henare but we lose the distinction between the types of comments. Particularly with this 'official response' text above, what if I think I'm making an official response, but I decide to write to a councillor. This is hard.

screen shot 2015-11-04 at 2 39 05 pm

equivalentideas:
We just added back a slightly simplified version of the privacy text into the intro in b1c4eb6:

screen shot 2015-11-04 at 3 53 41 pm

equivalentideas:
We’re happy with this so I'm merging

Issue #769, opened 2015-09-14
As someone interested in local development, when I'm viewing comments on a DA, provide me a link to the comment I can share, so I can promote and continue discussion in my networks


equivalentideas:
We should help people to take the discussion of a DA to more people by sharing comments from the comment list. While it is possible to link to comments with their ids, we don't currently expose the links in any way.

Here are some examples of how other discussion platforms provide this:

![screen shot 2015-09-14 at 11 17 58 am](https://cloud.githubusercontent.com/assets/1239550/9840254/88a636d6-5ad2-11e5-9dad-6b3e6f1d9b3b.png)
![screen shot 2015-09-14 at 11 17 16 am](https://cloud.githubusercontent.com/assets/1239550/9840255/88aa4e7e-5ad2-11e5-87f8-4295aa0120f1.png)
![screen shot 2015-09-14 at 11 16 59 am](https://cloud.githubusercontent.com/assets/1239550/9840256/88ac8f5e-5ad2-11e5-89a9-97b8bc1cae30.png)
![screen shot 2015-09-14 at 11 22 03 am](https://cloud.githubusercontent.com/assets/1239550/9840271/dee5ee1a-5ad2-11e5-9cd0-7cc546c17406.png)

The conventional pattern is make the time stamp of the post a link.

Facebook also have a share button as part of the post module, and Discourse provides links for sharing on social networks specifically once you click the link icon:

![screen shot 2015-09-14 at 11 18 42 am](https://cloud.githubusercontent.com/assets/1239550/9840289/2ab33528-5ad3-11e5-8566-9bff1c7906d1.png)

equivalentideas:
I think the simple step to do this is to make our timestamp on the comment a link to the comment:

![screen shot 2015-09-14 at 11 28 02 am](https://cloud.githubusercontent.com/assets/1239550/9840335/c0a782c8-5ad3-11e5-975d-32989ba29976.png)

equivalentideas:
[The commit that closed this was reverted](https://github.com/openaustralia/planningalerts/commit/8dba983d6e915a8864b4adffe8b53487b1cb376d) so I'm reopening this.

I reverted it because we currently make the whole comment a link in the comments index. Wrapping a link in a link is invalid markup and creates an inconsistent experience in Safari, Firefox and Chrome: if there is no link in the comment the whole comment is clickable; if there is, just the heading is clickable and the link in the body is also clickable. This is an issue when a commenter adds a link or when we add a link to the time.

[...]

Issue #679, opened 2015-06-26
Switch JSON API responses to JSON format


henare:
They're currently JS which is pretty harmless but a little confusing (and confuses Rails a bit too). We'll have to put in redirects or something so it doesn't affect current API users.

mlandauer:
I keep forgetting that this was an early mistake!

When we do this we should update the documentation but still support the old extension (without redirects). When we're confident that nobody is using the old way we switch it off

mlandauer:
We're still supporting the old js extension (without redirection) for the time being. We should wait a few months (until 1st November 2021) see if the main users are still using the js extension. If so, we contact them asking them to switch with a plan to setup automatic redirects a month after that. Then, we can delete the unused code. When that's all done, then we can close this issue :-)

mlandauer:
As of now (17 Sep 2024) the majority of API calls are still being done with the `js` extension. That's because they are for commercial users who haven't changed anything at their end. It doesn't feel like important enough to warrant the effort of getting them to change. Let's just hope the current plan of changing the documentation and waiting for the old users to fall off will work at some point in the future.

PR #838, opened 2015-11-03
Add static prototypes for comments to and replies from councillors and differentiate them from submission comments

equivalentideas:
Why:

In order to feel like they can write to their local councillor, people need to recognise this is an option. If they see other people writing to councillors then it will seem more approachable.

These commits are about designing how message to and replies from councillors will appear, and making sure they are recognisably different from official submission comments without straying too far becoming confusing.

The messages to councillors and replies are static prototypes that aren’t visible to users without special params. The current comments are updated in preparation for this change, and I think generally look more human and work better with the reset of the site.

This PR:

* Is too big, I'm sorry :hurtrealbad: 
* 6779eea Adds a static prototype of a comment to a councillor that you can see on an application page if you add `with_councillor_message=true` to the url query string.
* 7316a29 Adds a static prototype of a reply from a councillor to that comment that you can see on an application page if you add `with_councillor_reply=true` to the url query string.
* Uses colour, text style and wording to differentiate messages to councillors and their replies from comments that are official submissions.
    * 68c5e1e Messages with councillors get a sunny yellow and official submission comments get a more formal blue
    * db90d5c The names of the people involved in the comment are bold, to help people quickly recognise whether this is a one way comment or an exchange. This reverses an earlier decision not to highlight these names like this, see discussion on the commit message.
    * 19fcb61 A footer panel is added to messages to councillors that are awaiting replies, with the text: “Delivered to Melissa Brooks about 5 hours ago. They are yet to respond.”
    * 8689064 The corners of all comments are rounded a little to make them more human and conform to the general site style better.
    * 83ab513 Adds a subtle footer to 'official submission' comments restating when they were posted and where they were sent, to re-enforce that these comments are official.
* Updates the comment style shown in the comment index for these changes.
* Removes the link wrapping comments in the index as it was buggy for starters (see [discussion on #769](https://github.com/openaustralia/planningalerts/issues/769#issuecomment-146047188))and I think I've replaced it with something that’s actually clearer. There’s now an obvious link to the application that the comment is on (which is what the wrapping link pointed to) on the address. There are also links to the comment itself now, on the time of the comment. This is more of a power user/admin link that people may discover, not a solution to #679 .

### Before
screen shot 2015-11-03 at 8 43 04 pm

### After
screen shot 2015-11-03 at 8 38 00 pm

screen shot 2015-11-03 at 8 41 30 pm

### Before
screen shot 2015-11-03 at 8 43 19 pm

### After

screen shot 2015-11-03 at 8 38 32 pm


equivalentideas:
Merged master into this branch to update it for the changes from other PRs.

app/views/comments/_comment.html.haml
```diff
@@ -1,9 +1,4 @@
 - if in_list
-  %li
-    - if with_link_wrapper
-      = link_to application_path(comment.application), title: "View application at #{comment.application.address}", class: "panel-link" do
-        = render 'comments/comment_item', comment: comment, with_address: true
-    - else
-      = render 'comments/comment_item', comment: comment, with_address: false
+  %li= render 'comments/comment_item', comment: comment, with_address: (with_address ? true : false)
```
henare: Does this need this ternary true/false statement if `with_address` already evaluates true or false?

app/controllers/applications_controller.rb
```diff
@@ -122,6 +122,24 @@ def show
                                             { name: "Daniel Barbar", party: "Labor" }]
     end
 
+    if params[:with_councillor_message] == "true" || params[:with_councillor_reply] == "true"
+      @message_to_councillor = { name: "Laurie Burdekin",
+                                 receiver: "Melissa Brooks",
```
henare: Maybe recipient instead of receiver?

equivalentideas: Nice, I agree :+1: I'll make that change

spec/helpers/comments_helper_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'spec_helper'
+
+describe CommentsHelper do
+  describe "#comment_path" do
+    let(:comment) { VCR.use_cassette('planningalerts') { create(:confirmed_comment) } }
+
+    it "returns the path for the application with an anchor for the comment" do
+      expect(helper.comment_path(comment)).to eq application_path(comment.application, anchor: "comment#{comment.id}")
```
henare: This test is generating both the test and the expectation. The expectation shouldn't be generated, instead it should just be a string, so that you're actually testing the generation properly.

app/views/comments/_comment_item.html.haml
```diff
@@ -1,20 +1,29 @@
...
-        = link_to "admin", admin_application_comment_path(comment), class: 'comment-action'
-      - unless current_page?(new_comment_report_path(comment.id))
-        = link_to "report comment", new_comment_report_path(comment), title: "Report this comment by #{comment.name} for removal", class: "comment-action"
+    %footer
+      %p.comment-meta
+        - # TODO: Change this to show the date in user local time.
+        - #       Could be nice and human to do 'today' and 'yesterday'
+        = link_to "Posted #{ comment.updated_at.strftime("%e %b %Y") }", comment_path(comment), title: "View comment"
+        and
+        %span delivered to #{comment.application.authority.full_name}
```
henare: Instead of "delivered", how about "sent". It's a simpler word and it's more commonly used when talking about emails.

spec/helpers/comments_helper_spec.rb
equivalentideas: The reason I used the path helper here was to isolate this test from any changing to the urls made. Rails generates that part, so it feels like testing the framework to hard code in the string. The test output it is still the full string so you can see exactly what’s happening:

```
expected: "/applications/1#commet1"
got: "/applications/1#comment1"
```

I think this is still testing quite tightly what we want to test. What do you think @henare ? This is an interesting issue.

app/views/comments/_comment_item.html.haml
equivalentideas: Thanks @henare . It was originally 'sent' but I changed it after seeing a user get a bit confused by it ( check out the discussion of this on the commit https://github.com/openaustralia/planningalerts/commit/0a80e5e6e97be7439f3e5daed43a3e982d5cd586 )

I agree 'sent' is more common, but 'delivered' makes our meaning clearer here I think. After making this change the people I tested this with understood the concept that this was sent *to* the councillor immediately.

henare:
comment

This isn't a deal-breaker but it's something you should consider @equivalentideas.

As you can see the comment looks a bit out of place in the above screenshots. The rounded corners are similar to the "Create Alert" widget in the sidebar but the rest of it is different enough to stand out. There are no other similar interface elements in PlanningAlerts that I know of.

If you add too many of these new elements a design can start to look really fractured. So just like code, every time you add a hack (or in this case a different design element) you need to accept you're taking on technical (or in this case design) debt.

app/views/comments/_comment.html.haml
equivalentideas: It doesn’t! Nice catch :baseball:.

This has also helped me see another way to improve this partial.

henare:
There are two dates on each comment, which I assume is a bug:

selection_002

henare:
Can we make the address part of the comment element? It looks a bit disjointed having it floating above the comment and I didn't initially associate the two.

selection_001

You could separate it in the same way as "they are yet to respond" and it would look quite similar to the current state which would be nice.

spec/helpers/comments_helper_spec.rb
henare: While you could certainly argue that this test is just testing the framework, you're not testing anything if you generate both sides of a test. It's like saying `expect(1 + 1).to eq (1 + 1)`, it needs to be `expect(1 + 1).to eq 2`.

It's up to you if you want to remove the test completely or fix this test.

henare:
Comment links should be underlined so you know you can click them (they're only shown on hover currently):

selection_002

app/views/comments/_comment_item.html.haml
henare: Okay but it still doesn't feel right to me.

spec/helpers/comments_helper_spec.rb
equivalentideas: I see what you mean. Something could happen that caused `comment.id` to return the wrong thing and this would still pass.

I'll update the test cause I think it’s better to keep the coverage so any changes to the helper are very intentional.

henare: Good call. I think it's much better to keep the test too. Now if you ever change the implementation of that method the test should still pass, yay :tada:

equivalentideas:
> As you can see the comment looks a bit out of place in the above screenshots. The rounded corners are similar to the "Create Alert" widget in the sidebar but the rest of it is different enough to stand out. There are no other similar interface elements in PlanningAlerts that I know of.

> If you add too many of these new elements a design can start to look really fractured. So just like code, every time you add a hack (or in this case a different design element) you need to accept you're taking on technical (or in this case design) debt.

This is an important point @henare thanks for raising this. I agree it does stand out a little knowing the site well. I'm happy in this case to recognise this as debt. I think we should be iterating more of the design forward to be friendlier and more welcoming/open.

equivalentideas:
@henare and I worked through all the comments here and we’re ready to merge and :shipit:

PR #840, opened 2015-11-04
Hotfix panel spacing

equivalentideas:
The spacing of the panel heading is a bit off, there's more space under the words than above. This fixes that to make them even.

This also includes a tiny refactor of the css on these selectors.

## Before

screen shot 2015-11-04 at 2 30 50 pm

## After

screen shot 2015-11-04 at 2 27 30 pm

### DIFF
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -40,11 +40,8 @@ ol#comments {
 }
 
 .panel-body {
-  padding: .75em 1em 1em;
-}
-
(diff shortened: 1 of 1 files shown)


equivalentideas:
@henare and I worked through this to simplify the text and also make the notice about the 'official commenting period' much clearer and focused on what matters to the user, whether their comment will be considered and impact the decision or not:

screen shot 2015-11-04 at 3 34 15 pm

equivalentideas:
Merging this as we’re happy and its tiny

PR #886, opened 2016-01-12
Improve the asthetics and usability of the charts on the authority pages

### LINKED ISSUE #884: Improve appearance of charts
The new charts are a big improvement and make it a lot easier to analyse the data contained in them but at the expense of their aesthetics. Aesthetics are important - we have to want to look at them and the more pleasing they are the more likely we are to absorb the information they contain.

I also find the combination of the horizontal scale opacity lines and the chunky y-axis legend marks make the chart _appear_ noisy and confusing.

The old charts looked pretty nice:

old charts

But the new ones look a bit like an 8-bit computer game :space_invader:

new chart

@equivalentideas pointed to the [nice](https://www.gov.uk/performance/carers-allowance/volumetrics) [charts](https://www.gov.uk/performance/carers-allowance) GDS have made as great references. And, "It would be nice to reuse their stuff" :+1: 

selection_001

equivalentideas:
This iterates on the previous update of the charts on the authority pages, to keep and improve the current usefulness of the chart and make them also look nicer.

This removes the noisy horizontal lines and changes the stepped chart for a slight curved interpolated line. This visually simplifies the chart, I think making it simpler to read. To compensate for removing the lines and interpolating the data, this adds tick marks back to the axis and also adds a 'callout' of the values and period under the cursor so you can still (and I think even more) get the precise detail of what you're looking at.

The 'callout' design pattern comes from the [GDS performance](https://www.gov.uk/performance/carers-allowance) examples, but this is a basic version. I wish we could just use their scripts completely, but as far as I can tell that would be pretty tricky.

@henare could you review this please? Is there anything else you can think to improve it at this point?

Fixes #884

## Before
screen shot 2016-01-12 at 4 10 38 pm

## After
screen shot 2016-01-12 at 4 08 02 pm

### DIFF
--- a/app/assets/javascripts/bar_graph.js
+++ b/app/assets/javascripts/bar_graph.js
@@ -21,8 +21,15 @@ function barGraph(selector, url, title) {
     var margin = { top: 20, right: 50, bottom: 20, left: 50 };
 
     var maxYValue = d3.max(data, function(datum) { return datum.values; });
-    var x = d3.time.scale().domain(d3.extent(data, function(datum) { return datum.key})).range([0, width]);
-    var y = d3.scale.linear()
+
+    var x = d3.time.scale()
+      .domain(
+        d3.extent(data, function(datum) { return datum.key; })
+      )
+      .range([0, width]);
+
+    var y = d3.scale
+      .linear()
       .domain([0, maxYValue])
       .rangeRound([height, 0]);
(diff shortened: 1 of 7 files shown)


henare:
Looks amazing :sparkles: Top stuff, I'll deploy this now.

PR #906, opened 2016-02-14
When checking state exceptions account for surrounding punctuation.

### LINKED ISSUE #625: Always capitalise Western Australia abbreviation "WA"
WA appears as “Wa”. My understanding is that both letters should be captialised.

See http://www.planningalerts.org.au/?q=129-131+Edward+Street%2C+Perth%2C+WA+6000 for example.

JWells:
Issue #625 looks to be caused by the state in the address being immediately bordered by punctuation. This changes the check to use a regexp to allow the possibility of non word characters before and after states.

Fixes #625

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -126,7 +126,7 @@ class Application < ActiveRecord::Base
 
     if address
       address.split(' ').map do |word|
-        if word != word.upcase || exceptions.include?(word) || word =~ /\d/
+        if word != word.upcase || exceptions.any? { |exception| word =~ /^\W*#{exception}\W*$/ } || word =~ /\d/
           word
         else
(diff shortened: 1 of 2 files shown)


equivalentideas:
Nice one @JWells this does the trick :hammer: I'll merge and :ship: 

It would be nice to refactor this to make [that long conditional](https://github.com/openaustralia/planningalerts/pull/906/files#diff-2ee7585cff5e8377195835a9f90cdaa3R129) a clearer so people don't have to read the regex ;-) :

        if word != word.upcase || exceptions.any? { |exception| word =~ /^\W*#{exception}\W*$/ } || word =~ /\d/

Maybe extracting the state stuff into a helpfully named method would help?

PR #908, opened 2016-02-17
Set supplied radius or its default to float.

JWells:
This one won't break the or condition so the default will still work.

### DIFF
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -41,7 +41,7 @@ class ApplicationsController < ApplicationController
 
   def address
     @q = params[:q]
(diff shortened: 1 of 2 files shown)


equivalentideas:
Thanks @JWells . I think this needs some controller tests for the radius functionality. Would you be up for including these in this PR?

JWells:
No problems, I've added some tests just on the radius variable to keep this PR small.

equivalentideas:
Thanks @JWells . This combined with the new integration tests makes me feel a little safer :)

equivalentideas:
That’s deployed @JWells thanks again :stars:

PR #919, opened 2016-03-01
Allow admins to control whether people can write to councillors for each authority

equivalentideas:
To control the roll out of the Write To Councillors feature, admins should be able to toggle the feature on or off for an individual authority.

These commits add that control by adding “Write to councillors enabled” checkbox to the authority admin page. Under the hood it's a new boolean attribute to authorities `write_to_councillors_enabled`.

This now means that both the global feature flag (environment variable) and the flag for the authority must be enabled to make Write To Councillors available for an authority. This adjusts the integration tests for commenting and writing to councillors accordingly.

screen shot 2016-03-01 at 5 40 55 pm
screen shot 2016-03-01 at 5 40 44 pm

**Very sloppily this PR includes a clean up of whitespace on a view spec where I first started this working through this :( **

### DIFF
--- a/spec/features/feedback_spec.rb
+++ b/spec/features/feedback_spec.rb
@@ -80,7 +80,7 @@ feature "Give feedback to Council" do
       end
 
       background do
-        application.authority.update(write_to_councillors_enabled: true)
+        application.authority.update!(write_to_councillors_enabled: true)
 
         create(:councillor, authority: application.authority)
       end
--- a/spec/features/feedback_spec.rb
+++ b/spec/features/feedback_spec.rb
@@ -80,7 +80,7 @@ feature "Give feedback to Council" do
       end
 
       background do
-        application.authority.update_attribute(:write_to_councillors_enabled, true)
+        application.authority.update(write_to_councillors_enabled: true)
 
         create(:councillor, authority: application.authority)
       end
--- a/spec/features/feedback_spec.rb
+++ b/spec/features/feedback_spec.rb
@@ -80,6 +80,8 @@ feature "Give feedback to Council" do
       end
 
       background do
+        application.authority.update_attribute(:write_to_councillors_enabled, true)
+
         create(:councillor, authority: application.authority)
       end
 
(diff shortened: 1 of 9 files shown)


equivalentideas:

equivalentideas:
Just remember to note here, if we merge this we must remember to switch the feature on for the 3 authorities it's meant to be enabled for.

spec/features/feedback_spec.rb
```diff
@@ -80,6 +80,8 @@
       end
 
       background do
+        application.authority.update_attribute(:write_to_councillors_enabled, true)
```
henare: @equivalentideas I'm pretty sure I told you about this method so it's my fault :smile: but a more concise version of this is:

```ruby
application.authority.update(write_to_councillors_enabled: true)
```

Unless you're capturing the return value and doing something with it you should also use the bang method so it raises and error if there's a problem (fail fast):

```ruby
application.authority.update!(write_to_councillors_enabled: true)
```

equivalentideas: :+1: That sounds good.

I can change it to use `update!`

henare: I've got it.

Speaking of this, I don't really understand what [`#update_attribute`](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attribute) is for. The implementation is quite different from `#update` and I don't really see why.

equivalentideas: http://api.rubyonrails.org/v3.2.22.1/classes/ActiveRecord/Persistence.html#method-i-update_attribute

Validations are skipped with `#update_attribute`. I think I might have been defaulting to using this in tests so that they're isolated from validations and whatever else we have/add.

All we care about in the test when we toggle this attribute is that the value is true/false, not whether the instance still validates after the change. Using `update!` does make the test a little less focused then. We just chatted about this, and thought that we wouldn't want the use of `update_attribute` to slip into the app code, so better/fine to use `update!` anyway.

henare: Great sleuthing! :mag: 

My change was definitely unnecessary then :frowning: As you were!

PR #931, opened 2016-03-30
Show and sort comments with the time they were confirmed_at

### LINKED ISSUE #930: The time we say a comment was posted is wrong
Every comment says when it was posted:

> Alexander commented about 16 hours ago

But this is not the time the person commented, it's the time the comment [was last updated](https://github.com/openaustralia/planningalerts/blob/master/app/views/comments/_comment_meta_sentence.html.haml#L10-L11) :(

screen shot 2016-03-29 at 3 09 23 pm

Most of the time this is the time the comment was confirmed by the user. That seems fine. But if for some reason the comment gets updated, by an admin for example, the update time changes and the 'commented' time becomes 'now'.

Even worse, if that comment has a reply, the site says the person wrote to the councillor 'about 2 minutes ago' and the councillor replied 'about 4 days ago'.

equivalentideas:
Fixes #930

Currently the time we say the person made a comment is the `updated_at`
time. For most comments this ends up being the time the `confirmed`
attribute was updated, i.e. when they confirmed the comment. But when an
Admin updates the comment for some reason, the `updated_at` time changes
and we then wrongly say that is when the person commented.
This has lead to situations where the website says someone commented
days after they actually commented.

These commits add new attribute 'confirmed_at' which is set when the person
confirms their comment.

This also adds a migration to add the `confirmed_at` value for existing confirmed comments. It uses the `updated_at` value, which is what we currently use to show and sort comments.

Lots of detail in the commit messages.

@henare could you please review this?

### DIFF
--- a/db/migrate/20160330040409_fill_confirmed_at_for_comments.rb
+++ b/db/migrate/20160330040409_fill_confirmed_at_for_comments.rb
@@ -1,0 +1,5 @@
+class FillConfirmedAtForComments < ActiveRecord::Migration
+  def change
+    Comment.fill_confirmed_at_for_existing_confirmed_comments
+  end
+end
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -101,7 +101,7 @@ class ApplicationsController < ApplicationController
     end
 
     @application = Application.find(params[:id])
-    @comments = @application.comments.visible.order(:updated_at)
+    @comments = @application.comments.visible.order(:confirmed_at)
     @nearby_count = @application.find_all_nearest_or_recent.size
     @add_comment = AddComment.new(
       application: @application,
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -11,7 +11,7 @@ class CommentsController < ApplicationController
       comments_to_display = Comment.all
     end
 
-    @comments = comments_to_display.visible.order("updated_at DESC").paginate page: params[:page]
+    @comments = comments_to_display.visible.order("confirmed_at DESC").paginate page: params[:page]
     @rss = comments_url(params.merge(format: "rss", page: nil))
 
     respond_to do |format|
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -113,7 +113,11 @@ class Alert < ActiveRecord::Base
 
   # Applications in the area of interest which have new comments made since we were last alerted
   def applications_with_new_comments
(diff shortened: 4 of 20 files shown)


henare:
I was just reviewing something unrelated and was reminded of the performance pages. Has that been taken into consideration in this PR? Are there other places this could affect?

equivalentideas:
> I was just reviewing something unrelated and was reminded of the performance pages. Has that been taken into consideration in this PR? Are there other places this could affect?

Good question @henare .

The Performance chart for comments collects comments by their created time so that wont be affected.
I think it should use the `confirmed_at` time instead so it doesn't include comments that aren't confirmed. If we wanted to track unconfirmed comments that would be as part of a funnel, not a simple time series like this.

The alerts do collect `recent` comments by `updated_at`. I think we should use `confirmed_at`, which also means we can simplify the code.

I'll update that as part of this PR, assigning back to me.

henare:
Turns out it wasn't the performance page I saw fly past in code, it was the authority graphs: https://github.com/openaustralia/planningalerts/blob/master/app/models/authority.rb#L183-L204

Please grep through the code and check instances where `updated_at` is used to make sure we're getting everything.

equivalentideas:
Thanks @henare !

equivalentideas:
Updated this to change from `updated_at` to `confirmed_at` for:

* [x] What gets included in alerts
* [x] The performance page
* [x] The by week charts and feeds for authority comments

This covers everything, except the vanity stats in the footer 'this week you made 77 comments'. This uses 'created_at' like the number for new alerts. Since these don't have the `updated_at` bug this targets, I think we should leave it for now and open a new issue if we think it needs to be changed at all.

I'm assigning back to @henare for review since this touches at lot of functionality.

henare:
I've tested this locally and deployed it to the test stage. Everything seems to be working so I'll merge and deploy.

henare:
Deployed :rocket:

For reference, it took 3:22 to migrate the production database which has 6,400 comments.

db/migrate/20160330040409_fill_confirmed_at_for_comments.rb
```diff
@@ -0,0 +1,5 @@
+class FillConfirmedAtForComments < ActiveRecord::Migration
+  def change
+    Comment.fill_confirmed_at_for_existing_confirmed_comments
```
henare: @equivalentideas I just noticed this was an irreversible migration. It's a good idea to always try to make migrations reversible.

PR #935, opened 2016-04-02
Fixes #232 Add no results text

### LINKED ISSUE #232: When a search returns no results, nothing informs the user of this (PA-217)
<p>It should say, "Sorry, no results found" at least.</p>

---
**Issue details imported from Jira:**

Detail | Value
------ | -----
Assignee | Unassigned
Reporter | Henare Degan
Created | 2010-10-20 13:48:00
Updated | 2011-02-01 17:05:40
---
### Imported Jira Comments
**Matthew Landauer** - 2010-12-14 15:06:24
><p>Now, it says there are no results and gives you some information as to why that is and what you can do about it</p>

**Henare Degan** - 2010-12-14 15:23:46
><p>Sorry, I meant this page: <a href="http://www.planningalerts.org.au/applications/search?q=DA201000452" class="external-link">http://www.planningalerts.org.au/applications/search?q=DA201000452</a></p>

CloCkWeRX:
Dunno if these pages are still linked to at all, but #232 was an easy fix.

### DIFF
--- a/app/views/applications/search.html.haml
+++ b/app/views/applications/search.html.haml
@@ -4,7 +4,7 @@
   = text_field_tag(:q, @q, :class => "textbox")
   = submit_tag "Search >>", :class => "button", :name => nil
 
-- if @applications
+- if @applications.any?
   = paginated_section @applications, :previous_label => "« Newer", :next_label => "Older »" do
     = render "applications", :applications => @applications
 - elsif @q
--- a/app/views/applications/search.html.haml
+++ b/app/views/applications/search.html.haml
@@ -7,3 +7,6 @@
 - if @applications
   = paginated_section @applications, :previous_label => "« Newer", :next_label => "Older »" do
     = render "applications", :applications => @applications
+- elsif @q
+  %br
+  %p.attention Sorry, no results found!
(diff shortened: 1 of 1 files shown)


henare:
@CloCkWeRX did you test this? It doesn't work for me.

selection_001

Please make sure you actually test your change before asking for a pull request to be reviewed.

CloCkWeRX:
I did indeed test locally. The reason it's checking ```@q``` is so it doesn't display 'no results found' on the first page, but I only realised that was a problem towards the end of the testing: potentially I saw it *not* displaying warnings and took that as a false positive in the wrong scenario.

Will give it another look to see what I missed

henare:
>I did indeed test locally.

That's strange, so the message shows up for you?

>Will give it another look to see what I missed

Inspecting with the REPL in better_errors shows `@applications` is an empty array when a search for no results is done on my machine. This means it evaluates as `true` and so that message is never shown (as far as I can tell).

CloCkWeRX:
Ah, I was manipulating that to be falsey

henare:
LGTM, thanks @CloCkWeRX :sunflower:

Issue #922, opened 2016-03-21
Say on the comment form that abusive comments are not tolerated


henare:
We should be clear and up front about this so everyone understands what is and isn't acceptable on PlanningAlerts. We should link to the complete abuse policy done as part of #923.

Extracted from #912 after some thought.

equivalentideas:
Why have you assigned this to me @CloCkWeRX ?

CloCkWeRX:
#937 is an implementation of this, you wanted to go another direction.

equivalentideas:
> #937 is an implementation of this, you wanted to go another direction.

👌 I think my basic thoughts on this at the moment are that we should do #923 first and also do some sketching of a few different ways we could approach this issue before implementing it.

mlandauer:
I agree that some reference to the abuse policy as a reminder that we're trying to create a safe space would be a good thing. It's not clear to me yet how to do that so it is effective

Issue #923, opened 2016-03-21
Clearly state our abuse policy somewhere we can link to it


henare:
Currently the only thing we say about abuse is on our "Report comment" page:

```
Reasons [for removing a comment] can include that the comment is spam, abusive, unlawful or harassing — in other words, where people are going out of their way to cause harm.
```

We need to move that somewhere we can link to it for #922. We also need to expand and clarity that to make it clear that racist abuse is not tolerated.

While we're doing that it might be a good idea to revise it completely. Are there other abuse policies out there somewhere that we could use? Of course we shouldn't let revising it completely get in the way of doing the simple things above.

Extracted from #912 after some thought.

CloCkWeRX:
http://www.google.com/intl/en-us/+/policy/content.html I think might be a good starting point, touching on a lot of areas

henare:
I like this [from The Conversation](https://theconversation.com/sign_in) too:

>We welcome debate and dissent, but personal attacks (on authors, other users or any individual), abuse and defamatory language will not be tolerated. Nor will we tolerate attempts to deliberately disrupt discussions. We aim to maintain The Conversation as an inviting space to focus on intelligent discussions.
>
>Please be courteous.
>
>Read our community standards

mlandauer:
Agree that this would be super valuable!

Issue #917, opened 2016-03-01
show details of hidden comments


kat:
show the display name of commenter and [this comment has been hidden because it is abusive]  instead of the original message. @henare suggested considering for transparency being able to also click on that message to see the original comment.

kat:
[this comment has been reported to moderators] and [this comment has been reported for abuse and has been hidden]

equivalentideas:
We could also link to our lovely new (to be) abuse policy on the removed comment https://github.com/openaustralia/planningalerts/issues/923

mlandauer:
We've implemented a very rough and ready early version of this which just shows that a comment was hidden by administrators.

I think we should leave this issue open to remind us to implement the rest of this

PR #937, opened 2016-04-02
Add visual reminder to play nice

CloCkWeRX:
#922

### DIFF
--- a/app/assets/stylesheets/partials/_page.scss
+++ b/app/assets/stylesheets/partials/_page.scss
@@ -665,7 +665,7 @@ ul.ui-autocomplete {
   display: inline-block;
 }
 
-#application-info .button {
+#application-info .comment-policy-reminder {
   margin-bottom: 2em;
 }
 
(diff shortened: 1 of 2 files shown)


CloCkWeRX:
image

henare:
Assigning to @equivalentideas to review.

equivalentideas:
Thanks for this @CloCkWeRX it's helped me clarify my thoughts a bit on how we frame the comments section.

I think this is too in-your-face and really just reminds/informs people that others are using PlanningAlerts to harass and abuse people, which most people never come across. Almost all the abuse is isolated to a few applications, and I doubt those people will head this warning.

It does tell people that this is meant to be a safe place where you wont be harassed, which is good—but I think there must be a more effective way to achieve this.


I think some of the other ideas like #917 are more likely to impact people and have less of a negative effect on all majority of non-abusive commenters.

I think it's more important that we do https://github.com/openaustralia/planningalerts/issues/923 first to develop our policy and the way we tell people about it.

I'm going to close this, please anyone reopen if you disagree.

Issue #317, opened 2013-12-29
Shorten very long comments on display (PA-303)


oajirabot:
<p>Then allow the comment to be shown in full.</p>

<p>We want this feature because it's very likely that the occasional person will write a very long comment</p>

---

**Issue details imported from Jira:**

| Detail | Value |
| --- | --- |
| Assignee | Unassigned |
| Reporter | Matthew Landauer |
| Created | 2011-04-28 21:39:58 |
| Updated | 2011-04-28 21:39:58 |

equivalentideas:
I'm a bit concerned this issue is based on an admin experience of looking through lots of comments. What is the user's goal when they have the problem of scrolling past long comments? Why is this problem?

@CloCkWeRX Were you looking at some comments and had a bad experience when you picked up #939 ?

CloCkWeRX:
The long comments UI problem isn't *just* an admin problem; but also for anyone browsing a specific application.

I tried to strike a balance between "reasonable long form complaint about a particular application made" and "really angry person maliciously puts in 16-32kb of text obscuring the comments of others by making a super long scrolling experience" with the PR (and just addressing the rendering part of it seemed quite easy).

equivalentideas:
I think this convo is happening over on #939 now.

mlandauer:
Closing this in favour of #939.

PR #939, opened 2016-04-02
Make very long comments less of a problem

CloCkWeRX:
Improves but doesn't entirely address #317 - doesn't stop someone posting in a 64kb comment, but you can scroll through the page at least without being forced to read it.

image

### DIFF
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -144,6 +144,9 @@ ol#comments {
   font-style: normal;
   border: 0;
   padding: 0;
+
+  max-height: 30em;
+  overflow-y: auto;
 }
 
 .comment-location {


henare:
I'm inclined to merge this but let's get @equivalentideas thoughts first.

equivalentideas:
Can you link to any other examples of where a comment is presented like this? Since comments are such a familiar pattern I'm keen to draw on familiar solutions to stuff like this if we can.

I've seen scroll bar interfaces pose really big challenges to users. They think they have to click and drag the bar, which is really fiddly for them. I can see them having a problem using this interface by both not seeing that there is more comment to scroll through and getting frustrated using the scroll bar.

I think a 'read more' or 'show full comment' button could be more usable and a more obvious signifier of what's happening here. I also think that's a more familiar pattern, appearing in major news sights to show full articles from an excerpt for example. What do you think @CloCkWeRX ?

Also any thoughts on why 30em for the max length?

equivalentideas:
@CloCkWeRX This prompted me to think more about whether the problem for users this is trying to solve is a real problem, see https://github.com/openaustralia/planningalerts/issues/317#issuecomment-205094698 .

I think we should hold off until we have a clear understanding of how this helps the user achieve their goal. This solution, or the one I suggested, add new interface elements to comments, making them less simple. We need to know what people gain from us adding this complexity. What do ya think @CloCkWeRX ?

CloCkWeRX:
Covered some of that in https://github.com/openaustralia/planningalerts/issues/317#issuecomment-205095861 

tl;dr: only kicks in for excessively long comments (defined as >= 30 lines) which allows one particular comment to drown out the others / force a lot of scrolling.

Though it's going to be rare, its likely comment length follows a power law distribution and there will be a small number of quite long comments - either well intentioned or malicious.

equivalentideas:
> drown out the others / force a lot of scrolling.

What's the action users are taking when they are forced to scroll passed long comments they don't want to read?

equivalentideas:
From @CloCkWeRX chat:

> Put on a slightly different hat from 'normal, rational user' for a sec and put yourself in irrational hatemailer/abusive user mode. You find out the local council is a going to let a Kitten shop open and you are a 100% Grade-A All Australian Dog person; and before you know it, long abusive rant filled with anti Cat sentiments come out.
>
> Others browse the application and see the anti-cat rant; but ​*can't see the report comment control*​ (it's below the page fold); or see other comments below this one as quite frankly, long-cat-rant-is-long.

So the activity people are trying to do here is 'report abusive comments'. The design failure is that on abusive comments, which are commonly very long, the report link is at the end of the comment, which makes it hard to spot. The user may not even know the option exists when they're reading the comment. Abusive comments go unreported, users suffer.

I strikes me that a simpler solution to this could be to move the report link to the top of the comment, possibly using a flag icon or something? Facebook have their report link at the top:

screen shot 2016-04-04 at 12 21 17 pm

>
> Lesser versions of this in production: 
> https://www.planningalerts.org.au/applications/600370 is a pretty normal application.
>
> Try finding Dragan Aleksic's comments without using ctrl+f / by scrolling alone; and there's three comments that eat up a lot of screen realestate.
>
> (You'll get there, but you spend a lot of time on the minutae of other people's innermost thoughts when it's not your goal)

What task are people trying to accomplish when looking for a specific commenter?

By the by I found the comment very quickly because the names are highlighted and I could scan for it. I am nothing like a typical PlanningAlerts user in my use of websites though, and especially when it comes to this website ;-)

> scenario two: be a mean tester or spammer and paste an entire wikipedia article into a comment field (up to 64kb of content) with a link for http://buy-puppies-not-cats.com/ tucked into the last two lines.

I see what you mean, but we don't seem to have much/any spam at the moment, so I don't think this is a real use case.

equivalentideas:
@CloCkWeRX As discussed above, I don't think this is quite the right solution—but thanks so much for contributing it because I think we articulated much better what the actually problem to solve might be, so that's awesome.

To clean things up I'm gonna close this, reopen if you disagree as always :)

Issue #442, opened 2013-12-29
Add link to home page of planning authority website on authority detail page (PA-429)


oajirabot:
<p>This came out of discussions with John from homesales.com.au</p>

* [x] Add a field in the database (done in https://github.com/openaustralia/planningalerts/pull/1242#issuecomment-327049719 by @hisayohorie 🍁 )
* [x] Fill in the values (we've already collected a lot of them, see https://github.com/openaustralia/planningalerts/issues/1244 )
* [ ] Show the values on the authority detail page
* [ ] Deliver this in the api

---

**Issue details imported from Jira:**

| Detail | Value |
| --- | --- |
| Assignee | Unassigned |
| Reporter | Matthew Landauer |
| Created | 2012-09-18 17:36:46 |
| Updated | 2012-09-18 17:36:46 |

mlandauer:
I think most of the values for the website have been filled in. So I'm going to tick that box as done.

mlandauer:
I definitely think we should show the link on the authority detail page. I'm not 100% sure we should include it in the api

PR #940, opened 2016-04-02
Add fields for primary_url on authority

CloCkWeRX:
First bit of #442

Probably can infer the majority of the homepages from authority.email.split("@").last; but didn't want to stick that in a migration.

### DIFF
--- a/db/migrate/20160402093711_add_homepage_url.rb
+++ b/db/migrate/20160402093711_add_homepage_url.rb
@@ -1,0 +1,5 @@
+class AddHomepageUrl < ActiveRecord::Migration
+  def change
+    add_column :authorities, :primary_url, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160301033011) do
+ActiveRecord::Schema.define(version: 20160402093711) do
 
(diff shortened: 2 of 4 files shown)


db/schema.rb
```diff
@@ -95,17 +95,21 @@
...
     t.boolean "disabled"
-    t.string  "state",                limit: 20
+    t.string  "state",                        limit: 20
     t.string  "email"
     t.integer "population_2011"
     t.text    "last_scraper_run_log"
     t.string  "morph_name"
     t.boolean "write_to_councillors_enabled",             default: false, null: false
+    t.string  "lga_name15"
+    t.string  "superceded_by"
```
henare: Looks like you've committed extra stuff that shouldn't be in this PR.

db/migrate/20160402093711_add_homepage_url.rb
```diff
@@ -0,0 +1,5 @@
+class AddHomepageUrl < ActiveRecord::Migration
+  def change
+    add_column :authorities, :primary_url, :string
```
henare: How about `home_page` instead? I found `primary_url` a bit confusing.

db/schema.rb
CloCkWeRX: Unfortunately that'll keep appearing due to me not dropping / recreating the db from scratch in a number of pull requests - only realised after the first one or two. Silly db/schema, why u not feature branch well.

henare: In that case you should `git add --patch` and not commit that stuff.

PR #941, opened 2016-04-02
Fixes #628 by adding link to footer

### LINKED ISSUE #628: Link to project GitHub on the site
We don't currently.

CloCkWeRX:
Before:
image

After:
image

### DIFF
--- a/app/views/application/_footer.haml
+++ b/app/views/application/_footer.haml
@@ -11,6 +11,9 @@
       %li
         Read our blog
         = link_to "openaustraliafoundation.org.au/blog", "http://www.openaustraliafoundation.org.au/blog"
+      %li
+        Fork this project
+        = link_to "github.com/openaustralia/planningalerts", "https://github.com/openaustralia/planningalerts"
     .oaf
       %h4
         An
--- a/app/views/application/_footer.haml
+++ b/app/views/application/_footer.haml
@@ -12,7 +12,7 @@
         Read our blog
         = link_to "openaustraliafoundation.org.au/blog", "http://www.openaustraliafoundation.org.au/blog"
       %li
-        Fork this project
+        Source code
         = link_to "github.com/openaustralia/planningalerts", "https://github.com/openaustralia/planningalerts"
     .oaf
       %h4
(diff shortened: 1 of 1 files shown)


app/views/application/_footer.haml
```diff
@@ -11,6 +11,9 @@
       %li
         Read our blog
         = link_to "openaustraliafoundation.org.au/blog", "http://www.openaustraliafoundation.org.au/blog"
+      %li
+        Fork this project
```
henare: "Fork" is a bit jargony for this project - the general audience aren't programmers so wouldn't understand what it meant. Can we improve this wording?

CloCkWeRX: While that's true, its the github link - wouldn't the intended users be developers?

How does just "Github" or "Us on Github" work for you?

equivalentideas: Could also be 'source code' like on morph.io . or "Source code on GitHub" ?

For people who aren't devs who want to see what's 'behind the scenes' “source code” is more likely a familiar term or something they can work out, I think.

equivalentideas:
Nice work with this @CloCkWeRX Could you please add a before and after screenshot so we can see the impact of adding another link there to the layout?

CloCkWeRX:
Updated

app/views/application/_footer.haml
henare: morph.io also has "GitHub".

equivalentideas:
Great thanks @CloCkWeRX .

It looks like the right is becoming more about OAF and the left about PlanningAlerts. Think we should move the blog link to the right to be with the conceptually similar stuff and rebalance the layout?

CloCkWeRX:
Hard to get the vertical heights right without changing a bunch more of the headings; or putting in some kind of grid layout.

image

equivalentideas:
@CloCkWeRX I think that's more balanced than it currently is ;-)

I'm happy with that, just make sure it's all good on small screens too.

CloCkWeRX:
image

equivalentideas:
Nice work @CloCkWeRX I've merged this locally.

I removed the `<br />` added and recreated the layout in css in 3cb78ff .

equivalentideas:
Deployed 🚢 Thanks @CloCkWeRX !

PR #942, opened 2016-04-02
Fixes #691 Add validation

### LINKED ISSUE #691: Mysql2::Error: Duplicate entry 'noosa' for key 'short_name_unique'
We should add a validation to the Authority model.

## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37075/faults/14455029)

CloCkWeRX:


henare:
Strange. I'm still getting the error from #691 when I test locally.

@CloCkWeRX any idea what I'm doing wrong?

CloCkWeRX:
I'll give it another look - was running into it with the 'add all lgas' work I was doing and tested mostly via that.

CloCkWeRX:
Added a spec and it works for me:
image

henare:
Wacky. I still get an exception when trying to create a duplicate:

selection_001

This should work. Assigning to @equivalentideas for a different set of :eyes:

spec/models/authority_spec.rb
```diff
@@ -1,6 +1,16 @@
 require 'spec_helper'
 
 describe Authority do
+  describe "validations" do
+    it 'should ensure a unique short_name' do
+      @a1 = create(:authority, short_name: "Existing")
```
equivalentideas: No need to assign this to a variable, it's not used.

spec/models/authority_spec.rb
```diff
@@ -1,6 +1,16 @@
...
+      @a2 = build(:authority, short_name: "Existing")
```
equivalentideas: This can be a local variable. Elsewhere in the file instance variables are assigned in a `before` block for use in a few examples.

It would also be good to assign it a useful name like 'new_authority' (maybe?) to help tell the story of the test.

spec/models/authority_spec.rb
```diff
@@ -1,6 +1,16 @@
...
+
+      @a2.valid?.should == false
+      @a2.errors.messages[:short_name].should == ["has already been taken"]
```
equivalentideas: Let's change these to the newer `expect` syntax here. I realise that a few of the other tests in this file use the `should` syntax, but we're using the newer syntax for new tests and trying to update old ones when we touch them. http://betterspecs.org/#expect

equivalentideas:
This works for me as you'd expect. I get the nice error message on the admin page, and also tested in the rails console. Not sure what's going on with your setup @henare .

equivalentideas:
I made a few comments on the test code that I think can be improved. Otherwise very nice @CloCkWeRX 

nice

Assigning back to @henare to see if he can workout the bug he’s seeing.

spec/models/authority_spec.rb
CloCkWeRX: If it starts getting re-used in more than one spot; I'll shift it to a context/before :each block... but it's only in one place right now.

spec/models/authority_spec.rb
CloCkWeRX: I'll convert the rest of the app with transpec at some point; as part of #946

spec/models/authority_spec.rb
```diff
@@ -3,11 +3,14 @@
...
     it 'should ensure a unique short_name' do
-      @a1 = create(:authority, short_name: "Existing")
-      @a2 = build(:authority, short_name: "Existing")
+      existing_authority = create(:authority, short_name: "Existing")
+      new_authority = build(:authority, short_name: "Existing")
 
-      expect(@a2.valid?).to eq false
-      expect(@a2.errors.messages[:short_name]).to eq(["has already been taken"])
+
+      expect(existing_authority.valid?).to eq true
```
equivalentideas: Isn't this test superfluous because if `create` succeeds on line 6 then this will always be true? In other words it's not possible for this to fail.

I think you could remove this and not assign the first Authority, and the test would be clear and focused.

spec/models/authority_spec.rb
equivalentideas: The only thing to remember with that is there can be trade off between DRYness and how clear an individual test is. That's why helpful variable names are important I guess ;-) If you have to look in a bunch of places to understand what a test is basically doing then I think it's worth considering inlining the variables, or improve the naming.

henare:
Tested this again and it still wasn't working but I've finally found why. The validation is case sensitive and my development machine had "Marrickville" and I was trying to create a new one with "marrickville".

It should be case insensitive and I've done that in 977d143ff8fca80da0fb7d3a1c2393441d6cb1ca.

Issue #889, opened 2016-01-12
Confusing error message on address search


equivalentideas:
When you [submit the address "123 Broadway St, Ultimo NSW"](https://www.planningalerts.org.au/?q=135+Broadway+St+Ultimo+Nsw) into the application search form, you get the error:

> Please enter a full street address like ‘36 Sowerby St, Goulburn, NSW’

The real error is that it can't find that address because it doesn't exist (should be 123 Broadway, Ultimo...). This is really confusing because you have put in a full street address. In our analytics we can see that when this happens, people just keep hitting submit hoping that it will work :( .

The error messages on this page get seen by lots of people, so it's important they're as helpful as possible. Our tracking says that last week, we picked up 534 clicks to submit address searches, and an error messages was displayed 435 times.

![screen shot 2016-01-13 at 7 17 32 am](https://cloud.githubusercontent.com/assets/1239550/12275724/c6f6ee56-b9c5-11e5-92fb-01eed27993a0.png)

henare:
Are you really saying 80% of address searches failed? That sounds like a really big problem and one that if we fix would have a massive benefit!

equivalentideas:
> Are you really saying 80% of address searches failed? That sounds like a really big problem and one that if we fix would have a massive benefit!

We can actually see the queries people are making in GA. It's mostly searches for suburbs. I'd say we could probably give these people a more helpful result than we currently do. A lot of people seem to get stuck in a cycle of submitting searches and getting errors. Only around 10% of people who see an error end up clicking 'create alert'.

henare:
I'm confused. Can you please answer yes or no to this:

>Are you really saying 80% of address searches failed?

If no, can you please elaborate on what you mean?

equivalentideas:
> Are you really saying 80% of address searches failed?

Yes I think that's right.

From our Google Analytics event tracking:

* the search form submit button was clicked 534 times
* an error message on this form was shown 435 times

henare:
Wow. How does that reconcile with the number of successful sign ups?

>you signed up for 295 new email alerts

equivalentideas:
Only 5% of sessions included a visit to "/" (the application search page), so most signups come via an application page. Does that answer your question?

henare:
Yes, thank you.

So fixing this seems like it could potentially lead to hundreds more sign ups a week - amazing find!

equivalentideas:
[...]

PR #948, opened 2016-04-04
Tolerate looser geocoding

CloCkWeRX:
As per #889 convos

Before; users get stuck here

Now:

This changes the hints for "Please enter a full street address like ..." to indicate a vaguer, street level address is OK 

```
Please enter a more detailed address like ‘Sowerby St, Goulburn, NSW’
```

... though I can't convince the geocoder to cough up a result of that accuracy).

Adds Polite nudge UI on search results for a vague address:
image

and doesn't reject suburb level addresses

### DIFF
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -46,19 +46,19 @@ class ApplicationsController < ApplicationController
     per_page = 30
     @page = params[:page]
     if @q
-      location = Location.geocode(@q)
-      if location.error
+      @location = Location.geocode(@q)
+      if @location.error
         @other_addresses = []
-        @error = location.error
+        @error = @location.error
       else
-        @q = location.full_address
+        @q = @location.full_address
         @alert = Alert.new(address: @q)
-        @other_addresses = location.all[1..-1].map{|l| l.full_address}
+        @other_addresses = @location.all[1..-1].map{|l| l.full_address}
         @applications = case @sort
                         when 'distance'
-                          Application.near([location.lat, location.lng], @radius / 1000, units: :km).reorder('distance').paginate(page: params[:page], per_page: per_page)
+                          Application.near([@location.lat, @location.lng], @radius / 1000, units: :km).reorder('distance').paginate(page: params[:page], per_page: per_page)
                         else # date_scraped
-                          Application.near([location.lat, location.lng], @radius / 1000, units: :km).paginate(page: params[:page], per_page: per_page)
+                          Application.near([@location.lat, @location.lng], @radius / 1000, units: :km).paginate(page: params[:page], per_page: per_page)
                         end
         @rss = applications_path(format: 'rss', address: @q, radius: @radius)
       end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -49,6 +49,10 @@ class Location < SimpleDelegator
     end
   end
 
+  def less_than_street_level_accuracy?
+    accuracy < 5
+  end
+
   # Distance given is in metres
   def endpoint(bearing, distance)
     Location.new(__getobj__.endpoint(bearing, distance / 1000.0, units: :kms))
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -43,8 +43,8 @@ class Location < SimpleDelegator
         "Sorry we don’t understand that address. Try one like ‘1 Sowerby St, Goulburn, NSW’"
       elsif !in_correct_country?
         "Unfortunately we only cover Australia. It looks like that address is in another country."
-      elsif accuracy < 5
-        "Please enter a full street address like ‘36 Sowerby St, Goulburn, NSW’"
+      elsif accuracy < 4
+        "Please enter a more detailed address like ‘Sowerby St, Goulburn, NSW’"
       end
     end
   end
--- a/app/views/applications/address_results.html.haml
+++ b/app/views/applications/address_results.html.haml
@@ -1,6 +1,9 @@
 - content_for :meta_description, @themer.default_meta_description
 
 %h3 Applications within #{meters_in_words(@radius)} of #{@q}
+- if @location.less_than_street_level_accuracy?
+  %p.attention
+    We didn't fully recognize your exact address, so you may not see the most relevant applications. Search with a street level or exact address to get better results. #{link_to "Search again", root_path}.
 
 - unless @other_addresses.empty?
   %h4 Or did you want?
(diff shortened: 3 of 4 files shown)


equivalentideas:
Nice one @CloCkWeRX could you add a screenshot of what the user would normally see in this situation?

CloCkWeRX:
Users only see the error result/can't proceed.

henare:
@equivalentideas can I assign this to you? It seems like it needs reviewing from a UX perspective most of all.

Before we merge it would be also good to think about how we measure if this has been successful or not by determining what analytics should change and making sure we review them at the appropriate time.

equivalentideas:
> @equivalentideas can I assign this to you? It seems like it needs reviewing from a UX perspective most of all.

> Before we merge it would be also good to think about how we measure if this has been successful or not by determining what analytics should change and making sure we review them at the appropriate time.

Yep this all sounds good. I think making this awesome will be a big help to lots of users so excited to work on it 👍

equivalentideas:
@CloCkWeRX Thanks again for this, it'll really help lots of people I think :)

From looking at our analytics there are a four main reasons that people get an error when searching:

1. They expect the form to work in a way that it doesn't and enter a suburb like "St Leonards VIC", "Raymond Terrace NSW" or "Burpengary QLD".
2. They expect the form to work in a way that it doesn't and enter a very specific address formatted in a way we can't parse, like "Jones Rd at IGA Shopping Centre, Toowoomba QLD" or "Gowrie St near Harold, Newtown NSW"
3. They have correctly or mistakenly entered a street address that the geocoder can't recognise like "7 Fake st, Ultimo NSW"
4. They have made a slip while typing and entered the a bad address "1 Dank st MelbVIC"

Situation 1 is _by far_ the most common.

In situations 2, 3 or 4, I think the most helpful thing is to show the user an error message on the form so that they can recognise and correct the error "Sorry we couldn't recognise that address, make sure it is formatted correctly". They're expecting to get results for a specific location and just need a prompt to check the way they're inputing it. I think it would be really confusing to show them results because they're expecting something specific but we don't know what it is.

Situation 1 is where we can help with this change, by showing some results and guiding people to make a useful alert for themselves.

In this situation, I think it would be nice, rather than saying 'sorry search again for more specific results', say 'here's your results, now sign up for an alert with a street address', and give an example of a local street address as the placeholder. Something like this:

screen shot 2016-05-03 at 5 01 08 pm

That's a bit of a change from the way this PR works currently @CloCkWeRX . Currently you'll get through to the results page if you submit an unrecognised address in a recognised suburb (like in your example above). What do you think?

> This changes the hints for "Please enter a full street address like ..." to indicate a vaguer, street level address is OK

I think we still want to encourage people to submit a full address search, because they'll get the best results possible and the process to sign up for an alert is smoother. So I don't think we should start encouraging people to submit vaguer searches by changing the address example.

kat:
A big risk here is if people put in a vaguer address like a suburb they'd expect to see results for the whole suburb. The results they see are going to be the usual default radius which may or may not cover the suburb. I think that's a bigger more silent kind of failure, in some ways worse than getting no results. What else might we do instead to solve this given that we've staunchly stayed away from providing suburb results before. What do users want who are looking for a suburb's worth of results and perhaps more importantly, is that a use we want to serve?

CloCkWeRX:
>  What do users want who are looking for a suburb's worth of results and perhaps more importantly, is that a use we want to serve?

I'm actually a good example of a user who's interested in a bigger area, as I use it to update openstreetmap near me. It'd be really neat if the shapefiles for suburb boundaries (ABS data) or entire LGAs (some work in other pull requests that could easily be adapted to this) were supported and the queries were spatial/polygon based instead of just radius; or even a simple map selection UI (generate a bounding box)

equivalentideas:
>  the queries were spatial/polygon based instead of just radius; or even a simple map selection UI (generate a bounding box)

Could you do this for yourself using the API as is to test out how useful the results would be? Also I think there's a way to make an alert for a polygon existing that you could use if you want it as an email alert (not sure were this is documented sorry).

equivalentideas:
> What do users want who are looking for a suburb's worth of results and perhaps more importantly, is that a use we want to serve?

@kat I think the question before this is, are users you enter a suburb looking for a whole suburbs worth of results?

I get the impression that this isn't the case, and they're actually just responding to what they think they site is asking them to enter:

> Find out for free what is being built and knocked down **in your area** and receive email alerts when something new is planned.

> Email alerts of planning applications **near you**

When someone asks me where I live, or where I am, I'd say “North Bondi, Sydney”.

I start typing “Bondi” into the form and it helpfully suggests some options, which are all suburbs...

If this is the case and they're just locating themselves at the level they think is required, then the results can just be a taste of what's around, with a strong direction to subscribe for alerts. That's also what we're told people they can get, rather than “search for and discover all the planning applications in your suburb”.

If on the other had they are looking to see all the DAs in that suburb—to @kat’s question “is that a use we want to serve”, I'd say ‘no’ at this stage. The best way we can currently help people is to get them set up for alerts. I think that's the idea the site is built around at the moment.

So if this is the case, I agree with @kat that not letting these people through to results is more helpful to them.

This suggests other solutions to me. Rather than trying to adapt the search results for people who’ve gleaned the wrong mental model of the job the site can do for them, why not provide more help for them to format their search correctly?

* don't suggest that people look up a suburb using the auto complete
* if you start entering a suburb, the auto complete could suggest streets to reinforce that you should enter a street address
* make the 'use my location' option much more obvious to avoid manual entry as much as possible all together ( 20% of people who click that link subscribe for alerts, compared to 10% for people who submitted manual search).

Those options feel like small changes we could make and learn from? What do you think @kat and @CloCkWeRX ?

kat:
@equivalentideas good thinking.

equivalentideas:
With https://github.com/openaustralia/planningalerts/commit/691b2a683f55674f023d3781fb7097c311240e46 we've implemented the main changes that this lead us to:

> * don't suggest that people look up a suburb using the auto complete
> * if you start entering a suburb, the auto complete could suggest streets to reinforce that you should enter a street address

and we've seen a [big reduction in users seeing the error this addresses](https://github.com/openaustralia/planningalerts/issues/889#issuecomment-227948871).

I think that in the now rare case the people make a loose search, it's better for them to try and fix it on the search form page, rather than let them through to something else.

We thought that people wanted to search for suburbs, but that could well have been because we were suggesting it to people. Now we're not, so I think we wait to see what happens, and if we get this request from our core users.

I think we should close this PR without merging. Please reopen if you think otherwise.

Thanks so much for this contribution though @CloCkWeRX . It lead to coming up with some other ideas which I think have made the site better for everyone 💯

Issue #945, opened 2016-04-03
Add ability to manually fetch Councillor replies from WriteIt


equivalentideas:
Add the importer for grabbing Replies from WriteIt’s API. This is the first step to #895, in which we'll make this automatic using a webhook.

Issue #895, opened 2016-02-09
Allow replies from councillors via WriteIt to be automatically imported


henare:
Currently replies to comments to councillors are sent to a special mailbox and manually added to the site. When we switch to sending them via WriteIt (#894) and a reply to a message to a councillor is received WriteIt should send a webhook that PlanningAlerts receives and imports the message.

Issue #894, opened 2016-02-09
Allow comments to councillors to be sent via WriteIt


henare:
At the moment when someone creates a comment to a councillor we send the email directly from PlanningAlerts. We should have the option of posting it to the WriteIt API instead.

PR #950, opened 2016-04-04
WriteIt for WriteToCouncillors

henare:
This is a really hacky spike to get WriteIt integrated for Write To Councillors.

## TODO

- [x] Fetch a reply to a comment
    - [x] Store writeit message id on comment
    - [x] Get comment with “answers” from the writeit api
    - [x] Save answer as a reply
    - [x] Save the Writeit answer id on the reply so that we can check for dupes
    - [x] It shouldn't check for replies if we've already had a reply
- [x] Respond to an inbound webhook from WriteIt http://requestb.in/18yqlwu1?inspect
- [x] Deploy to test so we can fully test all the webhooks etc.
    - [x] When we deploy to test we'll need to let the webhooks through our http auth barrier (We need to use this URL http://nsw.test.planningalerts.org.au/comments/reply_webhook because we force SSL on the main test site and the certificate is broken)
- [x] Make reply loader accept multiple replies
- [x] Resolve TODOs in sending code
- [x] Massive tidy up of hacky code

### DIFF
--- a/.env
+++ b/.env
@@ -43,6 +43,12 @@ SECRET_KEY_BASE=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 # councillors associated with them.
 #COUNCILLORS_ENABLED=true
 
+# WriteIt configuration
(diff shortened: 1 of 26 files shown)


henare:
Looks like we're good to go :bouquet:

henare:
Fixes #945, #895, #894.

Issue #559, opened 2014-05-20
Create page for every council regardless of whether we have a scraper or not


henare:
I can't remember who came up with this idea, maybe @DylanFM?

It's a good one though, especially now we're fully on Morph. We can have nice little calls to action on each page where there's no scraper yet to get people to contact their council or write a scraper.

CloCkWeRX:
Going to base it off of https://stevage.cartodb.com/tables/lga_2015_aust/public

henare:
>CloCkWeRX self-assigned this

Woo! :tada:

I think it would be a very good idea as a first step to actually define what this issue means (always the hardest part). My guess is that by doing that we'll also discover that there are a few distinct things needed and we can then extract them into specific issues to work on.

e.g. I've been thinking it would be nice to load this data from Popolo (but maybe not as a first step?). Also how should we deal with defunct authorities (and how are we dealing with them now?).

CloCkWeRX:
Minimal mockups:
![authorities_mockup](https://cloud.githubusercontent.com/assets/365751/14224213/09b551ba-f8dc-11e5-95dc-d3487abaea31.png)

![mockup2](https://cloud.githubusercontent.com/assets/365751/14224214/09bb58d0-f8dc-11e5-8a78-0ce1e19f251f.png)

I'll have to look a bit closer to see how things like WA or VIC via SPEAR can be best rendered - was planning on setting the morph_name to the most appropriate one, but may introduce a superceded_by column instead if that's used more extensively.

![aggregated](https://cloud.githubusercontent.com/assets/365751/14224236/279656e2-f8dd-11e5-910f-bdb9f7d1e0c3.png)

LGAs don't really change too much, either merging or splitting on a year long time scale - evolving the ```disabled``` field from a bool into more of an ```end_date``` may well handle that a bit better. That would allow the council page to still be rendered, but with a notice.

![end_date](https://cloud.githubusercontent.com/assets/365751/14224237/2798041a-f8dd-11e5-886a-41ea70c636af.png)

Nice to haves: could stick the LGA boundaries into mysql if the spatial extensions are available; allowing us to render a map on the authority page / answer spatial questions...

mlandauer:
Still an extremely good idea

PR #965, opened 2016-04-23
Add superceded_by and lga_name15 modelling, so we can load all councils

CloCkWeRX:
For #559 we want to load in a bunch of council data.

I've broken this down into a few parts, so that we can look at the actual import of data separately; and work out what kind of UI we want.


henare:
See my comment [on the other PR](https://github.com/openaustralia/planningalerts/pull/967#issuecomment-234157027).

equivalentideas:
I'm going to close this as it's over a year without movement. Please feel welcome to re-open if need be :)

PR #966, opened 2016-04-23
Add lgas UI

CloCkWeRX:
Assumes #965 is accepted, part of #559 

In the original pull request.

> I don't think all the authorities that aren't covered should be intermingled with those that are. Maybe split the ones that aren't covered into a separate section. That way people looking to see what's covered can do that quickly and it's less of a change to that page.

Looks like:
image

Ignore the "(A)" etc on the names for the moment, that'll be addressed in data importing.

### DIFF
--- a/app/admin/authorities.rb
+++ b/app/admin/authorities.rb
@@ -22,10 +22,12 @@ ActiveAdmin.register Authority do
     attributes_table do
       row :full_name
       row :short_name
+      row :lga_name15
       row :state
       row :email
       row :write_to_councillors_enabled
       row :population_2011
+      row :superceded_by
       row :morph_name
       row :disabled
     end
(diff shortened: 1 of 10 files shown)


CloCkWeRX:
@henare  want to thrash out the detail of this one and see what you want out of the UI? Once we have this bedded down; even if we choose not to load data via rake task/generated sql/etc; we can at least start manually adding 'em in.

henare:
@CloCkWeRX thanks, that's a very good idea :+1: @equivalentideas is the expert here so I'll pass it over to him.

mlandauer:
This has been open a long time without merging. Also, the coverage page has changed a bit since this PR was written. So, closing this. Sorry.

PR #967, opened 2016-04-23
Add lgas data generation tool

CloCkWeRX:
#559

Assumes other two pull requests are suitable

Usage:

```
bundle exec rake data:fetch_lgas
bundle exec rake data:generate_sql > authorities_to_import.sql
```

Basically, this needs a human to carefully review the matches in small batches; to make sure we don't create duplicates or update the wrong council.

Sample output: _be warned, it doesn't escape commas etc and may not be well formed sql._
https://gist.github.com/CloCkWeRX/23249634c072d4101968d6348a330db6

### DIFF
--- a/app/admin/authorities.rb
+++ b/app/admin/authorities.rb
@@ -22,10 +22,12 @@ ActiveAdmin.register Authority do
     attributes_table do
       row :full_name
       row :short_name
+      row :lga_name15
       row :state
       row :email
       row :write_to_councillors_enabled
       row :population_2011
+      row :superceded_by
       row :morph_name
       row :disabled
     end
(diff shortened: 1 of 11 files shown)


henare:
Me, from Slack:

>I'm going to remove my assignment on [these] PlanningAlerts pull requests to load all councils. I'll keep the PRs open because I think they're definitely still worth using but I don't have the time at the moment. I think the hardest thing for me is not understanding how the rollout would happen in production so I'm wary of the amount of time it would take me. Sorry, I really hope that when we get to that we can still use them though

equivalentideas:
I'm going to close this as it's over a year without movement. Please feel welcome to re-open if need be :)

Issue #969, opened 2016-04-27
Load councillors should use cached image if it's present


henare:
https://morph.io/openaustralia/australian_local_councillors_images caches images from Popolo on S3 so that the files are served over HTTPS and reliably available. Currently when we load councillors we just use the image URL from Popolo and we need to switch to these cached images instead.

henare:
@equivalentideas I've assigned this to you in case you're looking for something to work on. It's the next blocker we need done.

PR #970, opened 2016-04-27
Admin instructions for write to councillors

henare:
This is the start of modifying the readme to make it more useful for admins adding councillors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,16 +110,16 @@ You can see which authorities have the feature enabled at the Authorities admin 
 
 If you'd like to add new councillors for a planning authority, there are a number of steps that you need to take across a few different online services.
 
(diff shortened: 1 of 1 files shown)


henare:
I think this is ready to go. We still need #969 done before we merge it though.

@equivalentideas I'm assigning it to you to review as you work through adding councillors today.

README.md
```diff
@@ -106,12 +108,16 @@ You can see which authorities have the feature enabled at the Authorities admin
 
 #### Adding councillors for an authority
```
equivalentideas: The current structure of this section makes it a bit hard to follow as a check list, which is how use use it as you work through the steps.

I think front loading the key actions in each par, and even using bold emphasis to highlight the steps could help.

I'll give that a go.

henare: :ok_hand:

equivalentideas:
> I think front loading the key actions in each par, and even using bold emphasis to highlight the steps could help.

I've done this and added a little more explanation where I think it's needed based on my experience of going through the process.

I'll merge this and we can continue to iterate. I wont wait for tests since this only changes the readme.

PR #971, opened 2016-04-28
Load cached images for councillors when possible

### LINKED ISSUE #969: Load councillors should use cached image if it's present

equivalentideas:
This adds a little logic to `Authority#load_councillors` so that we store the url for the cached S3 image of a councillor if it's available. Closes  #969

If the image isn't available, then it falls back to the whatever is stored in the popolo source data.

Could you please review this for me @henare ?


henare:
Nice work getting through this today @equivalentideas :+1: 

>If the image isn't available, then it falls back to the whatever is stored in the popolo source data.

I'm not sure this is the best approach. If we use URLs from the source data here and there then PlanningAlerts will be using inconsistent URLs and ones we don't control.

I'd like to consistently know that we're using images from a good, reliable source so I think we should change this to only use URLs from our cache.

app/models/authority.rb
```diff
@@ -261,7 +261,16 @@ def load_councillors(popolo)
 
     persons.map do |person|
       councillor = councillors.find_or_create_by(name: person.name)
-      councillor.update(email: person.email, image_url: person.image, party: person.party, popolo_id: person.id)
+
+      councillor.popolo_id = person.id
+
+      councillor.image_url = if councillor.cached_image_available?
```
henare: Using the ternary operator here will be more concise and clear. Of course if my suggestion about not falling back to the popolo url is accepted then this doesn't matter :smile:

app/models/authority.rb
```diff
@@ -261,7 +261,16 @@ def load_councillors(popolo)
...
+        councillor.cached_image_url
+      else
+        person.image
+      end
+
+      councillor.update(email: person.email, party: person.party)
```
henare: Will this also persist a change from setting the `popolo_id` above?

app/models/councillor.rb
```diff
@@ -13,4 +13,12 @@ def prefixed_name
   def writeit_id
     authority.popolo_url + "/person/" + popolo_id
   end
+
+  def cached_image_url
+    "https://australian-local-councillors-images.s3.amazonaws.com/#{popolo_id}.jpg"
+  end
+
+  def cached_image_available?
+    Net::HTTP.get_response(URI(cached_image_url)).kind_of? Net::HTTPSuccess
```
henare: To avoid pulling the whole image down we should do a HEAD request instead.

equivalentideas:
> If we use URLs from the source data here and there then PlanningAlerts will be using inconsistent URLs and ones we don't control.

> I'd like to consistently know that we're using images from a good, reliable source so I think we should change this to only use URLs from our cache.

That sounds really good to me @henare I'm make that change.

app/models/authority.rb
equivalentideas: Yep, like the `image_url` being assigned. This is because the [#update method](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update) just calls [#save](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save) on the record, persisting all changed attributes.

henare: Shiny, I didn't know that - thanks for checking :sunglasses:

equivalentideas:
I've updated this to only save an image_url if we have a cached image. I agree this is much nicer for us.

I've also changed the checking method to only do a HTTP HEAD request like you suggested @henare , a nice improvement 👍 .

Did you have any other feedback on this @henare or do you think we can merge?

henare:
Awesome, great work! :guitar: Can I leave this for you to deploy? :rocket:

PR #979, opened 2016-05-12
Add a Alert sign up form to authority applications pages

### LINKED ISSUE #977: Add a sign up form to authority/applications pages
A lot of organic search lands on pages like [`authorities/brisbane/applications/`](https://www.planningalerts.org.au/authorities/brisbane/applications), about 6% of users or currently about 1300 people per week.

These people are probably interested in planning applications and would like to get alerts about planning in their area.

We might help more people subscribe for alerts by putting a sign up form on this page.

## How will we know if this helps?

In the last month only 1.11% percent of people who landed on “Recent applications for authority” pages signed up for alerts.

If this is a good change then we should expect to see that increase towards 1.88% (hopefully past), the conversion rate for all users (this would be about an extra 10 people signing up for PlanningAlerts a week at 1200 sessions a week).

### LINKED ISSUE #997: When I'm looking at a DA, tell me if the address is heritage listed, so I better understand the impact and context of the development
This might be useful for this http://api.heritagecouncil.vic.gov.au/documentation/VHD-v1

equivalentideas:
Closes #977 to help people sign up for alerts from the Authority recent applications pages. Over a thousand people a week hit this page and many of them might like to sign up for alerts.

It also adds:

* integration tests for the main alert sign up scenarios
* DRY’s up the views that included the Alert signup form

screen shot 2016-05-12 at 5 37 01 pm

### DIFF
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -13,6 +13,7 @@ class ApplicationsController < ApplicationController
       apps = @authority.applications
       @description << " from #{@authority.full_name_and_state}"
       @alert = Alert.new
+      @alert.address_for_placeholder = apps.last.address if apps.any?
     else
       @description << " within the last #{Application.nearby_and_recent_max_age_months} months"
       apps = Application.where("date_scraped > ?", Application.nearby_and_recent_max_age_months.months.ago)
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -12,6 +12,7 @@ class ApplicationsController < ApplicationController
       @authority = Authority.find_by_short_name_encoded!(params[:authority_id])
       apps = @authority.applications
       @description << " from #{@authority.full_name_and_state}"
+      @alert = Alert.new
     else
       @description << " within the last #{Application.nearby_and_recent_max_age_months} months"
       apps = Application.where("date_scraped > ?", Application.nearby_and_recent_max_age_months.months.ago)
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -12,14 +12,14 @@ class ApplicationsController < ApplicationController
       @authority = Authority.find_by_short_name_encoded!(params[:authority_id])
       apps = @authority.applications
       @description << " from #{@authority.full_name_and_state}"
-      @alert = Alert.new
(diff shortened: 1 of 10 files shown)


app/controllers/applications_controller.rb
```diff
@@ -12,6 +12,8 @@ def index
       @authority = Authority.find_by_short_name_encoded!(params[:authority_id])
       apps = @authority.applications
       @description << " from #{@authority.full_name_and_state}"
+      @alert = Alert.new
+      @alert.address_for_placeholder = apps.last.address if apps.any?
```
equivalentideas: Why should this be restricted to when there is an authority. Why not always show it on the applications page? Hardly anyone hits the applications index page, but if they do the same logic about signing up from there holds I think. I'll remove the form from the conditional logic to simplify this.

equivalentideas:
I've double checked through this and it looks good to me. I'll merge and deploy and keep an eye on it's impact.

henare:
Marvellous :tada: Any idea if this has had an impact?

equivalentideas:
> Marvellous 🎉 Any idea if this has had an impact?

@henare It's been a few weeks now so lets have a quiz.

Turns out I thought I little about measuring this before I did it ⭐ :

> ## How will we know if this helps?
>
> In the last month only 1.11% percent of people who landed on “Recent applications for authority” pages signed up for alerts.
> 
> If this is a good change then we should expect to see that increase towards 1.88% (hopefully past), the conversion rate for all users (this would be about an extra 10 people signing up for PlanningAlerts a week at 1200 sessions a week).

So this is looking at the conversion rate of people who lang on this authority/application pages, where we've added this form. We track all confirmed alert conversions under the goal "Register for Alert via the home page" (confusing I know).

[We made another change that may have improved conversion on the same day](https://github.com/openaustralia/planningalerts/issues/889#issuecomment-227948871), so take that into account.

screen shot 2016-06-23 at 3 30 13 pm

So before we made this change, about 1.2% of people who landed on the impacted page ended confirmed an alert sign up each week. Since we made the change it now sits around 5%. Much higher than the 1.8% I was hoping for.

That's about an extra 30+ people signing up each week, about a 7% increase over all.

Remember, I dropped out of maths—but it looks like this has had a nice positive impact for users 👍 .

henare:
This data is smoking :fire:

Issue #985, opened 2016-06-01
the list of councillors makes them look corporate and unapproachable


kat:

equivalentideas:
I think we've made a big step forward on this with #996 , moving from

![screen shot 2016-06-09 at 11 16 32 am](https://cloud.githubusercontent.com/assets/1239550/15951032/fc979c98-2ef7-11e6-9f3b-1c3730219c93.png)

to

![screen shot 2016-06-10 at 9 41 04 am](https://cloud.githubusercontent.com/assets/1239550/15951036/014c1336-2ef8-11e6-93ac-fc0fd14c0efc.png)

I'm gonna close this with the expectation that we open new issues for iterations. @kat feel free to reopen this as well if you think it should be :)

PR #996, opened 2016-06-10
Make councillors more approachable, clearer and councillor list more usable

### LINKED ISSUE #929: List of councillors to write to is really long
I've got a massive screen and when I want to write to my councillor they take up the whole screen and they're the only thing on it:

308-314 stanmore road petersham nsw 2049 19 mar 2016 planningalerts - mozilla firefox_002

Maybe we should put them into two columns?

### LINKED ISSUE #973: Councillor list looks confusing when some councillors don't have images and some do
The layout becomes wonked up. See [this example](https://www.planningalerts.org.au/applications/648617):

screen shot 2016-04-29 at 11 57 59 am

One solution could be to always have a placeholder image to fallback on.

### LINKED ISSUE #986: make it less fiddly to select a councillor
eg. I had no idea you could select the image- as the image and button don't seem connected. There's no signifier.

equivalentideas:
This is an iteration on the presentation of councillors to:

* Make the councillors list more friendly, human and approachable by making their images circles, giving them a friendly yellow if they have no image (consistent yellow colour to match councillor comments), and makes their names bigger and clearer ( first pass on #985 ),
* If councillors don’t have images, it gives them a friendly yellow avatar with their initials. This means the all the councillors have a consistent layout whether they have images or not both in their replies and in the councillor select list ( fixes #973 ).
* Wraps the list round into two columns if there's space making the list more compact on widescreens ( fixes #929 )
* Makes it clearer you can click the nice big faces by giving them a hover effect and hidding the radio button, so you don't think you need to fiddle with the radio button (fixes #986 )
* Makes the councillors names bigger and clearer
* Makes the list fit a bit better on small screens
* refactors some of the councillor styles

## Before

screen shot 2016-06-09 at 11 16 32 am

screen shot 2016-06-09 at 11 16 24 am

## After

screen shot 2016-06-10 at 9 41 04 am

screen shot 2016-06-09 at 11 13 27 am

### DIFF
--- a/app/models/councillor.rb
+++ b/app/models/councillor.rb
@@ -6,6 +6,12 @@ class Councillor < ActiveRecord::Base
   validates :authority, :email, presence: true
   validates :image_url, format: { with: /\Ahttps/, message: "must be HTTPS" }, allow_blank: true
 
+  def initials
+    first_letters = name.split(/\s/).collect { |word| word.first.upcase }
+    first_letters = first_letters.values_at(0, -1) unless first_letters.one?
+    first_letters.join(" ")
+  end
+
   def prefixed_name
     "local councillor #{name}"
   end
--- a/app/assets/stylesheets/partials/_comments.scss
+++ b/app/assets/stylesheets/partials/_comments.scss
@@ -218,28 +218,6 @@ ol#comments {
   .councillor-image {
     width: 4em;
     height: 4.5em;
-    overflow: hidden;
-    border-radius: .25em;
-    background-color: $lightest-cool-highlight;
-  }
-
-  .councillor-initials {
-    font-size: 1.5em;
-    text-transform: uppercase;
-    display: block;
-    width: 100%;
-    height: 100%;
-    border: none;
-    text-decoration: none;
-    text-align: center;
-    display: flex;
-    align-items: center;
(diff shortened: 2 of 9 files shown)


app/models/councillor.rb
```diff
@@ -6,6 +6,12 @@ class Councillor < ActiveRecord::Base
   validates :authority, :email, presence: true
   validates :image_url, format: { with: /\Ahttps/, message: "must be HTTPS" }, allow_blank: true
 
+  def initials
```
equivalentideas: This feels a bit unclear cause it's not just 'initials' it's also formatting them specifically for our view.

Should the formatting step to split into a view helper?

equivalentideas: I'm going to leave this as is. I think it's clear from the tests and implementation what it does and if it get's used elsewhere and needs to be made more flexible we can change it then.

equivalentideas:
I've done some guerrilla testing of this with a few people and it's definitely easier to use. I think it's a step in the right direction and solves a bunch of our bugs with the current layout.

I raised one issue above with the new Councillor#initials method, but I think that can be refactored/improved when it's touched again.

I'll merge and deploy.

PR #1006, opened 2016-06-22
Fix resending duplicate comments

### LINKED ISSUE #981: Duplicate messages created in WriteIt from a single comment
Sometime a comment is posted twice to the WriteIt API. This means that when the councillor replies, there is a 50/50 chance that their message will go to a message that the original PlanningAlerts comment does not know about. This means the reply cannot be found and loaded (discovered in #980).

It looks like CommentNotifier#send_comment_via_writeit! is failing after it has posted the comment to WriteIt. There are no checks to see if a comment exists, so it just DelayedJob tries to run the whole method again and posts a new WriteIt message to the API and updates the comment’s `writeit_message_id`.

One solution to this might be to do some kind of check to see if the comment is already on WriteIt. If it does then we would make sure we've recorded the correct `writeit_message_id` on the comment. But why is it failing in the first place?

henare:
Fixes #981.

This looks like a lot of change to fix what seems like a simple problem. That's because it is. I got carried away refactoring. I'm pretty happy that the code is in a much better place thanks to it but it doesn't make this very easy to review. I'm sorry.

There's even a tiny drive-by improvement mixed in here. I'm very naughty, I know. In my defence it's the first code I've been let loose on since holidays.


spec/features/write_to_councillors_spec.rb
```diff
@@ -173,5 +173,12 @@
       expect(page).to have_content "Matthew Landauer wrote to local councillor Louise Councillor"
       expect(page).to have_content "Delivered to local councillor Louise Councillor"
     end
+
+    context "twice" do
+      scenario "councillor doesn't get two messages" do
+        comment.confirm!
```
equivalentideas: This should be in the context setup, not the scenario.

spec/models/comment_spec.rb
```diff
@@ -3,6 +3,17 @@
 describe Comment do
   it_behaves_like "email_confirmable"
 
+  describe "#confirm!" do
+    let(:comment) do
+      VCR.use_cassette('planningalerts') { build(:comment, confirmed: true) }
+    end
+
+    it "should run callback once already confirmed" do
```
equivalentideas: I find this description confusing. Is there a typo? Could it be that the test describes something not happening, but the description says something should happen?

equivalentideas: I think I get what this is about. I think a context, like "when already confirmed", might make this clearer.

app/models/comment.rb
```diff
@@ -26,8 +26,14 @@ class Comment < ActiveRecord::Base
...
-  # Send the comment to the planning authority
-  def after_confirm
+  def confirm!
+    unless confirmed
+      update!(confirmed: true, confirmed_at: Time.current)
+      send_comment!
+    end
+  end
+
+  def send_comment!
```
equivalentideas: This is much nicer 👍

app/models/comment.rb
```diff
@@ -44,7 +44,7 @@ def send_comment!
   end
 
   def to_councillor?
-    councillor ? true : false
+    councillor.present?
```
equivalentideas: 💖

spec/models/comment_spec.rb
equivalentideas: I'm gonna leave this one to you @henare , I had a quick look at it but couldn't discern what the intention is.

equivalentideas:
Nice one @henare this is a bit improvement I think. There's just one outstanding thing that's unclear https://github.com/openaustralia/planningalerts/pull/1006#discussion_r68164535

We also chatted about covering the validations `EmailConfirmable`. I'm gonna leave that for now and come back around if I get through other stuff in next could of days.

spec/models/comment_spec.rb
henare: Yup, there's a typo. @equivalentideas the context makes it way more readable, thanks :ok_hand:

PR #1028, opened 2016-09-04
Testfix/rails 4 2 7

davwood:
I assume this is the correct branch to branch off. 

The introduction of Active Job in Rails 4.2 is useful and since you're using delayed_job I set that up so that we can just use `deliver_now` or `deliver_later`.

The writeItApi post I currently have as `deliver_now` so that the test passes but should be fine to rewrite the test and use `deliver_later` as well.


davwood:
Note for future in rails 5.
This section https://github.com/openaustralia/planningalerts/blob/rails-4-2-7/app/mailers/comment_notifier.rb#L32 can be improved with this update.

http://blog.bigbinary.com/2016/01/08/rendering-views-outside-of-controllers-in-rails-5.html

henare:
@davwood excitement! :sparkles: 

I'll assign @equivalentideas since it's merging into his PR :zap:

spec/features/reply_from_councillor_spec.rb
```diff
@@ -1,10 +1,8 @@
 require "spec_helper"
 
 feature "Councillor replies to a message sent to them" do
-  given (:councillor) { create(:councillor, name: "Louise Councillor", email: "louise@council.nsw.gov.au") }
-  given (:application) { create(:application, id: 8, address: "24 Bruce Road Glenbrook", description: "A lovely house") }
-  given (:comment) do
   let(:councillor) { create(:councillor, name: "Louise Councillor", email: "louise@council.nsw.gov.au", popolo_id: "marrickville_council/chris_woods") }
+  let(:application) { create(:application, id: 8, address: "24 Bruce Road Glenbrook", description: "A lovely house") }
```
equivalentideas: Hey @davwood I think we did it this way to follow the [Capybara “descriptive” DSL](https://github.com/jnicklas/capybara#using-capybara-with-rspec). I kinda think it is nicer to be consistent through the tests and use let, before and describe here as you've changed—but we’ve used this DSL through the integration tests, so I think we should stick to it or change them all. What do you think?

spec/features/reply_from_councillor_spec.rb
```diff
@@ -14,9 +15,8 @@
     end
   end
 
-  background { comment.confirm! }
-
   scenario "it's entered by an admin and displayed on the application page" do
+    comment.confirm!
```
equivalentideas: @davwood this looks fine, but could you explain why this change was necessary? That would be a great thing for the commit message ;-)

Gemfile.lock
```diff
@@ -251,7 +251,7 @@ GEM
       activesupport (>= 3.0.0)
     launchy (2.4.3)
       addressable (~> 2.3)
-    libv8 (3.16.14.11)
+    libv8 (3.16.14.15)
```
equivalentideas: Nice 👍

equivalentideas:
@davwood Hey Dave, this is great thanks so much for the contribution. I had a few questions that I've left on the commits. If you could have a look at them that would be awesome :)

spec/features/reply_from_councillor_spec.rb
davwood: @equivalentidea.  Got it. Hadn't used that syntax before so thought it must have been old. :)
Reverted back now to use `given` and `background`

spec/features/reply_from_councillor_spec.rb
davwood: this was in a background block but only needed to be run for the first test - hence why it's here.

The 2nd test i.e. `it’s loaded from WriteIt by an admin` is already started with a confirmed comment.

spec/features/reply_from_councillor_spec.rb
```diff
@@ -13,9 +13,8 @@
     end
   end
 
-  background { comment.confirm! }
```
davwood: More info re: removing this background here - for the WriteIt test there are 2 http requests potentially here:
1) When creating a comment 
GET http://maps.google.com/maps/api/geocode/json?address=24%20Bruce%20Road%20Glenbrook&region=au&sensor=false
2) When confirming a comment that is for WriteIt
POST http://writeit.ciudadanointeligente.org/api/v1/message/

These are currently in the same VCR cassette (planningalerts) and so would need to record a new cassette. Seemed simpler this way for now (and less changes).

As the confirm wasn't actually being tested for that test I avoided

davwood:
@equivalentideas thanks for taking a look.
made those changes back to let and a few more comments. Let me know what you think.

spec/features/reply_from_councillor_spec.rb
equivalentideas: That's awesome @davwood thanks for explaining. In future would be cool to chuck that kinda thing on the commit message :)

spec/features/reply_from_councillor_spec.rb
equivalentideas: ⭐

equivalentideas:
Thanks heaps for picking this up @davwood I'll have another look at this asap and hopefully merge.

equivalentideas:
Great work on this @davwood it's really clear. I'm merging 🚢

Issue #887, opened 2016-01-12
Chart y axis is weird occasionally


equivalentideas:
On the 'Comments over time' chart of Scenic Rim Regional Council, the number labels on the y axis are wrong. There should be 3 ticks with "0", "1" and "2", but instead there are 5 ticks with "0", "00m", "1", "1.5" and "2".

![screen shot 2016-01-12 at 4 45 39 pm](https://cloud.githubusercontent.com/assets/1239550/12255823/006440f0-b94c-11e5-9bd8-3dd21cd5e01a.png)

henare:
[Confirmed fixed](https://www.planningalerts.org.au/authorities/scenic_rim) in #1033.

Issue #1003, opened 2016-06-15
Address autocomplete suggests splitting the street number from the street name


henare:
e.g. _123**,** King St, Newtown_ instead of _123 King St, Newtown_

![tooltip_002](https://cloud.githubusercontent.com/assets/48945/16069348/05d947f8-3311-11e6-8bb4-2b7661b8dac4.png)

I don't think this is having a massive impact but it's weird.

davwood:
@henare take a look at the PR I just referenced to fix this when you get a chance.

PR #1033, opened 2016-09-21
allow for 2 ticks when max value is less than 3

davwood:
@henare @equivalentideas
This fixes the issue here:
https://github.com/openaustralia/planningalerts/issues/887

Basically when the max Y axis value was 2 - the 'yTickCount' was being set to 3. This was causing the yAxis to show weird numbers to try to fill 3 ticks when there were only going to be 2. Would have thought d3 would have just shown this more gracefully.. maybe in version 4. :)

Anyway I did a quick jsbin so could play around with the real data and see that the change works:
http://jsbin.com/zuqimuhocu/edit?html,js,console,output

Take a look when you have time.

### DIFF
--- a/app/assets/javascripts/bar_graph.js
+++ b/app/assets/javascripts/bar_graph.js
@@ -39,9 +39,9 @@ function barGraph(selector, url, metric) {
     var bisectDate = d3.bisector(function(d) { return d.key; }).left
 
     var yTickCount;
-    if (maxYValue < 2) {
-      yTickCount = 1;
-    } else  {
+    if (maxYValue < 3) {
+      yTickCount = Math.floor(maxYValue)
+    } else {
       yTickCount = 3;
     }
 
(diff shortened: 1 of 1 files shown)


henare:
Very nice @davwood, you're on :fire: 

Looks good in the jsbin and doesn't seem to cause regressions locally so I'll merge and deploy :shipit:

henare:
@davwood my tip of the day:tm: is if you write "Fixes #887" in a pull request then it automatically closes the issue (works in commit messages too, as you may know) :smiley:

davwood:
@henare great! 

Thanks for the tip. Hadn't used github issues actively before. That is handy.

davwood:
On that note @henare. Can you then close this one? It's still open
https://github.com/openaustralia/planningalerts/issues/1003

henare:
Duh, I thought I'd done that - thanks for the reminder @davwood.

PR #1039, opened 2016-11-02
Track unsubscribers

equivalentideas:
Currently in our performance api we’re tracking data about new subscribers, but not unsubscribes.

We're interested in knowing about unsubscribes because:

1. we'll soon be working on “PlanningAlerts Subscribers” where we want people to keep using the project so they continue to support it.
2. we think we might see things in this data that will help us make the project more useful to people.

This adds the number of people who completely leave PlanningAlerts each day (unsubscribe their last remaining alert). It also fixes a bug in the subscribers data where it only counted people if they were still subscribed.


equivalentideas:
This passes tests, so we're merging and 🐑 ing

PR #1041, opened 2016-11-02
Database independence

crowbot:
Hiya! 

Have been working with planningalerts for a few months for a local project - it's great! I was running off the mysociety fork, but I think it's going to make more sense to work from master, so that I can push stuff back to you guys. Here's a first PR - how do you feel about making the project a bit more db-agnostic? Probably more changes to come here if you're amenable to the idea.

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -157,7 +157,11 @@ class Alert < ActiveRecord::Base
     comments = []
     # Doing this in this roundabout way because I'm not sure how to use "near" together with joins
     applications_with_new_comments.each do |application|
-      comments += application.comments.visible.where('comments.confirmed_at > ?', cutoff_time)
+      comments += application
+                    .comments
+                      .visible
+                        .where('comments.confirmed_at > ?', cutoff_time)
+                          .order('created_at')
     end
     comments
   end
@@ -166,7 +170,10 @@ class Alert < ActiveRecord::Base
     replies = []
     # Doing this in this roundabout way because I'm not sure how to use "near" together with joins
     applications_with_new_replies.each do |application|
-      replies += application.replies.where('replies.received_at > ?', cutoff_time)
+      replies += application
+                  .replies
+                    .where('replies.received_at > ?', cutoff_time)
+                      .order('created_at')
     end
     replies
   end
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -48,8 +48,10 @@ class Alert < ActiveRecord::Base
   end
 
   def self.count_of_new_unique_email_created_on_date(date)
-    alerts = where(confirmed: true).where("date(created_at) = ?", date).group(:email)
-
+    alerts = where(confirmed: true).where("date(created_at) = ? AND id =
+                                           (SELECT min(a.id) FROM alerts a
+                                            WHERE a.email = alerts.email
+                                            AND date(a.created_at) = ?)", date, date)
     alerts.reject do |alert|
       where(email: alert.email).where("created_at < ?", alert.created_at).any?
     end.count
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -157,11 +157,11 @@ class Alert < ActiveRecord::Base
     comments = []
     # Doing this in this roundabout way because I'm not sure how to use "near" together with joins
     applications_with_new_comments.each do |application|
-      comments += application
-                    .comments
-                      .visible
-                        .where('comments.confirmed_at > ?', cutoff_time)
-                          .order('created_at')
+      comments += application.
+                  comments.
+                  visible.
+                  where('comments.confirmed_at > ?', cutoff_time).
+                  order('created_at')
     end
     comments
   end
@@ -170,10 +170,10 @@ class Alert < ActiveRecord::Base
     replies = []
     # Doing this in this roundabout way because I'm not sure how to use "near" together with joins
     applications_with_new_replies.each do |application|
-      replies += application
-                  .replies
-                    .where('replies.received_at > ?', cutoff_time)
-                      .order('created_at')
+      replies += application.
+                 replies.
+                 where('replies.received_at > ?', cutoff_time).
+                 order('created_at')
     end
     replies
   end
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -175,13 +175,25 @@ class Authority < ActiveRecord::Base
   def applications_per_week
     # Sunday is the beginning of the week (and the date returned here)
     # Have to compensate for MySQL which treats Monday as the beginning of the week
-    h = applications.group("CAST(SUBDATE(date_scraped, WEEKDAY(date_scraped) + 1) AS DATE)").count
+    h = by_week_from_sunday(applications, 'date_scraped')
     min = h.keys.min
     max = Date.today - Date.today.wday
     (min..max).step(7) do |date|
       h[date] = 0 unless h.has_key?(date)
     end
     h.sort
+  end
+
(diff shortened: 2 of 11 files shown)


henare:
Wonderful to see you here @crowbot :wave: 

In principle I'm definitely keen on this. I think we had an idea years ago to switch to Postgres (because of PostGIS?) and we're certainly keen for our projects to be reused. I also personally like the concept of Rails applications being database agnostic.

I'll need to spend some time looking at this though and having more of a think about it so it might take me up to a week to look at this.

In the mean time can we add Postgres to Travis in this PR?

henare:
>OK, have pushed a fix for the spec failures...

Woohoo, nice one!

>...maybe you'd rather have an explicit order on the associations though? 

It looks like the method in question is used in alert emails so you'd expect the order was important. However reading the code I think the order of comments and replies actually goes by the date_received of the application, not the comment or reply.

That seems weird. My guess is that we never really thought about it. I think we should get the same thing on both databases though so we should probably pick an order and be explicit about it.

Keeping the same order by application date_received seems like the dumb simple thing to do and keeps comments grouped by application, which is nice. So does that just mean we need to be explicit about ordering by created_at for [comments](https://github.com/openaustralia/planningalerts/blob/d5bbd4fd0f94e35851689e08926ba1036acd8ddc/app/models/alert.rb#L160) and [replies](https://github.com/openaustralia/planningalerts/blob/d5bbd4fd0f94e35851689e08926ba1036acd8ddc/app/models/alert.rb#L169)?

equivalentideas:
> This PR now fixes all the immediate database-specific bits I've found. 

Nice one @crowbot 👍 Let us know when this is ready for review :)

equivalentideas:
So weird, I thought that would appear on the actual commit https://github.com/openaustralia/planningalerts/pull/1041/commits/d0f4e0611868e04f258b0b2621184ec14e96849f that's a comment for that comment :)

spec/models/authority_spec.rb
```diff
@@ -102,6 +102,48 @@
...
+      Timecop.return
+    end
+
+    context "when the authority has no applications" do
+      it "returns an empty list" do
+        expect(authority.applications_per_week).to eq []
+      end
+    end
+
+    context "when the authority has applications" do
```
equivalentideas: Fantastic ❤️ great to get this covered!

There are a few extra whitespace lines in here compared to our style (123, 125, 144). Could you snip em please? And there’s a line missing between blocks (between 136 and 137) :)

app/models/authority.rb
```diff
@@ -175,7 +175,7 @@ def median_applications_per_week
   def applications_per_week
     # Sunday is the beginning of the week (and the date returned here)
     # Have to compensate for MySQL which treats Monday as the beginning of the week
```
equivalentideas: This comment can probably be moved with the extraction, or removed 👍

app/models/authority.rb
```diff
@@ -175,7 +175,7 @@ def median_applications_per_week
...
-    h = applications.group("CAST(SUBDATE(date_scraped, WEEKDAY(date_scraped) + 1) AS DATE)").count
+    h = by_week_from_sunday(applications, 'date_scraped')
```
equivalentideas: This is awesome @crowbot 👍 it simplifies to method we are about even without this extra work flexibility.

equivalentideas: One thing, I think the method name should make it clear this is going to return a count, rather than groups of records 💯

app/models/authority.rb
```diff
@@ -176,6 +176,7 @@ def applications_per_week
     # Sunday is the beginning of the week (and the date returned here)
     # Have to compensate for MySQL which treats Monday as the beginning of the week
     h = by_week_from_sunday(applications, 'date_scraped')
+    return [] if h.empty?
```
equivalentideas: I actually think the comments_per_week version of this is much more readable. Do you think we should make this one consistent with that while we're here?

app/models/alert.rb
```diff
@@ -157,7 +157,11 @@ def new_comments
     comments = []
     # Doing this in this roundabout way because I'm not sure how to use "near" together with joins
     applications_with_new_comments.each do |application|
-      comments += application.comments.visible.where('comments.confirmed_at > ?', cutoff_time)
+      comments += application
```
equivalentideas: 👍 Is this because Postgres defaults to some other order?

I can't find an example in PlanningAlerts, but here's an example of our preferred style for wrapping methods over lines like this https://github.com/mysociety/alaveteli/pull/3544/files#diff-4676c008b11a5480d73d4a6de01e45b9R216

I think splitting this over lines if definitely an improvement in readability 👍

crowbot: I think as the order is not specified, postgres sorts by some internal criteria - although the spec doesn't always fail under postgres, you can see some failing tests in https://travis-ci.org/openaustralia/planningalerts/jobs/173758076

app/models/alert.rb
```diff
@@ -48,8 +48,10 @@ def self.alerts_in_inactive_areas
   end
 
   def self.count_of_new_unique_email_created_on_date(date)
-    alerts = where(confirmed: true).where("date(created_at) = ?", date).group(:email)
-
+    alerts = where(confirmed: true).where("date(created_at) = ? AND id =
```
crowbot: This query as originally generated won't run in Postgres - it produces the error:
```
PG::GroupingError: ERROR:  column "alerts.id" must appear in the GROUP BY clause or be used in an aggregate function
```
What's happening is that we're asking the database for some rows (confirmed, created on a specific date) and then we're asking for those to be grouped into one row, based on the value of the email field - essentially we want only one row for each email address. But we're not telling the database what to do about fields that are different in the records in each group - so, if there are two records, do we want the grouped record to give us the lower or higher id field? MySQL is just giving us a record in this case, based on some internal logic of its own - we haven't explicitly specified what we want. Postgres won't run the query because we haven't specified what we want. There seems to be some debate about whether MySQL is being non-standards compliant in running the query, but in any case, it seems safer to specify exactly that we _actually_ want the row with the first comment on that day associated with that email address.

equivalentideas: That's great @crowbot 👍 I'm going to amend that into the commit message.

config/database.yml.example
```diff
@@ -14,25 +14,47 @@
...
+  pool: 5
+  username: postgres
+  password:
+  database: pa_test
+
+# Warning: The database defined as "test" will be erased and
+# re-generated from your development database when you run "rake".
+# Do not set this db to the same as development or production.
+test: &TEST
+  <<: *<%= ENV['DB'] || "mysql" %>
```
equivalentideas: Should there also be a note to update this line in the readme?

equivalentideas:
> Thanks for the review @equivalentideas - have pushed some fixup commits.

Looking great @crowbot 

I've just tried to set this up locally from scratch following the readme. When I run `bundle exec rake db:setup`, after the migrations run, I get:

```
rake aborted!
Errno::ECONNREFUSED: Connection refused - connect(2)
/Users/Luke/git/planningalerts-postgres/db/seeds.rb:1:in `<top (required)>'
/Users/Luke/.rbenv/versions/2.0.0-p353/bin/bundle:23:in `load'
/Users/Luke/.rbenv/versions/2.0.0-p353/bin/bundle:23:in `<main>'
Tasks: TOP => db:setup => db:seed
(See full trace by running task with --trace)
```

Have I missed an implied step or maybe there's more config to change somewhere?

mlandauer:
It pains me to close a PR from you @crowbot without merging it but this has been open for too long. Any chance that you or someone else could redo this against current master?

PR #1043, opened 2016-11-02
Theme encapsulation

crowbot:
Hello again! I'd like to add a new theme, and wondering how to go about that - should it go in this repo? Opening up a PR for some early review of the approach if so.

### DIFF
--- a/.env
+++ b/.env
@@ -21,6 +21,8 @@ SCRAPE_DELAY=5
 
 COUNTRY_CODE=AU
 
+EXAMPLE_ADDRESS="1 Sowerby St, Goulburn, NSW 2580"
+
 DEVISE_SECRET_KEY=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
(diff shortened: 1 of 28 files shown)


henare:
@crowbot the theme stuff was added for a project we completed ages ago but was sadly never launched by the client :cry: It was pretty basic stuff because we were just adding it quickly and intended to iterate on it. Because the project wasn't launched that never happened.

Given your knowledge of Alaveteli, what would you do differently if you were trying to theme a reusable civic tech Rails application? Maybe we can draw on that and transform this to be something much better :sparkles:

henare:
That all sounds good Louise :ok_hand: Let us know when you're ready for this to be reviewed and merged :+1:

henare:
>@henare, I've done some more work on this - would be great to get some feedback on direction. I think 940fc9b is potentially the most controversial change - it sets the theme from the environment, so you could only run one theme per install - but that simplifies a lot of other things, including the enviroment variables for themes - they can just set the main environment variables to different values rather than using variables named for the theme. Any thoughts? 

Lots more good progress @crowbot! Unfortunately I think the ability to run multiple themes per install is actually a requirement. We definitely want it for the (unused) NSW theme because you want people signed up to that to get any alert from the main PlanningAlerts site. I can see us potentially white-labelling this for councils too, like mySociety does with FixMyStreet.

>For the example address used - is it deliberate that there were a few variations? 

What are the variations? The different street number? I don't think it's deliberate.

henare:
Hey @crowbot, just checking in to see if you have any plans to come back to this?

PR #1074, opened 2016-12-29
Enable distinguishing between past and current councillors, users can only write to cuurent councillors, but all councillors replies remain

### LINKED ISSUE #1073: Don't show people who are no longer councillors as an option to write to, but keep their replies public
After council elections, we get some new councilors, and some current ones are removed. If a person is no longer a councilor, we shouldn't show them as an option for users to write to.

We do want to keep their replies around though so people can see what they did while they were in office.

equivalentideas:
See story here on #1073

Closes #1073


equivalentideas:
Looks good to me, merging and I'll :shipit:

PR #1091, opened 2017-02-06
Fix Google Maps sensor warning

CloCkWeRX:
https://www.planningalerts.org.au/applications/774752

Google Maps API warning: SensorNotRequired https://developers.google.com/maps/documentation/javascript/error-messages#sensor-not-required

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -190,8 +190,7 @@ GEM
       actionpack (>= 3.2.13)
     formtastic_i18n (0.4.1)
     geocoder (1.2.9)
-    geokit (1.6.7)
-      multi_json (>= 1.3.2)
+    geokit (1.10.0)
     globalid (0.3.7)
       activesupport (>= 4.1.0)
     growl (1.0.3)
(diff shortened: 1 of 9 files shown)


henare:
Hey @CloCkWeRX thank you for this! :sparkles: 

It's best if PRs are merged and then deployed immediately afterwards. This means that master and what's deployed is always as close together as possible. This also means that if other things are merged into master in the mean time and you deploy and see problems, you're not confused about the source of those problems.

I've assigned this PR to me to deploy now :rocket: 

For future PRs we either need to get you deploying or you'll need to assign it to someone to review and deploy. How does that sound?

henare:
I just ran `bundle exec cap --set-before stage=production deploy:pending` to check that I'm only deploying these commits :+1:

henare:
Deployed and confirmed we're not seeing the warning anymore.

CloCkWeRX:
> assign it to someone to review and deploy.
I prefer that, I can explode less things that way :)

PR #949, opened 2016-04-04
Fix duplicated key warning in specs

CloCkWeRX:
```
spec/controllers/layar_controller_spec.rb:16: warning: duplicated key at line 21 ignored: "actions"
```

PR #1093, opened 2017-02-06
Upgrade to latest supported Ruby, version 2.3

CloCkWeRX:
Add ruby 2.2, ruby 2.3 to builds as ruby 2.0, 2.1 are EOL

We should bump to ruby 2.1 sooner rather than later, as there are issues with nokogiri 1.7+ requiring it (and nokogiri can often have security issues it seems).

To get us started down that path, I've added newer rubies to the builds - ruby 2.4 doesn't work with rails 4.2.7.1 yet though, so have left that out.

Warnings like #949 should show up over time


CloCkWeRX:
Odd that ruby 2.2/2.3 change geokit geocoding URLs, will look further in a bit

CloCkWeRX:
sysadmin bits:
```
$ rvm list known
$ rvm install 2.1.10
```

henare:
Thanks @CloCkWeRX - this PR is now on our task list.

>sysadmin bits

In the unlikely event someone else gets to this it's worth noting that there's a lot more than those sysadmin bits to do. For example we need to make sure that Passenger, cron, and the background queue will operate under the correct Ruby.

.travis.yml
```diff
@@ -2,6 +2,8 @@ language: ruby
 rvm:
   - 2.0
   - 2.1
```
henare: Should we also drop the EOL ones?

equivalentideas: Yeah I recon 👍 no one should be running it on EOL rubies if possible.

henare:
I've changed this PR to being just about upgrading to the latest supporting Ruby version, version 2.3.

There's no reason for us to support multiple Ruby versions so let's simplify and make it clear what we're planning to use in development and production by only specifying Ruby 2.3.

henare:
I've fixed up a bunch of warnings so our test suite runs cleaner.

I've also done a decent amount of smoke-testing. Now on to working out how this will deploy.

henare:
This is now deployed :rocket: 

Here's what I did:

* Deployed master to test
* Merged PR
* Installed Ruby 2.3: `bundle exec cap rvm:install_ruby -S stage=test`
* Deployed to test `bundle exec cap deploy -S stage=test`
* The application didn't boot because Passenger needed to be reconfigured to point to the new Ruby
* After reconfiguring I discovered that Passenger was running under Ruby 1.8.7 and would not boot this app running under 2.3
* Installed the latest Passenger Gem under Ruby 2.3 under RVM and configured Apache to use that and Ruby 2.3 instead
* Restarted Apache. test was running OK under 2.3 while production was still running OK under 2.0
* Discovered delayed_job worker wasn't started. I had to [change the ruby-version file](https://github.com/openaustralia/planningalerts/commit/91179cd251341b7e685e69e38668cfe5c5c00078) and upgrade RVM. Once I exported the foreman procfile to upstart and restarted it was working again
* The latest version of RVM doesn't allow `/` in aliases so we've [stopped using](https://github.com/openaustralia/planningalerts/commit/5d83856f4bf8d1d7367daddb28350bb94f1f418d) aliases/application wrappers. We now set the version of Ruby directly which is an improvement
* Updated production Apache config but didn't restart Apache
* Deployed production, using the capistrano debug flag so we could step through the deployment. This was to reduce downtime because I knew installing gems would take a while. So instead of putting up a maintenance page before deployment I only did it at the last minute
* Once the app was deployed in maintenance mode I restarted Apache and brought us out of maintenance mode
* It deployed OK and we probably had a maintenance page up for about 5 seconds total :+1: 
* To finish up I also re-exported the foreman upstart jobs and restarted the queue then I updated the daily cronjob to use Ruby 2.3 too.

PR #1094, opened 2017-02-06
2017 02 06 bundle update

CloCkWeRX:
Non security related updates, in smaller batches to catch any regression.

Eventually can turn on deppbot or similar to do this automatically - see https://github.com/CloCkWeRX/planningalerts-app/pull/1


CloCkWeRX:
Thats probably enough for now

Gemfile
```diff
@@ -36,7 +36,7 @@ gem 'validates_email_format_of', '~> 1.6', '>= 1.6.3'
 gem "geocoder"
 # Rails 4 support is a work in progress so requires tracking master
 gem 'activeadmin', '~> 1.0.0.pre2'
-gem "devise"
+gem "devise", '~> 3.5'
```
henare: Why are we specifying the version of devise? I find it handy to say so in a comment so that down the track you understand the rationale.

henare: I just tried the update myself and got:

```
Gem::InstallError: devise requires Ruby version >= 2.1.0.
An error occurred while installing devise (4.2.0), and Bundler cannot continue.
Make sure that `gem install devise -v '4.2.0'` succeeds before bundling.
```

So we know what the problem is :smile: I'll add a note myself.

henare:
I've smoke tested this locally and things look OK.

equivalentideas:
@CloCkWeRX I really appreciate these nice maintenance updates 👍

henare:
@equivalentideas thanks!

@CloCkWeRX thanks for updating these gems, this is now deployed :rocket:

CloCkWeRX:
Re
> It would be great if you could add that comment. Once that's done we can smoke test it before deploying it.

Devise tends to have rather large breaking changes - ie; https://github.com/plataformatec/devise/blob/master/CHANGELOG.md#400rc1---2016-02-01 - drops ruby 2.0 support in version 4.

It tends to be a top level gem - ie, the app directly depends on it, and the onus is on you to ensure compatibility. I've had good success pinning to the relevant versions and incrementally fixing deprecation issues before bumping to the next.

The plan was to pin to that, then basically use deppbot (https://github.com/CloCkWeRX/planningalerts-app/pull/1) to automatically get security fixes, other maintenance work that was unlikely to break the build flowing.

Can likely do a little more now that the decision is to go to 2.3 (had similar plan to incrementally edge us to the latest stable with that too)

In subsequent devise version bumps, I'll add more annotations.

henare:
>Devise tends to have rather large breaking changes - ie; https://github.com/plataformatec/devise/blob/master/CHANGELOG.md#400rc1---2016-02-01 - drops ruby 2.0 support in version 4.

Yep, cool, I added that in b3fa824a3885a97925207cf2f57b47051824973f.

>In subsequent devise version bumps, I'll add more annotations.

:ok_hand:

PR #1031, opened 2016-09-16
Bugfix/auto complete comma

davwood:
Saw this issue and thought I could resolve it.
https://github.com/openaustralia/planningalerts/issues/1003

Simplified the JS a lot as google does a lot of this for you. 

screen shot 2016-09-16 at 9 25 30 am

Only thing not exactly the same is that the states show the full name (New South Wales vs NSW) but don't think that's too big of a deal.

PR #1100, opened 2017-02-09
Fix autocomplete, which isn't working at /alerts/signup or /authorities/marrickville/applications

equivalentideas:
This fixes a regression introduced last September that's causing autocomplete to not work on the alert signup forms, it only works in the address search on the homepage.

This line https://github.com/openaustralia/planningalerts/pull/1031/commits/95b8c5ca0cecbdf09cea0de241d93a6668880008#diff-d4b9d0943f985c5e598c24a7b46462faR2 replaced the element lookup here https://github.com/openaustralia/planningalerts/pull/1031/commits/95b8c5ca0cecbdf09cea0de241d93a6668880008#diff-d4b9d0943f985c5e598c24a7b46462faL3 which was targeting the signup forms for autocomplete.

We actually see in Google Analytics about a 15% increase in people seeing errors on the alert signup form after late September when that was introduced.

screen shot 2017-02-09 at 1 50 54 pm

This PR fixes this by introducing a new class name we can use to add address autocomplete anywhere, rather than making the autocomplete script more complex again.

It also adds tests to make sure autocomplete results are being displayed on the homepage, alert signup page and in the signup on authority applications pages.

I couldn't for the life of me get capybara to actually click the suggested address to trigger the form submission—so the tests just look for results being displayed. It's a start.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -99,12 +99,6 @@ RSpec.configure do |config|
     ActionMailer::Base.deliveries = []
   end
 
-  # This is a hack to get around the ssl_required? method in
-  # the application controller which redirects poltergeist to https.
-  config.before(:each, js: true) do
-    allow(Rails.env).to receive(:development?).and_return true
-  end
-
   config.include EmailSpec::Helpers
   config.include EmailSpec::Matchers
   config.include FactoryGirl::Syntax::Methods
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -106,4 +106,5 @@ RSpec.configure do |config|
   config.include SessionHelpers, type: :feature
   config.include EnvHelpers
   config.include MockLocationHelpers
+  config.include AutocompleteHelpers
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -99,6 +99,12 @@ RSpec.configure do |config|
     ActionMailer::Base.deliveries = []
   end
 
+  # This is a hack to get around the ssl_required? method in
+  # the application controller which redirects poltergeist to https.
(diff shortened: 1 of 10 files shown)


equivalentideas:
@henare would you mind giving this a quick look over for me? The test arrangement is a bit hacky:

> I couldn't for the life of me get capybara to actually click the suggested address to trigger the form submission—so the tests just look for results being displayed. It's a start.

spec/spec_helper.rb
```diff
@@ -99,11 +99,18 @@
     ActionMailer::Base.deliveries = []
   end
 
+  # This is a hack to get around the ssl_required? method in
```
henare: Rather than adding an exception for javascript testing could we disable SSL in tests and only enable in the tests that require it? That might make our intention clearer.

i.e. change `ssl_required?` to only return true in production but modify it for the [ssl redirects tests](https://github.com/openaustralia/planningalerts/blob/52d940ccf115724b98101aaa1d04bc691dccd1e3/spec/requests/redirects_spec.rb#L30-L40).

equivalentideas: That sounds like a good idea. Its pretty bad that we're modifying the way all the tests runs just for those redirect tests, rather than making them the exception.

I'd like to fix that here cause it's been annoying me for a while and feels like something that might be having side effects.

equivalentideas: @henare is this what you were imagining https://github.com/openaustralia/planningalerts/commit/e45a180d6a28e72908855a9b85f1bb8c4c2da6bf ?

equivalentideas:
I'm gonna merge this with that refactor of the ssl redirect tests 👍

equivalentideas:
This is deployed 🚢

PR #1103, opened 2017-02-09
Bundle update 2017 02 09, including upgrades to modern Devise

CloCkWeRX:
Mostly test related stuff, but with also Devise and the related deprecation warnings being fixed.


CloCkWeRX:
I'm fairly confident the -6% coverage is actually just coveralls itself being updated :)

henare:
>I'm fairly confident the -6% coverage is actually just coveralls itself being updated :)

Since there are no application code changes this is my guess too :smile:

henare:
Deployed :rocket:

PR #1105, opened 2017-02-09
Use :local flag to add rbenv support to capistrano deploy

### LINKED ISSUE #1104: Can't deploy if you're using Rbenv locally
Currently, with the version of Ruby (2.3.1) we're using, people (me) using [rbenv](https://github.com/rbenv/rbenv) can't deploy to production. 

Here's how our capistrano config decides which ruby version to deploy with `set :rvm_ruby_string, ENV['GEM_HOME'].gsub(/.*\//,"")` https://github.com/openaustralia/planningalerts/blob/master/config/deploy.rb#L7

```
if stage != "development"
  require 'rvm/capistrano'
  set :rvm_ruby_string, ENV['GEM_HOME'].gsub(/.*\//,"")
end
```

The `ENV['GEM_HOME']` is the path where RubyGems/bundler keeps your ruby gems. The `gsub` extracts what it expects to be the current ruby version out of that path. The problem is that why that seems to work for where RMV keeps it's ruby gems, it doesn't work for where rbenv keeps them.

Here's @henare’s description of the challenge:

> work out how we get capistrano to understand we want to use 2.3.1
> and to make it work on RVM machines and rbenv
> and on the server :slightly_smiling_face:
> and no, hardcoding isn't the answer :slightly_smiling_face:
> however we do have the luxury of only using 1 Ruby version
> so if that was specified in only one place that would be a very good thing
> it's not bad right now. It's specified in `.ruby-version`
> and I guess travis.yml
> so if you can work out a solution that maintains that and works on rvm and rbenv that would be superfly

equivalentideas:
In rvm-capistrano, if you use the :local flag here, it will first check to
see if you're using rbenv and detect the local ruby verion. If you're
not using rbenv it falls back to expect rvm and uses the same statement
I'm replacing here to detect the version.

See this https://github.com/readly/rvm-capistrano/blob/master/lib/rvm/capistrano/base.rb#L77-L88

Fixes #1104

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -4,7 +4,7 @@ require 'delayed/recipes'
 set :stage, "test" unless exists? :stage
 if stage != "development"
   require 'rvm/capistrano'
-  set :rvm_ruby_string, ENV['GEM_HOME'].gsub(/.*\//,"")
+  set :rvm_ruby_string, :local
 end
 
 # This adds a task that precompiles assets for the asset pipeline


henare:
Seems like just the ticket @equivalentideas, nice work :+1: 

I'll merge this and then can you please try a deploy and confirm it works OK?

equivalentideas:
blarg, didn't work:

```
> bundle exec cap --set-before stage=production deploy                 Luke@Lukes-MacBook-Pro-4[14:45]
    triggering load callbacks
  * 2017-02-13 14:47:31 executing `deploy'
  * 2017-02-13 14:47:31 executing `deploy:update'
 ** transaction: start
  * 2017-02-13 14:47:31 executing `deploy:update_code'
    executing locally: "git ls-remote https://github.com/openaustralia/planningalerts.git HEAD"
    command finished in 1922ms
  * executing "git clone -q https://github.com/openaustralia/planningalerts.git /srv/www/www.planningalerts.org.au/app/releases/20170213034733 && cd /srv/www/www.planningalerts.org.au/app/releases/20170213034733 && git checkout -q -b deploy 38e88bb9c4bf6528db2494bc59b5a5403d87d067 && (echo 38e88bb9c4bf6528db2494bc59b5a5403d87d067 > /srv/www/www.planningalerts.org.au/app/releases/20170213034733/REVISION)"
    servers: ["kedumba.openaustraliafoundation.org.au"]
Enter passphrase for /Users/Luke/.ssh/id_rsa:
    [kedumba.openaustraliafoundation.org.au] executing command
 ** [kedumba.openaustraliafoundation.org.au :: err] Error: RVM was unable to use '2.3.0'
    command finished in 4298ms
*** [deploy:update_code] rolling back
  * executing "rm -rf /srv/www/www.planningalerts.org.au/app/releases/20170213034733; true"
    servers: ["kedumba.openaustraliafoundation.org.au"]
    [kedumba.openaustraliafoundation.org.au] executing command
*** [err :: kedumba.openaustraliafoundation.org.au] Error: RVM was unable to use '2.3.0'
    command finished in 1744ms
 ** [deploy:update_code] exception while rolling back: Capistrano::CommandError, failed: "rvm_path=$HOME/.rvm $HOME/.rvm/bin/rvm-shell '2.3.0' -c 'rm -rf /srv/www/www.planningalerts.org.au/app/releases/20170213034733; true'" on kedumba.openaustraliafoundation.org.au
failed: "rvm_path=$HOME/.rvm $HOME/.rvm/bin/rvm-shell '2.3.0' -c 'git clone -q https://github.com/openaustralia/planningalerts.git /srv/www/www.planningalerts.org.au/app/releases/20170213034733 && cd /srv/www/www.planningalerts.org.au/app/releases/20170213034733 && git checkout -q -b deploy 38e88bb9c4bf6528db2494bc59b5a5403d87d067 && (echo 38e88bb9c4bf6528db2494bc59b5a5403d87d067 > /srv/www/www.planningalerts.org.au/app/releases/20170213034733/REVISION)'" on kedumba.openaustraliafoundation.org.au
```

henare:
Fixed in  #1106.

PR #1108, opened 2017-02-14
Authority count refactor

henare:
During some recent work I realised we were setting `@authority_count` on every request but we stopped using it in the header quite a while back. This refactor is mainly to address that but I've also tried to make things a bit nicer too.


equivalentideas:
Looks good 👍🏻

Issue #794, opened 2015-09-29
The important information that there's a limited time to comment on an application is buried


equivalentideas:
If there's only a few days to comment on an application, that's quite important information "You have 1 day left to officially respond to this application.". It's also adds an urgency to commenting.

Currently this information in one of several paragraphs under that 'Add your comment' heading:

![screen shot 2015-09-29 at 12 06 18 pm](https://cloud.githubusercontent.com/assets/1239550/10153809/8e06b010-66a3-11e5-8a85-ea35231a674c.png)

I think it should be moved above this section for two reasons:
1. This is information that will prompt you to comment so should lead you to the 'add your comment' section, not be a sub point of it.
2. This is important information that should be seen quite quickly when looking at the application.

It could go above the ''We found this application for you on the planning authority's website 7 days ago. It was received by them 1 day earlier." sentence, which is conceptually related but less important to people:

![screen shot 2015-09-29 at 12 08 48 pm](https://cloud.githubusercontent.com/assets/1239550/10153811/921503fa-66a3-11e5-98eb-17ccca244504.png)

We could AB test this to see what impact it has on commenting and alert signups.

henare:
:+1: I like the idea of making it part of the text immediately below the comment header:

>You have 1 day left to officially respond to this application. Add your own comment!

equivalentideas:
> :+1: I like the idea of making it part of the text immediately below the comment header:

Thanks @henare do you mean position 2 in this image? I was suggesting position 1 above.

![screen shot 2015-09-29 at 4 48 34 pm](https://cloud.githubusercontent.com/assets/1239550/10157081/42b4d28c-66ca-11e5-9366-1ab8a713f514.png)

henare:
Position 2 is the most obvious because it enhances what's already there.

We currently have a separator between the application and the comments so including information about commenting in the application section would be muddling things up.

equivalentideas:
> We currently have a separator between the application and the comments so including information about commenting in the application section would be muddling things up.

hmmm. I actually think it's more confusing to have this information inside the comment section and it's something that should get you interested in the comments and commenting, rather than something you discover once you're looking at the comments/commenting. We should have a play and go with whatever works best :balloon:

equivalentideas:
Looks like we thought a good course of action would be to move the notice to position 2 discussed above. That also feels like a nice simple iteration.

mlandauer:
I think this change is a good one

kat:
agreed!

—
[...]

PR #1125, opened 2017-03-25
Fix for the issue https://github.com/openaustralia/planningalerts/issues/794

Anishka0107:
This PR tries to fix the issue #794 
The important information that there's a limited time to comment on an application is buried and it should be moved to the new position as specified in the issue.


equivalentideas:
Hi @Anishka0107 . Did you mean to close this PR?

Anishka0107:
Hi @equivalentideas, sir the Travis check is unsuccessful for this commit. So, I had to close it. I also submitted another PR for this repository, but the same happened for it as well, the check was unsuccessful.
I do not know what is going wrong. Could you help me

Issue #826, opened 2015-10-27
Yahoo Pipes is dead


henare:
On our API page we suggest Yahoo Pipes. Thanks to Steve for emailing us to point out that Yahoo Pipes is [long] dead!

We need to remove Yahoo Pipes from the page but it would also be good to replace it with another suggestion. The question is what?

Anishka0107:
I guess [Zapier](https://zapier.com/) is good substitute to Yahoo! Pipes.

equivalentideas:
> I guess Zapier is good substitute to Yahoo! Pipes.

I'm not super familiar with Zapier, why do you think it would be good @Anishka0107 ?

It would be good to look at a few alternatives in the mix :) are there any open source options ❤️

Anishka0107:
@equivalentideas 
https://github.com/superfeedr/superpipes seems like the best alternative for Yahoo! Pipes and is open-source as well. I guess I should use superpipes instead of Zapier in the PR.

equivalentideas:
> I guess I should use superpipes instead of Zapier in the PR.

Before deciding on an option we should consider the pros and cons of a few alternatives. We might way up:

* what's the most common tool that people already use?
* what will be easy for API users to get started with?
* is there a tool that aligns with our aims, e.g. is FOSS and an ethical company etc. .

I suspect nothing will be perfect, but let's way up the options.

Anishka0107:
@equivalentideas 
I have submitted a PR with Superpipes instead of Zapier as I feel it is the best open source available alternative and  was designed to actually replace Yahoo! pipes, when they had suspected the closing of Yahoo pipes.

henare:
Rather than finding a Yahoo! Pipes replacement another approach to fixing this would be to rework this sentence to something useful for what API developers would expect these days.

jamezpolley:
Thanks @Anishka0107

PR #1126, opened 2017-03-26
Fix the issue #826

Anishka0107:
Fix for the issue https://github.com/openaustralia/planningalerts/issues/826.
Yahoo! Pipes does not exist anymore. So, a good alternative to it seems to be Zapier https://zapier.com/ 
The link for Yahoo Pipes on the Get the Data - API page is fixed.

### DIFF
--- a/app/views/api/howto.html.haml
+++ b/app/views/api/howto.html.haml
@@ -4,7 +4,7 @@
 
 %p
   Planning application data is available programmatically as feeds which can
-  be used in most web mapping APIs and desktop GIS software like #{link_to "Yahoo Pipes", "http://pipes.yahoo.com/"}. Details of the API are listed below.
+  be used in most web mapping APIs and desktop GIS software like <a href="https://zapier.com/">Zapier</a>. Details of the API are listed below.
 
 %p
   Non&ndash;commercial, low&ndash;volume use of this service is free.
--- a/app/views/api/howto.html.haml
+++ b/app/views/api/howto.html.haml
@@ -4,7 +4,7 @@
 
 %p
   Planning application data is available programmatically as feeds which can
-  be used in most web mapping APIs and desktop GIS software like <a href="https://zapier.com/">Zapier</a>. Details of the API are listed below.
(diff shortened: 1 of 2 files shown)


Anishka0107:
I am unable to understand why the checks are unsuccessful. Can somebody please help me here?

equivalentideas:
> I am unable to understand why the checks are unsuccessful. Can somebody please help me here?

That check is the tests failing. But those failures have nothing to do with the change you made, so all good 👍 Those tests were added recently and maybe they're a bit brittle.

app/views/api/howto.html.haml
```diff
@@ -4,7 +4,7 @@
 
 %p
   Planning application data is available programmatically as feeds which can
-  be used in most web mapping APIs and desktop GIS software like #{link_to "Yahoo Pipes", "http://pipes.yahoo.com/"}. Details of the API are listed below.
+  be used in most web mapping APIs and desktop GIS software like <a href="https://zapier.com/">Zapier</a>. Details of the API are listed below.
```
equivalentideas: This should use the [Rails link_to helper](http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) like the original.

app/views/applications/show.html.haml
```diff
@@ -14,7 +14,7 @@
   #application
     %article#application-info
       %header
-        %h3.address= @application.address
+        %h1.address= @application.address
```
equivalentideas: This change is related to a different issue that #826 so it shouldn't be part of this PR (that would be an awesome issue to fix in a separate PR though 🎉 ).

app/views/api/howto.html.haml
Anishka0107: As you said sir,
Updated <a> to link_to and also changed Zapier to Superpipes as it seems a better alternative and is open-source as well.

app/views/applications/show.html.haml
Anishka0107: I have changed it back to the original one and will submit a separate PR when this PR gets merged.

PR #1127, opened 2017-03-28
Updated Yahoo! Pipes link to Superpipes

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/planningalerts/issues/826

The Yahoo! Pipes link on the API - Get the Data page is dead as Yahoo! Pipes does not exist anymore.
Hence the link has been updated to Superpipes ( https://github.com/superfeedr/superpipes ).
Superpipes is an open-source Yahoo! Pipes equivalent, which was developed to replace it, when they had anticipated their closing.

### DIFF
--- a/app/views/api/howto.html.haml
+++ b/app/views/api/howto.html.haml
@@ -4,7 +4,7 @@
 
 %p
   Planning application data is available programmatically as feeds which can
-  be used in most web mapping APIs and desktop GIS software like #{link_to "Yahoo Pipes", "http://pipes.yahoo.com/"}. Details of the API are listed below.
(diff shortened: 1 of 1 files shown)


henare:
Superpipes hasn't been updated in 5 years so I suspect it could be unmaintained. I've never used it so I couldn't recommend it. As [@equivalentideas said](https://github.com/openaustralia/planningalerts/issues/826#issuecomment-289656671) on #826,

>Before deciding on an option we should consider the pros and cons of a few alternatives.

We should do that on the issue first so I'm going to close this PR until that's happened.

Issue #614, opened 2015-01-31
Headings do not use correct semantic heading levels


equivalentideas:
Currently the application address/page title is an `<h3>`. It's the title of the page and should be an `<h1>`. 

This is an accessibility problem and is likely not helping search engine crawlers understand what this page is about.

mlandauer:
Would be good to address this

PR #1128, opened 2017-03-28
Correct semantic heading for address page

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/planningalerts/issues/614

The title of the application address page uses an h3, while it should be an h1 to improve its accessibility for web crawlers. It has now been changed to h1.
Although this increases the size of the header, it is a better option as it is important for search engines to index the page properly.
The heading now uses the correct semantic heading level.

### DIFF
--- a/app/views/applications/show.html.haml
+++ b/app/views/applications/show.html.haml
@@ -14,7 +14,7 @@
   #application
     %article#application-info
       %header
-        %h3.address= @application.address
+        %h1.address= @application.address
 
       - if @application.location
         #map_div


equivalentideas:
Hi @Anishka0107 does this impact the visual appearance of the page? Could you please include a screenshot of the page here so we can see what it looks like?

Thanks :)

Anishka0107:
@equivalentideas 
Yes it does! And I guess, I need to keep the look of the page consistent, while changing the h3 to an h1.
I will work on it soon.

equivalentideas:
I'm going to close this to clear up our list of open PRs 🙂  @Anishka0107 please reopen and pick this up if/when you feel like it! Looks like you've been busy 😁

PR #1129, opened 2017-03-28
Shift important information to correct position

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/planningalerts/issues/794
The important information of the limited time left to comment on the application is buried and not very visible when present below the Add your own comment section.
Being placed immediately below the comment header is an appropriate position for it.
So, it has been shifted to the desired position.


app/views/comments/_comment_form.html.haml
```diff
@@ -7,5 +7,3 @@
   %p
     = link_to "How to comment", @application.comment_url
     on this application (from #{@application.authority.full_name}).
-    - if @application.on_notice_to
```
equivalentideas: It looks like you're shifting this from the wrong place. Have a look at when this notice would be displayed, only when the authority isn't contactable and the comment form isn't displayed. In that case there would be no comments either, so the notice isn't actually being pushed down the page—it's kinda in the perfect place with the comment instructions.

We want to keep this one here and move the notice that appears with the form. Let me know if you need a hand finding that.

equivalentideas: Oh, one other thing :) Those two commits should be one commit moving the notice, rather than delete and insert. That way the whole intention is captured in the single commit we can look back on.

equivalentideas:
@Anishka0107 Could you add a screenshot of how the change looks to users to this PR please?

equivalentideas:
I'm going to close this to clear up our list of open PRs 🙂  @Anishka0107 please reopen and pick this up if/when you feel like it.

Anishka0107:
Sorry, I didn't update the PRs.
Sure, I'll do so when some work is done :)

equivalentideas:
No worries @Anishka0107 :) open it when you're ready. I'll close the PRs on publicwhip as well till then just to clear up those lists. Thanks again for all your contributions :)

Issue #964, opened 2016-04-21
Reply permalinks on comments index don't link to comment correctly


henare:
To reproduce:
- Go to comments page: https://www.planningalerts.org.au/comments
- Find a reply from a councillor
- Click permalink
- It links to https://www.planningalerts.org.au/comments#reply20 instead of https://www.planningalerts.org.au/applications/635635#reply20

equivalentideas:
Fixed by @hisayohorie in https://github.com/openaustralia/planningalerts/pull/1131

PR #1131, opened 2017-03-30
fixed the permalink path to application page reply

hisayohorie:
#964 changed the permalink in councillor replies to go to the reply under the application page.

### DIFF
--- a/app/views/replies/_reply.html.haml
+++ b/app/views/replies/_reply.html.haml
@@ -9,7 +9,7 @@
         %span.comment-time-block
           replied to
           = link_to reply.comment.name, "#comment#{reply.comment.id}", title: "View comment"
-          = link_to "#reply#{reply.id}", title: "View reply" do
+          = link_to "#{application_path(reply.comment.application)}\#reply#{reply.id}", title: "View reply" do
             %time.comment-time{datetime: reply.received_at.strftime("%F")}
               #{time_ago_in_words(reply.received_at)} ago
     %blockquote.comment-text= simple_format(reply.text)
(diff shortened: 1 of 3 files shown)


app/views/replies/_reply.html.haml
```diff
@@ -9,7 +9,7 @@
         %span.comment-time-block
           replied to
           = link_to reply.comment.name, "#comment#{reply.comment.id}", title: "View comment"
-          = link_to "#reply#{reply.id}", title: "View reply" do
+          = link_to "#{application_path(reply.comment.application)}\#reply#{reply.id}", title: "View reply" do
```
equivalentideas: This definitely does the job, but the implementation should be consistent with how the links on comments work, see https://github.com/openaustralia/planningalerts/blob/master/app/views/comments/_comment_meta_sentence.html.haml#L9

Basically that's using a helper method that gets this logic out of the view https://github.com/openaustralia/planningalerts/blob/master/spec/helpers/comments_helper_spec.rb#L23-L30

equivalentideas:
This is merged 🎸 Thanks @hisayohorie . This is a small but solid usability enhancement 👍 

I made a few cleanup commits on the end https://github.com/openaustralia/planningalerts/compare/cfbeaa24b333...9144162cf2af
 
Keep a close eye on the style of the code throughout a project when you make a contribution. Things like whitespace and hash style (`{key: value}` or `{:key => value}`) are important to keep consistent, it’s makes code much more readable.

equivalentideas:
This is deployed :)

PR #1132, opened 2017-04-02
 added CommentHelper in confirmation mailer using add_template_helper…

hisayohorie:
… and used comment_as_html to preserves line breaks in the comment.html.haml.

let me know what you think!

### DIFF
--- a/app/views/confirmation_mailer/comment.html.haml
+++ b/app/views/confirmation_mailer/comment.html.haml
@@ -7,9 +7,9 @@
 
 %p If your email program does not let you click on this link, just copy and paste it into your web browser and hit return.
 
-%p
-  Your comment:
-  = comment_as_html(@comment.text)
+%p Your comment:
+
+%blockquote= comment_as_html(@comment.text)
 
 %p
   On the application:
--- a/app/views/confirmation_mailer/comment.html.haml
+++ b/app/views/confirmation_mailer/comment.html.haml
@@ -9,7 +9,7 @@
 
 %p
   Your comment:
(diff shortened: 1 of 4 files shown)


app/views/confirmation_mailer/comment.html.haml
```diff
@@ -9,7 +9,7 @@
 
 %p
   Your comment:
-  = @comment.text
+  = comment_as_html(@comment.text)
```
equivalentideas: This method, which will generate html (mostly <p> tags) is currently a child of a <p> element. This is going to generate invalid markup.

I actually think in this context the comment is a blockquote, as we're quoting their comment. It's semantically a different type of text to the rest of the text in this file.

I recon something like:

```
%p Your comment:

%blockquote= comment_as_html(@comment.text)
```

Or, it would be worth having a look at how it's inserted into alert emails.

hisayohorie:
screen shot 2017-04-03 at 2 49 10 pm
Here is the screenshot, I used letter_opener to preview. I changed the mark up so it is no longer <p> nested in <p>, but there is still a <p> inside of the <blockquote>. is that ok, or should I find out the way to remove the nested <p> in <blockquote>?

equivalentideas:
Nice one @hisayohorie 🍰 I've merged and deployed this :shipit: 

I made a change to the way the comment helpers were included to make it consistent with the ReplyNotifier mailer, and then made it consistent across all the mailers: https://github.com/openaustralia/planningalerts/compare/5973e7b129b3...b394be25cd35

Issue #947, opened 2016-04-04
When I want to look at the json api data for an application/search/authority, give me a link to view it if I'm logged in


equivalentideas:
I was just wanting to look at the JSON data for a council. I just wanted to see an example of the data.

I went to the coverage page, found a council, when to all their recent applications, and put .json on the end of the url. Blank page.

I then went to the 'Get the data' page and scrolled through the sections for the authority example.

I then copied the `?key=xxxx` bit and put that on the end of my other tab. Still nothing. I then when back and looked at the example again and noticed that the extension was '.js`.

This is a lot of manual messing around to get there.

Could we just have a `View as JSON` link that already has your key shown if you're logged in? Would that have saved me all the confusion?

hisayohorie:
I'll work on this!

PR #1133, opened 2017-04-04
added link to JSON data of applications on applications/index.html.haml

hisayohorie:
I'm working on the issue #947, to create a link to see the list of applications as JSON data.
I haven't write the test yet as I have a question.

in this PR, I created application_json_path(authority) to direct user to /authorities/authority.short_name_encoded/application.js?key=current_user.api_key

I'm wondering if this link should *only* be on the applications index for a specific authority, or if it should also be on the general applications index like 
/applications.js?key=current_user.api_key  
(in which case the user needs bulk api access)

### DIFF
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -1,6 +1,4 @@
 module ApplicationsHelper
-
-
   def display_description_with_address(application)
     display_description =
       if application.description
@@ -70,7 +68,7 @@ module ApplicationsHelper
     "#{@application.address} | #{@application.date_scraped.to_date.to_formatted_s(:rfc822)}"
   end
 
-  def applications_json_path(authority)
+  def authority_applications_json_path_for_current_user(authority)
     authority_applications_path(authority.short_name_encoded, format: :js, key: current_user.api_key)
   end
 
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -67,9 +67,14 @@ module ApplicationsHelper
     # Include the scraping date in the title so that multiple applications from the same address have different titles
     "#{@application.address} | #{@application.date_scraped.to_date.to_formatted_s(:rfc822)}"
   end
+  
+  # TODO: extract to theme
+  def api_host
+   "api.planningalerts.org.au"
+  end
 
-  def authority_applications_json_path_for_current_user(authority)
-    authority_applications_path(authority.short_name_encoded, format: :js, key: current_user.api_key)
+  def authority_applications_json_url_for_current_user(authority)
+    authority_applications_url(authority.short_name_encoded, host: api_host, format: :js, key: current_user.api_key)
   end
 
   def google_static_map(application, options)
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -1,4 +1,6 @@
 module ApplicationsHelper
+
+
   def display_description_with_address(application)
     display_description =
       if application.description
@@ -68,6 +70,10 @@ module ApplicationsHelper
     "#{@application.address} | #{@application.date_scraped.to_date.to_formatted_s(:rfc822)}"
   end
 
+  def applications_json_path(authority)
+    authority_applications_path(authority.short_name_encoded, format: :js, key: current_user.api_key)
+  end
+
   def google_static_map(application, options)
     google_static_map2(options.merge(lat: application.lat, lng: application.lng, label: "Map of #{application.address}"))
   end
@@ -92,4 +98,5 @@ module ApplicationsHelper
     size = options[:size] || "350x200"
     image_tag(google_static_streetview_url(application, options), size: size, alt: "Streetview of #{application.address}")
   end
+
 end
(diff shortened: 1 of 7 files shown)


equivalentideas:
> I'm wondering if this link should only be on the applications index for a specific authority, or if it should also be on the general applications index like

I think for now just complete the issue (just the authority pages) and then we can build on that.

app/helpers/applications_helper.rb
```diff
@@ -68,6 +70,10 @@ def page_title(application)
     "#{@application.address} | #{@application.date_scraped.to_date.to_formatted_s(:rfc822)}"
   end
 
+  def applications_json_path(authority)
```
equivalentideas: I think it would be good to name this something more descriptive, even if it's more verbose, like `authority_applications_json_path_for_current_user(authority)` so it's super clear what you're gonna get in the view.

equivalentideas: That also matches the similar existing `authority_applications_path`.

spec/helpers/applications_helper_spec.rb
```diff
@@ -171,6 +171,18 @@
       end
     end
 
+    describe "applications_json_path" do
```
equivalentideas: This does the job, but it makes the [separate test phases hard](https://robots.thoughtbot.com/four-phase-test) to read.

Also using [BetterSpecs](http://betterspecs.org/#subject) advice about using 'subject' etc. something like this might be clearer:

```
    describe "applications_json_path" do
      let(:authority) { create(:authority, short_name: "marrickville", ) }
      let(:user) { build(:user, api_key: "ABCDE12345") }

      before { expect(helper).to receive(:current_user).and_return(user) }

      subject { helper.applications_json_path(authority) }

      it { is_expected.to eq("/authorities/marrickville/applications.js?key=ABCDE12345") }
    end
```

Note that it uses `build` not `create` for the user, to avoid the before_create callback.

equivalentideas: This should also point to the api subdomain, see https://github.com/openaustralia/planningalerts/blob/master/app/helpers/api_howto_helper.rb#L39-L42

and 

https://github.com/openaustralia/planningalerts/blob/master/app/helpers/api_howto_helper.rb#L11-L13

I think the best way to do that for now is to hard code it into a method on applications_helper like `api_host` to match the example. It should really be extracted into the theme, but if you don't feel like doing that now, I wouldn't worry about it. But implementing it the same way as in [`api_howto_helper.rb`](https://github.com/openaustralia/planningalerts/blob/master/app/helpers/api_howto_helper.rb#L11-L13) will help us extract it. If you could also add a comment like `# TODO: extract to theme` in the code, that would help.

app/views/applications/index.html.haml
```diff
@@ -14,5 +14,8 @@
     %span.highlight It looks like something might be wrong. The latest application was received #{time_ago_in_words(@authority.latest_application_date)} ago.
     = link_to "Why?", faq_path(:anchor => "broken_scraper")
 
+- if current_user && @authority
+  = link_to "View as JSON", applications_json_path(@authority)
```
equivalentideas: I think this would go better above the pagination links at the end of the list. This is because this is a kinda nice to have and we don't know if people will really use it—so better to not add more links to the already overcrowded top of the page. I think put it down there and api users will discover it. We can also add click tracking to it to see if people do.

app/helpers/applications_helper.rb
```diff
@@ -1,4 +1,6 @@
 module ApplicationsHelper
+
+
```
hisayohorie: ah got it. will be mindful next time

app/mailers/confirmation_mailer.rb
```diff
@@ -1,6 +1,8 @@
 class ConfirmationMailer < ActionMailer::Base
   include ActionMailerThemer
 
+  add_template_helper(CommentsHelper)
```
hisayohorie: got it!

equivalentideas:
Hey @hisayohorie looks like you've made progress 🚀 Let me know when I can review and merge this 🚢

hisayohorie:
@equivalentideas yes! I have changed:
- the name of authority_json_path to authority_applications_json_url_for_current_user
-  added api_host method in applications_helper.rb
- reformatted the spec

I think it is ready for your review!

equivalentideas:
Real nice work @hisayohorie 🚀 This has been merged 🔨 

I added a few things https://github.com/openaustralia/planningalerts/compare/fedb4bb27192...8c2f4965c71b (there's also a stray commit in there with some unrelated tests I wrote). You can read about them in the commit messages.

This is cool and it gave me a little idea for another way to point people to sign up for the API, ones we make some of the changes to commercial API sign up stuff ;-)

henare:
Is this deployed? I was excited to see it in action but I can't find it :eye:

equivalentideas:
> Is this deployed?

Nup, it was pretty late night here when I merged so set a reminder to deploy this morning. Should have noted that here 👍

henare:
>Nup, it was pretty late night here when I merged so set a reminder to deploy this morning.

No worries - good call :smile:

equivalentideas:
Deployed! https://www.planningalerts.org.au/authorities/act/applications nice one @hisayohorie :shipit:

Issue #1038, opened 2016-09-30
Limit the number of comments in email alerts


henare:
We limit applications to 10 per alert, we should do the same with comments.

I noticed this when 100 comments were posted on 1 application recently - the email alerts weren't very usable.

hisayohorie:
I'll work on this!

equivalentideas:
From @hisayohorie in chat:

> When the comments are more than 10 in the email, where should I link them to see more comments? Since the comments could be from multiple applications in the area, should I link them to `/comments`? but then, the comments lose context. Or `address_application_url` like in `alert_notifier`?

Great question. It really feels to me like you'd want to be linked to a list of local applications that were getting all these comments.

This feels related to https://github.com/openaustralia/planningalerts/issues/866 .

I can imagine adding this in a series of steps.

1. changing the [application list display to include if the application has comments](https://github.com/openaustralia/planningalerts/issues/866)
2. adding a 'sort by recent comments' option or something to the applications search results so people can find DAs that have caused discussion locally
3. finally this issue of linking to that page and limiting down the comments in emails

What do you recon @hisayohorie ?

hisayohorie:
@equivalentideas so that it will show the *list of the applications with comments, not the specific application the user commented on.  It does make sense since there can be more than one applications. 

So the link should be address_applications_url

Please ignore the above PR, I will fix it with the appropriate link. Also I will take on the issue #866

equivalentideas:
> Please ignore the above PR, I will fix it with the appropriate link.

No worries, could you close it and leave a comment so it's clear why when people come across it :)

>  Also I will take on the issue #866

🎸 

> 
> @equivalentideas so that it will show the *list of the applications with comments, not the specific application the user commented on. It does make sense since there can be more than one applications.
> 
> So the link should be address_applications_url

I'm not sure what you mean by this?

hisayohorie:
@equivalentideas in this PR I used `link_to comment_url` for view more comments if there are more than 10 comments on the application. And what I meant was to use `link_to address_applications_url`, if I am understanding it correctly. Sorry I wasn't very clear here :man_facepalming:

mlandauer:
Would be a good thing having this

PR #1135, opened 2017-04-11
Issue#1038 display the latest 10 comments per email

hisayohorie:
This is a PR for #1038 

Also in the process, I noticed probably it is a good idea to implement the same idea of displaying 10 latest comments on application#show using pagination, just as it is implemented on comment#index.


hisayohorie:
This PR is closed because there steps need to be done before the issue of linking to that page and limiting down the comments in emails.
1)  changing the application list display to include if the application has comments
2) adding a 'sort by recent comments' option or something to the applications search results so people can find DAs that have caused discussion locally

Issue #1085, opened 2017-02-06
Improve opengraph headers specifically for facebook sharing of development applications


CloCkWeRX:
Parent issue #1083

https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.planningalerts.org.au%2Fapplications%2F774752

### Warnings
 * The 'og:url' property should be explicitly provided, even if a value can be inferred from other tags.
 * The 'og:title' property should be explicitly provided, even if a value can be inferred from other tags.
 * The 'og:description' property should be explicitly provided, even if a value can be inferred from other tags.
 * The 'og:image' property should be explicitly provided, even if a value can be inferred from other tags.
 * The 'fb:app_id' property should be explicitly provided, Specify the app ID so that stories shared to Facebook will be properly attributed to the app. Alternatively, app_id can be set in url when open the share dialog.
 * The 'og:image' property should be explicitly provided, even if a value can be inferred from other tags.

### Current appearance - Facebook
![image](https://cloud.githubusercontent.com/assets/365751/22635778/0aaba6de-ec86-11e6-8fee-c502230df157.png)

 * Needs a better description, such as "Development application for (address) to (modifications)" or similar
 * Could use a call to action, like "Comment on this application if it affects you"
 * Could use lat/long data (https://developers.facebook.com/docs/reference/opengraph/object-type/place/)
 * Could use a proper image - either a static maps API rendering or a planning alerts logo

### Other metadata
Worth checking schema.org for useful types, such as https://schema.org/Event for the notification period.

equivalentideas:
A little old, but this is great reading on this topic https://alistapart.com/article/like-able-content-spread-your-message-with-third-party-metadata

trang20t:
Hi there! In order to contribute to this, do I submit a html page or png image?

henare:
@trang20t you'll need to set up the Rails application and then add these additional meta tags.

TheChirpyWitch:
@equivalentideas I have resolved this issue. Please merge my pull request. Thanks!

equivalentideas:
> @equivalentideas I have resolved this issue. Please merge my pull request. Thanks!

Thanks @TheChirpyWitch 🏞 I'll leave this till that PR is merged, there appears to be ongoing discussion there :) https://github.com/openaustralia/planningalerts/pull/1137

PR #1137, opened 2017-04-13
Adding Facebook Og Tags

TheChirpyWitch:
I have added the og meta tags for the app. I tested it with Facebook Sharing Debugger and the warnings are now removed. Below is the image of the updated page in the facebook sharing debugger
Image of Facebook sharing debugger.

### DIFF
--- a/app/views/applications/show.html.haml
+++ b/app/views/applications/show.html.haml
@@ -6,6 +6,11 @@
   %meta(name="twitter:title" content="#{@application.address}")
   %meta(name="twitter:description" content="#{@application.description[0..199] if @application.description}")
   %meta(name="twitter:image" content="#{google_static_streetview_url(@application, size: "120x120")}")
+  %meta(property="fb:app_id" content="335651659855671")
+  %meta(property="og:url" content="#{application_url(@application)}")
+  %meta(property="og:title" content="#{@application.address}")
+  %meta(property="og:description" content="#{@application.description[0..199] if @application.description}")
+  %meta(property="og:image" content="#{google_static_streetview_url(@application, size: "200x200")}")
 
 = render "applications/facebook"
 = render "applications/twitter"


TheChirpyWitch:
Resolving Issue #1085

app/views/applications/show.html.haml
```diff
@@ -6,6 +6,11 @@
   %meta(name="twitter:title" content="#{@application.address}")
   %meta(name="twitter:description" content="#{@application.description[0..199] if @application.description}")
   %meta(name="twitter:image" content="#{google_static_streetview_url(@application, size: "120x120")}")
+  %meta(property="fb:app_id" content="335651659855671")
```
henare: What app ID is this?

app/views/applications/show.html.haml
```diff
@@ -6,6 +6,11 @@
...
+  %meta(property="og:url" content="#{application_url(@application)}")
+  %meta(property="og:title" content="#{@application.address}")
```
henare: Would it be a good idea to include something more than the address in the title?

app/views/applications/show.html.haml
```diff
@@ -6,6 +6,11 @@
...
+  %meta(property="og:description" content="#{@application.description[0..199] if @application.description}")
```
henare: Why is this limited to 200 characters?

henare: In #1085 it says,

>Needs a better description, such as "Development application for (address) to (modifications)" or similar

I agree it needs more than just the application description. But maybe we should say "PlanningAlert for (address) to (modifications)" :smile:

app/views/applications/show.html.haml
```diff
@@ -6,6 +6,11 @@
...
+  %meta(property="og:image" content="#{google_static_streetview_url(@application, size: "200x200")}")
```
henare: Why 200x200?

app/views/applications/show.html.haml
TheChirpyWitch: Hey! The app_id is the one present in 

> app/views/applications/_facebook.html.haml as is set in bold here:

`%div(id="fb-root")
:javascript
  (function(d, s, id) {
  var js, fjs = d.getElementsByTagName(s)[0];
  if (d.getElementById(id)) return;
  js = d.createElement(s); js.id = id;
  js.src = "//connect.facebook.net/en_US/all.js#xfbml=1&**appId=335651659855671**";
  fjs.parentNode.insertBefore(js, fjs);
  }(document, 'script', 'facebook-jssdk'));

  window.fbAsyncInit = function() {
    FB.Event.subscribe('edge.create', function(targetUrl) {
      _gaq.push(['_trackSocial', 'facebook', 'like', targetUrl]);
    });
  };

`

app/views/applications/show.html.haml
TheChirpyWitch: Sure, we could include something else. This is was what was present in the Twitter title, so I simply copied it. We could change it on both places.

app/views/applications/show.html.haml
TheChirpyWitch: Again, this was included in the Twitter app and so I figured there was some reason to it, but I was a little apprehensive because Facebook recommends 90 characters in its design recommendations.

TheChirpyWitch: Sure, what would be included in "modifications" though? Please clarify. :smile:

app/views/applications/show.html.haml
TheChirpyWitch: No particular reason. We could make it 120x120 to correspond with the Twitter image. :sweat_smile: 
This [link](http://stackoverflow.com/questions/18835027/new-ogimage-size-for-facebook-share) says it should be at least 600x315.

app/views/applications/show.html.haml
henare: Now you caught me copying :smile: It would be the `application.description`.

henare: Twitter and Facebook are different so we should use what's appropriate for each application. If Facebook recommend 90 characters [citation needed] ;) then we should try and design for that.

app/views/applications/show.html.haml
henare: As I said on another comment Twitter and Facebook are different applications and so should have different and appropriate things set for each. Basically we want it to look awesome and native in each place :sparkles:

app/views/applications/show.html.haml
henare: \*weird\* and well spotted! :smile: I've opened #1146 to work out what to do with this and to make it configurable. It's :+1: to use in this PR.

app/views/applications/show.html.haml
henare: As I've commented elsewhere, let's make it specific to Facebook and show something that looks great, is useful for the user, and feels native to the application (i.e. Facebook).

PR #1139, opened 2017-04-13
Url encode Morph API Key - Fixes #1138

### LINKED ISSUE #1138: Encoding issue with Morph API Key
While trying to resolve issue #1085 , for some time, my Morph API key would throw an error like this:
```
I, [2017-04-13T20:54:36.873656 #17311]  INFO -- : Scraping 1 authorities
E, [2017-04-13T20:54:38.615018 #17311] ERROR -- : Error 401 Unauthorized while getting data from url https://api.morph.io/planningalerts-scrapers/marrickville/data.json?query=select+%2A+from+%60data%60+where+%60date_scraped%60+%3E%3D+%272017-04-08%27+and+%60date_scraped%60+%3C%3D+%272017-04-13%27&key=<redacted>. So, skipping
I, [2017-04-13T20:54:38.691172 #17311]  INFO -- : 0 new applications found for Marrickville Council, NSW with date from 2017-04-08 to 2017-04-13
I, [2017-04-13T20:54:38.757675 #17311]  INFO -- : Took 1 s to collect applications from Marrickville Council, NSW
```
This issue arose because my MORPH_API_Key had a '+' in it and the url was malformed (needed to be url encoded).
For the moment, I simply use `%2B` instead of `+`, but this issue will affect other newcomers in the future and hence should be resolved.

TheChirpyWitch:
Resolved #1138

### DIFF
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -220,7 +220,7 @@ class Authority < ActiveRecord::Base
 
   def morph_feed_url_for_date_range(start_date, end_date)
     query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
-    key = CGI.escape(ENV['MORPH_API_KEY']) || ""
+    key = CGI.escape(ENV['MORPH_API_KEY'] || "")
     "https://api.morph.io/#{morph_name}/data.json?query=#{query}&key=#{key}"
   end
 
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -220,7 +220,7 @@ class Authority < ActiveRecord::Base
 
   def morph_feed_url_for_date_range(start_date, end_date)
     query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
-    key = CGI.escape ENV['MORPH_API_KEY']
+    key = CGI.escape ENV['MORPH_API_KEY'] || ""
     "https://api.morph.io/#{morph_name}/data.json?query=#{query}&key=#{key}"
   end
 
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -220,7 +220,7 @@ class Authority < ActiveRecord::Base
 
   def morph_feed_url_for_date_range(start_date, end_date)
     query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
-    key = CGI.escape ENV['MORPH_API_KEY'] || ""
+    key = CGI.escape ENV['MORPH_API_KEY']
     "https://api.morph.io/#{morph_name}/data.json?query=#{query}&key=#{key}"
   end
 
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -220,7 +220,7 @@ class Authority < ActiveRecord::Base
 
   def morph_feed_url_for_date_range(start_date, end_date)
     query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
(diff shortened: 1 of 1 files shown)


app/models/authority.rb
```diff
@@ -220,7 +220,8 @@ def morph_url
 
   def morph_feed_url_for_date_range(start_date, end_date)
     query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
-    "https://api.morph.io/#{morph_name}/data.json?query=#{query}&key=#{ENV['MORPH_API_KEY']}"
+    key = CGI.escape ENV['MORPH_API_KEY'] || ""
```
henare: Defaulting this to an empty string doesn't feel like the right thing to do here, especially not to just [fix a test](https://github.com/openaustralia/planningalerts/pull/1139/commits/154f3f0eeb9fb9ede76dfcb45839e2feb6f0ea1a). Since morph.io requires an API key it seems reasonable to throw an exception if it's not set.

To get the tests passing it seems like we don't need a valid API key (there wasn't one before) but we do need to set one. So we can just add an obviously fake one to `.env.test`.

henare: I'd also suggest keeping the style consistent in this method by using parentheses around the parameter passed to `.escape`, like the previous line where `query` is being set.

equivalentideas: > Since morph.io requires an API key it seems reasonable to throw an exception if it's not set.

I really like this idea, it provides an opportunity to give some helpful advice for a developer in this situation to point the in the right direction.

TheChirpyWitch: Hey @henare ! By adding the null string, we are not defaulting it, merely forcing CGI.escape to behave like a string. When the the whole query is written as:
```
query = CGI.escape("select * from `data` where `date_scraped` >= '#{start_date}' and `date_scraped` <= '#{end_date}'")
    "https://api.morph.io/#{morph_name}/data.json?query=#{query}&key=#{ENV['MORPH_API_KEY']}"
```

> ENV['MORPH_API_KEY'] isn't converted to utf-8 and hence the URL is malformed.

In adding null string, I was making sure that the key becomes a string, but I checked once again and even if I simply make **ENV['MORPH_API_KEY']** a different variable and then CGI.escape it, it seems to work fine. It was only because the **ENV['MORPH_API_KEY']** was passed as a simple string in the older code, it wasn't working.

TheChirpyWitch: Hey @equivalentideas ! It does seem reasonable to throw an exception when the API key isn't set at all, but in this case the key was wrong and rake did throw an error. In any case, let me if I can help with adding an exception. :smile:

TheChirpyWitch:
Hey again! Now I remembered why I had added the null string - for some reason, Travis CI failed without that. I have no clue why this is happening, if either of you( @henare and @equivalentideas ) have any clue, please do tell me.
I adding the null string - don't want the Travis CI build to fail. :smile:

henare:
>I have no clue why this is happening, if either of you( @henare and @equivalentideas ) have any clue, please do tell me.

@TheChirpyWitch I mentioned it in my [comment above](https://github.com/openaustralia/planningalerts/pull/1139#discussion_r111857211):

>To get the tests passing it seems like we don't need a valid API key (there wasn't one before) but we do need to set one. So we can just add an obviously fake one to `.env.test`.

Don't forget [this from my review](https://github.com/openaustralia/planningalerts/pull/1139#pullrequestreview-33128085):

>...I also think we should add a test for this method. It currently doesn't have a test so it's a good idea to add one while we're working on it. We're also changing the behaviour here so it's a good idea to add a test to check for that behaviour.

Once you've done that let me know and I'll review this again :smile:

equivalentideas:
@TheChirpyWitch this PR sounds like it's really close to finishing, any chance you'll be able to hack on this again? or was it just for the Outreachy application. No pressure :) I'm just looking to close inactive PRs 🎐

Issue #1120, opened 2017-03-07
"Return to" text and link on "now check your email" page broken


henare:
![selection_001](https://cloud.githubusercontent.com/assets/48945/23643881/87fc2cce-0358-11e7-94ba-b13acf7b383b.png)

equivalentideas:
That should say something like 'applications near 123 Parramatta Rd, Annandale NSW 2038' and link to the search for that page, e.g. https://www.planningalerts.org.au/?q=123+Parramatta+Road%2C+Annandale%2C+New+South+Wales

ShubhiL:
I would like to solve this issue.
Since, it's my first time on Github, I will ask you questions and work on this.

Liceth:
Hi @henare 
I will try to work on this.

akve17:
Hi @henare 
I have recently made some commits on this issue. Please review them.

equivalentideas:
@akve17 could you please open a [Pull Request](https://help.github.com/articles/about-pull-requests/) with your fix?

equivalentideas:
@akve17 could you please open a [Pull Request](https://help.github.com/articles/about-pull-requests/) with your fix?

equivalentideas:
Fixed by https://github.com/openaustralia/planningalerts/issues/1142

PR #1140, opened 2017-04-13
Fix for Broken Link "Applications Near "

akve17:
The broken link "Applications Near" was not working because the parameter '@address ' was not set. This parameter has been set in the alerts controller file.


akve17:
It gives an error "secret_key was not set". This error occurs because the morph key is not set.

equivalentideas:
@akve17 Thanks for this contribution and for explaining how the fix works in your PR description 👍.

A good tip is that if you include the text "Fixes #1120" in your PR description, GitHub creates a link back to the issue and will close it automagically when this is merged.

I'm going to close this PR however because it includes many code changes that are unrelated to the fix and are unexplained. Are they to do with getting your local installation to work?

Could you open I new PR with just the code changes required to fix the issue? I'm happy to help you with the steps to get there if that helps :)

akve17:
@equivalentideas Thanks Luke. I had purposely added those files , so that the new candidates who are working on planning alerts do not face the same issues, but if they are not necessary i will soon make another commit with only the relevant changes.

equivalentideas:
In a Pull Request, it's best to only include the changes that fix the issue you're referring to and exclude all other changes. Submit an individual PR with each change to the application you want to make. If you're making a big change, that changes a lot of code, it's best to explain why each bit is being changed and why it needs to be. Otherwise it's really hard to review the PR. 

>  I had purposely added those files 

There's no way for us to know this, because your PR description says you're just trying to fix the "Applications Near" text. We just have to assume it's a mistake. 

Does that help clear things up? :)

I see you've made [a new PR](https://github.com/openaustralia/planningalerts/pull/1141/files), but it also doesn't explain the purpose of most of the code changes, such as removing the `.env` file.

PR #1142, opened 2017-04-18
'Return to' text and link on 'now check your email' page fixed (#1120)

akve17:
This will fix the issue #1120 -'Return to' text and link on 'now check your email' page
The link was broken because the 'address' parameter was not set. This parameter has been set in the alerts controller file to the street address of the respective application.

### DIFF
--- a/app/controllers/alerts_controller.rb
+++ b/app/controllers/alerts_controller.rb
@@ -12,6 +12,7 @@ class AlertsController < ApplicationController
   end
 
   def create
+    @address = params[:alert][:address]
     @alert = NewAlertParser.new(
       Alert.new(
         email: params[:alert][:email],


equivalentideas:
Nice work @akve17 This does the job. Very nice clear PR that'd really easy to review.

One tip, your useful description of why you made the change would also be great to add to the longer description on the commit message :)

> The link was broken because the 'address' parameter was not set. This parameter has been set in the alerts controller file to the street address of the respective application.

henare:
I'd also be interested to know how this regression was introduced.

Including tests to ensure it doesn't happen again is a good idea too.

equivalentideas:
@henare I introduced it when overhauling the signup logic in January :) https://github.com/openaustralia/planningalerts/pull/1077/files#diff-6b87d4897e0f8f3a184e3b24d9ad7142L15

I'll add a check for this to the signup integration tests so it doesn't happen again 👍

henare:
Oh, oops.

I also note you could DRY this up by using the already created `@application`: https://github.com/openaustralia/planningalerts/pull/1142/files#diff-6b87d4897e0f8f3a184e3b24d9ad7142R19

equivalentideas:
Nice 👍

Issue #676, opened 2015-06-19
Send devise emails using background queue


henare:

hisayohorie:
https://github.com/plataformatec/devise#password-reset-tokens-and-rails-logs

According to this doc, we could change the log level. since we are using recoverable module, we should lower the log level so the token won't get leak from the log? Would it be ok if I make the change? @equivalentideas

henare:
@hisayohorie well spotted! :eyeglasses: When we make the change this issue describes we should also set the log level to warn.

hisayohorie:
@henare 🤓 ok will make the change and PR!

PR #1143, opened 2017-04-19
added a devise method send_devise_notification in user.rb, and change…

hisayohorie:
…d the config.log_level to :warn from :info to send the devise email using background que

this is to fix <a href="https://github.com/openaustralia/planningalerts/issues/676"> issue#676</a>
@henare I haven't wrote the test for this, because I wasn't sure if I should since it's a kind of change that buried in devise..please let me know the next step 🤓

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,4 +9,8 @@ class User < ActiveRecord::Base
   def set_api_key
     self.api_key = Digest::MD5.base64digest(id.to_s + rand.to_s + Time.now.to_s)[0...20]
   end
+
+  def send_devise_notification(notification, *args)
+    devise_mailer.send(notification, self, *args).deliver_later
+  end
 end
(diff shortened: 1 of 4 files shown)


henare:
Nice one @hisayohorie :sparkles: 

It would have been good to link off to this to explain what the new method you've added is doing: https://github.com/plataformatec/devise#activejob-integration

Next time you should also split these two changes into separate commits. Generally if your commit message includes "...and..." then they should be two separate commits :smile: This makes it much easier to work out what the purpose of a commit is in the future which will help you when you're debugging the application.

I agree that I'm not sure there's much point in a unit test for the new method but we should test that this works somehow. A good idea would be an integration test and what-do-you-know there's already one that covers [_some_ of this functionality](https://github.com/openaustralia/planningalerts/blob/fedb4bb271920207a216e4e16c06aded70e41266/spec/features/user_sign_up_spec.rb#L3-L17) but not the email part. So I think you should expand that test to cover the confirmation email :email: :writing_hand:

hisayohorie:
@henare I had hard time figuring out why the testsI added don't pass, and found <a href="https://stackoverflow.com/questions/37753251/actionmailer-not-delivering-confirmation-emails-in-test-environment-rails-4">this</a>. is this the right way to go?

PR #1144, opened 2017-04-19
Issue#669 bad page number parameters causing ArgumentError

### LINKED ISSUE #669: Bad page number parameters cause ArgumentError to be raised
https://www.planningalerts.org.au/?page=2[&q=George+Street%2C+Redfern+NSW
## Backtrace

```
line 18 of [PROJECT_ROOT]/app/controllers/applications_controller.rb: index
```

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/37075/faults/5869537)

hisayohorie:
Fixes #669.

I wrote 2 test for this method because
when the string passed starts with integer, to_i returns the same integer (ex. "2K%&)".to_i = 2) 
and other time it returns 0 (ex. "%&43{".to_i = 0)

and I'm not super sure this is the appropriate way to do it? please let me know!
@equivalentideas @henare

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -49,10 +49,10 @@ class ApplicationController < ActionController::Base
   # this method is to respond to the will_paginate bug of invalid page number leading to error being thrown.
   # see discussion here https://github.com/mislav/will_paginate/issues/271
   def validate_page_param
-    unless params[:page].present? && params[:page].to_i > 0
-     params[:page] = nil
+    if params[:page].present? && params[:page].to_i > 0
+      params[:page] = params[:page].to_i
     else
-     params[:page] = params[:page].to_i
-   end
+      params[:page] = nil
+    end
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -46,6 +46,8 @@ class ApplicationController < ActionController::Base
     devise_parameter_sanitizer.permit(:sign_up, keys: [:name, :organisation])
   end
 
+  # this method is to respond to the will_paginate bug of invalid page number leading to error being thrown.
+  # see discussion here https://github.com/mislav/will_paginate/issues/271
   def validate_page_param
     unless params[:page].present? && params[:page].to_i > 0
      params[:page] = nil
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::Base
   # Scrub sensitive parameters from your log
   # filter_parameter_logging :password
 
-  before_filter :set_header_variable, :set_view_path
+  before_filter :set_header_variable, :set_view_path, :validate_page_param
   before_action :configure_permitted_parameters, if: :devise_controller?
 
   def authenticate_active_admin_user!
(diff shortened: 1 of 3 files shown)


hisayohorie:
@henare sure! I will look into it!

hisayohorie:
@henare this is the discussion around the issue, and seems the best practice isn't solidified? 
https://github.com/mislav/will_paginate/issues/271

henare:
@hisayohorie nice one on tracking down that discussion. That makes me :+1: this solution then.

I think we should add a comment above that method so people know why we've added it. You could even include a link to that discussion. Presumably when it's resolved upstream we'll be able to remove this workaround :sparkles: 

Once that's done let's merge and :shipit:

henare:
@hisayohorie I noticed you updated this so I thought I'd give it a more thorough check.

>Once that's done let's merge and :shipit:

OK I got a bit ahead of myself :smile: I'll add a couple of comments :speech_balloon:

app/controllers/application_controller.rb
```diff
@@ -45,4 +45,14 @@ def ssl_required?
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:sign_up, keys: [:name, :organisation])
   end
+
+  # this method is to respond to the will_paginate bug of invalid page number leading to error being thrown.
+  # see discussion here https://github.com/mislav/will_paginate/issues/271
+  def validate_page_param
+    unless params[:page].present? && params[:page].to_i > 0
```
henare: [Do not use unless with else. Rewrite these with the positive case first.](https://github.com/bbatsov/ruby-style-guide#no-else-with-unless)

I also notice your indentation is a little out on the params lines - you need an extra space.

spec/controllers/applications_controller_spec.rb
```diff
@@ -15,6 +15,21 @@
         get :index
         expect(assigns[:rss]).to be_nil
       end
+
+      it "should try to convert page param to an integer" do
```
henare: In my first review I didn't check where these tests were placed. They're under `#index rss feed` and since this isn't related to the RSS feed these tests should not be here.

But where should they go? I didn't know so I hunted around. I discovered that `ApplicationController` behaviour that does things like global error checking should be tested by [defining an anonymous controller](https://www.relishapp.com/rspec/rspec-rails/v/3-0/docs/controller-specs/anonymous-controller).

So create a new file in this directory called `application_controller_spec.rb`, with something like this (and removing the tests from this file):

```ruby
require "spec_helper"

describe ApplicationController do
  controller do
    def index
      render text: nil
    end
  end

  describe "#validate_page_param" do
    it "should try to convert page param to an integer" do
      get :index, page: "2%5B&q"

      expect(controller.params[:page]).to eq(2)
    end

    it "should default to page nil when no page number param is given" do
      get :index, page: "%5B&q"

      expect(controller.params[:page]).to eq(nil)
    end
  end
end
```

henare:
@hisayohorie Once you're ready for me to take another look please assign back to me :handshake:

spec/controllers/applications_controller_spec.rb
hisayohorie: will do!

spec/controllers/applications_controller_spec.rb
```diff
@@ -15,6 +15,7 @@
         get :index
         expect(assigns[:rss]).to be_nil
       end
+
```
henare: Remove this stray extra line.

henare:
Deployed :shipit:

Issue #681, opened 2015-06-29
Automatically convert HTML entities from scrapers back into normal characters


henare:
Often scrapers miss converting HTML entities in the description field, e.g. https://www.planningalerts.org.au/applications/530099

When we suck in data from scrapers we should automatically convert these back into normal characters.

hisayohorie:
I am working on this issue!

mlandauer:
As mentioned in #1145 it's probably simpler to delegate responsibility for this always to the scraper. It has full knowledge of where the data is coming from so should handle it in whatever way makes sense to turn it in to proper normal utf8.

However, maybe we could do some sanity checking that warns us in some way if there are fields with things that look like html entities in them. Then, we know we need to fix the scraper.

PR #1145, opened 2017-04-20
Issue#681 Automatically convert HTML entities from scrapers back into normal characters

hisayohorie:
this is to fix <a href="https://github.com/openaustralia/planningalerts/issues/681"> issue#681</a>

I added htmlentities gem, and created a method called decode in applications_helper, and used decode in applications/index.html.haml as well as _application.html, and a wrote a test in applications_helper_spec.rb
@equivalentideas @henare

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source "http://rubygems.org"
+source "https://rubygems.org"
 
 gem 'rails', '4.2.7.1'
 gem 'mysql2', '> 0.3'
@@ -79,7 +79,6 @@ group :test do
 end
 
 group :development do
-  gem 'letter_opener'
   gem 'guard'
   gem 'listen', '< 3' # Used by guard. 3.0.0+ includes ruby_dep 1.5.0 which needs Ruby 2.2+
   gem 'guard-rspec'
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source "https://rubygems.org"
+source "http://rubygems.org"
 
 gem 'rails', '4.2.7.1'
 gem 'mysql2', '> 0.3'
@@ -78,6 +78,8 @@ group :test do
 end
 
 group :development do
+  gem 'htmlentities'
+  gem 'letter_opener'
   gem 'guard'
   gem 'listen', '< 3' # Used by guard. 3.0.0+ includes ruby_dep 1.5.0 which needs Ruby 2.2+
   gem 'guard-rspec'
@@ -99,6 +101,14 @@ end
 
 group :test, :development do
   gem 'rspec-rails', '~> 3'
+  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+ gem 'byebug'
+ gem 'pg'
+ # Access an IRB console on exception pages or by using <%= console %> in views
+# gem 'web-console', '~> 2.0'
+
+ # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
+ gem 'spring'
 end
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,7 @@ gem 'everypolitician-popolo', git: 'https://github.com/everypolitician/everypoli
 # Using master until an updated version of the Gem is released https://github.com/ciudadanointeligente/writeit-rails/issues/4
 gem 'writeit-rails', git: 'https://github.com/ciudadanointeligente/writeit-rails.git', branch: 'master'
 gem 'mime-types', '~> 2.99' # our writeit gem version is incompatible with newer versions
+gem 'htmlentities'
 
 group :test do
   gem 'capybara'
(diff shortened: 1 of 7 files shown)


Gemfile
```diff
@@ -1,4 +1,4 @@
-source "https://rubygems.org"
+source "http://rubygems.org"
```
henare: This shouldn't be changed to insecure HTTP.

Gemfile
```diff
@@ -78,6 +79,7 @@ group :test do
 end
 
 group :development do
+  gem 'letter_opener'
```
henare: There's a few additions to the Gemfile from here down that don't seem to be related to this change so they should not be included in this pull request.

app/helpers/applications_helper.rb
```diff
@@ -24,6 +24,12 @@ def scraped_and_received_text(application)
     text
   end
 
+  def decode(description)
+    coder = HTMLEntities.new
+    first_decoding = coder.decode(description)
```
henare: Why are there 2 decode calls?

app/helpers/applications_helper.rb
```diff
@@ -24,6 +24,12 @@ def scraped_and_received_text(application)
     text
   end
 
+  def decode(description)
```
henare: We already do a lot of manipulation to the description in `Application#description`. Since this would put this logic in one place I think we should probably put this there too.

spec/helpers/applications_helper_spec.rb
```diff
@@ -85,6 +85,12 @@
     end
   end
 
+  describe "decode" do
+    it "should decode the html entities such as &amp; to &" do
```
henare: For some reason my instincts are telling me we need more tests around the edge cases. My instincts could be wrong :smile: 

Are there situations where we wouldn't want to decode HTML entities? Are the potential security issues we could be introducing to the page?

It might be good to have a look on PlanningAlerts for real-world examples of where this is a problem and specifically test them too. I always think if you've found a problem it's best to add tests for that exact problem, and the other success, failure, and edge cases you can think of.

Gemfile
henare: It looks like one of these additions has also [broken the build](https://travis-ci.org/openaustralia/planningalerts/builds/223779317) so hopefully fixing this up with fix the build too :sparkles:

app/helpers/applications_helper.rb
hisayohorie: @henare I had to call it twice to convert some entities. The first call converts &amp; to &, and the second call converts &#8217 to ', for example. If you know of better ways to do it or some insight to start, please let me know.

spec/helpers/applications_helper_spec.rb
hisayohorie: @henare hmm I can't think of the case we don't want to decode it or potential security issues...any insights? I agree with looking into PlanningAlerts real world examples. I will look into it this week.

Gemfile
hisayohorie: true!

Gemfile
hisayohorie: also true!

app/helpers/applications_helper.rb
henare: That's really helpful @hisayohorie :+1: I don't know of any better ways off the top of my head but this would be a great place to put a comment to explain why we're calling this twice.

Whenever code is difficult to understand first try to simplify it so it is easy to understand :smile: but if you can't then that's when a comment is the way to go :ok_hand:

spec/helpers/applications_helper_spec.rb
henare: >I can't think of the case we don't want to decode it or potential security issues...any insights?

Is there a case where it could potentially inject a `<script>` tag? e.g. by decoding an HTML entity for `<`? I'm honestly not sure if this is a problem here but this area of code, where you're manipulating outside external data and injecting it onto a page, is the place to be careful for this kind of security issue.

hisayohorie: I tried using <script></script> both encoded and decoded, and both cases browser treated it like a text. Can you think of any other ways  of testing it?

henare: Nice :smile: that should be all good then. I guess Rails is still automatically escaping it.

hisayohorie: I moved the decode under application#description and added a comment. still need to look into real life examples and write a test accordingly!

hisayohorie:
@henare Here is the PR for #1145 !

handelaar:
(If we are fixing stray entities in the description field we should do the same to the address field. Plenty of street names contain apostrophes. )

mlandauer:
It pains me to close an open PR without merging it when it's clear how much work has gone into this but unfortunately this has been open for long enough that it would take a bit to get it working against the current master.

Also, I wonder if #681 is suggesting fixing the problem in the right way? Strictly the scraper should be the one that knows how to handle html entities and any other encoding issues because it has the context of the page the data is being scraped from.

PR #1151, opened 2017-05-05
Change how abuse reports are sent so they get delivered

### LINKED ISSUE #689: Comment report emails can be rejected
Because they appear to come from the person, Google Groups is rejecting some. See #688.

henare:
Before this change we were sending abuse reports with the reporter's
address as the from address. It's common that a domain with a DMARC
policy would make the report get hard bounced and then Cuttlefish would
blacklist all further reports so we'd never see them.

This changes the from address to our address and sets the reporter's
address as the reply_to. This means the messages should get delivered
and administrators can still just hit reply if they want to respond to
the reporter.

Fixes #689.

### DIFF
--- a/.env
+++ b/.env
@@ -87,7 +87,7 @@ THEME_NSW_CUTTLEFISH_PASSWORD=xxxxxxxxxxxxxxxxxxxx
 
 THEME_NSW_GOOGLE_ANALYTICS_KEY=UA-3107958-12
 
-# CAPTCHA for reporting comments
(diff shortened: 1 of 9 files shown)


henare:
I've tested this in staging and confirmed it fixes the problem. I'm not yet merging and deploying because we're currently getting about 30 reports a day, almost all of which are spam. That means the existing anti-spam honeypot isn't working and so we need to put Google's CAPTCHA on this page.

PR #1154, opened 2017-05-25
Add 'Back PlanningAlerts' banner to bottom of email alerts

equivalentideas:
This replaces the more subtle call for donations we have now, and introduces the backers concept.

This is much more obvious visually—it's high contrast to the rest of the email. This is a tough balance because we don't want to overly distract from the content people came for. Setting this apart from the other content does reflect the conceptual difference from the other information in the email, so that also could make it simpler overall.

I want to test out what impact this has. We currently get about 30 people a month clicking through this link currently, I think we'll get more if we make it more obvious. We'll be able to judge the effectiveness of this by how many people click through, and donate. We've got tracking set up for that already.

screen shot 2017-05-25 at 3 26 52 pm
screen shot 2017-05-25 at 3 29 02 pm


equivalentideas:
I'm gonna merge this cause I wanna get it out and see what happens today. Tests clear.

Issue #1058, opened 2016-11-23
Send new users welcome email that explains who is behind the project


equivalentideas:
Most people who donate (excluding end of financial year prompt) do so soon after signing up to their first alert. These people should be invited to join the Crew.

Send them an email, inviting them to join the Crew, and regular, increasing intervals, e.g. after 1st week, then 2 weeks later, then 1 month later, etc.

We should also ask for their feedback about their experience with PlanningAlerts so far. They probably have great ideas for improvements etc. and would appreciate being asked.

These emails should include a way to opt out of these emails permanently.

equivalentideas:
Here's the text from our three most successful end of financial year (EOFY) emails to PlanningAlerts users asking them to donate.

## EOFY 2016 Early June email

Subject: Last chance to support PlanningAlerts this financial year

> 
> Hi! I’m Henare Degan. I’m a director of the OpenAustralia Foundation, the charity that runs PlanningAlerts.
> 
> Thank you for being part of another huge year for PlanningAlerts – our biggest year yet!
> 
> We run PlanningAlerts as a free service because we believe you should be informed and have a say on changes to you local area. We can do this because of [donations](http://oaf.org.au/donate/planningalerts) we receive from the people who use it, like you.
> 
> This year we’ve seen thousands more people commenting on development applications, and now writing to local councillors as well. These really do make an impact. With the help of many wonderful volunteers we’ve extended the project’s coverage to make it available to over 65% of people in Australia.
> 
> If everyone using PlanningAlerts made a small donation today, the project would be easily sustained for another year. All donations go directly to supporting and developing our projects to transform democracy in Australia.
> 
> **[If PlanningAlerts is useful to you, please take a moment to make a tax-deductible donation or become a regular supporter to keep it running and growing for another year.](http://oaf.org.au/donate/planningalerts)**
> 
> **http://oaf.org.au/donate/planningalerts**
> 
> Thanks for your support
> 
> Henare Degan,
> OpenAustralia Foundation

## EOFY 2016 End of June email

Subject: Last chance to support PlanningAlerts this financial year

> Hi! I’m Henare Degan. I’m a director of the OpenAustralia Foundation, the charity that runs PlanningAlerts.
> 
> I emailed you a couple of weeks ago asking for your support and hundreds of people have donated since then—thank you all! I realise you’re busy so this is the last time we’ll ask for your help this year.
> 
[...]

PR #1170, opened 2017-06-16
Extract new Class AlertSubscriber from Alert for representing the person who is subscribed

equivalentideas:
This is a first step towards being able to send people welcome emails to PlanningAlerts. We'll also be able to use this new class to simplify a bunch of another existing behavior too.

This PR:

* Adds a new model and table for AlertSubscriber
* Adds a before_create callback on Alert that associates new alerts with an AlertSubscriber
* Creates AlertSubscribers for all existing Alerts

There's a few long TODOs on there because the bigger feature this is part of is still a bit of a journey to finish, but I think this initial bit is deployable and I'd like to get the migrations run on production before too many build up.

This is part of #1058


app/models/alert.rb
```diff
@@ -12,13 +12,26 @@ class Alert < ActiveRecord::Base
   scope :active, -> { where(confirmed: true, unsubscribed: false) }
   scope :in_past_week, -> { where("created_at > ?", 7.days.ago) }
 
+  before_create :attach_alert_subscriber
```
equivalentideas: What happens if the alert can't be saved for some reason? Does it still create the subscriber? We should check this.

equivalentideas:
Working on the next steps I realised there's a problem in Alert.create_alert_subscribers_for_existing_alerts that I missed. I've added "[WIP]" to the title and assigned back to me to fix.

equivalentideas:
I've fixed that issue above and tested this some more locally. I'm going to merge this and deploy with migrations.

equivalentideas:
This is deployed and the migrations run. Took 7836.3201s e.g. 2.17 hours to run the migration on production with 120,000+ alerts. Created 83029 AlertSubscriber records.

PR #1171, opened 2017-06-16
Fix timezone dependent tests

henare:
@equivalentideas here's the fixes for the timezone dependent tests you asked about in 3d1a9ac07c52...7b2c8cff45b1. Can you please review and merge? :timer_clock:

PR #1174, opened 2017-06-18
Added model, controller and view for suggested councillors. 

hisayohorie:
It passes test for now!  some polishing may needed.


.gitignore
```diff
@@ -9,6 +9,7 @@ tmp/sass-cache
 tmp/binlog
 tmp/rspec_guard_result
 tmp/capybara
+tmp/letter_opener
```
equivalentideas: What's this for?

app/assets/javascripts/suggested_councillors.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
equivalentideas: Let's remove this file, we can add a script file when we need it.

app/assets/stylesheets/suggested_councillors.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the suggested_councillors controller here.
```
equivalentideas: Let's remove this file, we can add a stylesheet when we need it.

app/helpers/suggested_councillors_helper.rb
```diff
@@ -0,0 +1,2 @@
+module SuggestedCouncillorsHelper
```
equivalentideas: I think we remove this too, and add a helper file when we need it.

config/routes.rb
```diff
@@ -102,11 +102,14 @@ def matches?(request)
         get :per_week
       end
     end
+    resources :suggested_councillors, only:[:new, :create] 
     collection do
       get :test_feed
     end
   end
 
+  resources :councillors, only: [:new, :create]
```
equivalentideas: What's this for?

spec/controllers/suggested_councillors_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
equivalentideas: This call to a `rails_helper` file that we don't have is breaking the tests suit. This is another file created by the rails generator. I think just remove the whole file until we want to write controller tests for this.

spec/features/admin_loads_councillors_spec.rb
```diff
@@ -59,7 +59,7 @@
 
   context "when some councillors have been removed from office" do
     around do |test|
-      Timecop.freeze(2016, 10, 11) { test.run }
+      Timecop.freeze(2016, 10, 15) { test.run }
```
equivalentideas: This doesn't seem relevant to these changes? I think we added this as a test while we were pairing, but @henare has fixed the timezone issue with #1171 . We should remove this change from the branch.

spec/features/admin_loads_councillors_spec.rb
```diff
@@ -72,7 +72,6 @@
       visit admin_authority_path(authority)
 
       click_button "Load Councillors"
-
```
equivalentideas: I think this blank line between the test 'exercise' and 'verify' stages of the test makes it clearer. Would you mind adding it back @hisayohorie ?

spec/helpers/suggested_councillors_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
equivalentideas: Because we're removing the helper file, we should remove this test file as well. We try not to have pending test files sitting around, they never seem to get filled in.

spec/models/suggested_councillor_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe SuggestedCouncillor, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
equivalentideas: I think you could have some simple tests here for the validations. They seem quite important and worth having coverage for.

.gitignore
hisayohorie: I was using the letter opener gem for other issues, indeed this is irrelevant to keep it here

config/routes.rb
hisayohorie: this is the time when I was trying to add a column in councillor table. not relevant now.

spec/features/admin_loads_councillors_spec.rb
hisayohorie: Ok will do!

spec/features/admin_loads_councillors_spec.rb
hisayohorie: Sure I will!

app/controllers/suggested_councillors_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    @suggested_councillor = SuggestedCouncillor.new
+    @authority = Authority.find(params[:authority_id])
+  end
+
+  def create
+    @suggested_councillor = SuggestedCouncillor.new(suggested_councillor_params)
+    @authority = Authority.find(params[:authority_id])
+    @suggested_councillor.authority= @authority
+
+      if @suggested_councillor.save
```
equivalentideas: Is the indentation of this block off?

app/controllers/suggested_councillors_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        flash[:notice] = "Thank you"
+        redirect_to authority_url(@authority.short_name_encoded)
```
equivalentideas: I'm interested in the choice to redirect to the authority page here. What's your thinking @hisayohorie ?

This is an area where having a single commit for this choice, with a description of why it was made, would make understanding the change easier.

app/controllers/suggested_councillors_controller.rb
```diff
@@ -0,0 +1,25 @@
+class SuggestedCouncillorsController < ApplicationController
+  def new
+    @suggested_councillor = SuggestedCouncillor.new
+    @authority = Authority.find(params[:authority_id])
+  end
+
+  def create
+    @suggested_councillor = SuggestedCouncillor.new(suggested_councillor_params)
+    @authority = Authority.find(params[:authority_id])
+    @suggested_councillor.authority= @authority
```
equivalentideas: Would it be possible to assign this when the SuggestedCouncillor is initialized if `@authority` is assigned first?

spec/models/suggested_councillor_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+
+RSpec.describe SuggestedCouncillor, type: :model do
+  it "should require a name" do
```
equivalentideas: You might be able to make these tighter using `subject` (see http://www.betterspecs.org/#subject ) and `be_valid` (using [Rails method `.valid?`](http://api.rubyonrails.org/classes/ActiveRecord/Validations.html#method-i-valid-3F) with [rspec predicate matchers](https://relishapp.com/rspec/rspec-expectations/v/3-6/docs/built-in-matchers/predicate-matchers))

Do you think something like this would be a bit clearer + more succinct?

```
describe "is invalid without an email" do
  subject { SuggestedCouncillor.new(name: "Jane", email: nil, authority_id: 1) }
  it { is_expected.to_not be_valid }
end

describe "is invalid without a name" do
  subject { SuggestedCouncillor.new(name: nil, email: "example@email.org", authority_id: 1) }
  it { is_expected.to_not be_valid }
end

describe "is invalid with a blank name" do
  subject { SuggestedCouncillor.new(name: "", email: "example@email.org", authority_id: 1) }
  it { is_expected.to_not be_valid }
end
````

Or maybe it's time to start using a library for these common validation tests we have https://github.com/thoughtbot/shoulda-matchers#activemodel-matchers

File: app/controllers/suggested_councillors_controller.rb (hunk 13 above)
Comment (hisayohorie) [inline app/controllers/suggested_councillors_controller.rb]: @equivalentideas you are suggesting to use `@authority.suggested_councillors.build(suggested_councillor_params)` , am I right?

File: app/controllers/suggested_councillors_controller.rb (hunk 11 above)
Comment (hisayohorie) [inline app/controllers/suggested_councillors_controller.rb]: yes 😞

File: app/controllers/suggested_councillors_controller.rb (hunk 12 above)
Comment (hisayohorie) [inline app/controllers/suggested_councillors_controller.rb]: good question, because I don't exactly know where it should be redirected 😅 
i wasn't sure if i should make a 'Thank you' page. I thought one I think we know and the contributor know for sure is which authority the suggested councillor belongs to, so I redirected to authority page. But it is a very loose thinking and I would love to know your suggestion!

File: spec/models/suggested_councillor_spec.rb [hunk 15]
```diff
@@ -0,0 +1,28 @@
...
+    subject { SuggestedCouncillor.new(name: "Milla", email: "", authority_id: 1) }
+    it { is_expected.to_not be_valid }
+  end
+
+  describe "is invalid without an name" do
+    subject { SuggestedCouncillor.new(name: nil, email: "test@test.com", authority_id: 1) }
+    it { is_expected.to_not be_valid }
+  end
+
+  describe "is invalid without a blank name" do
```
equivalentideas: Should that be "with a blank name" ?

app/controllers/suggested_councillors_controller.rb
```diff
@@ -0,0 +1,24 @@
+class SuggestedCouncillorsController < ApplicationController
+  def new
+    @suggested_councillor = SuggestedCouncillor.new
+    @authority = Authority.find(params[:authority_id])
+  end
+
+  def create
+    @authority = Authority.find(params[:authority_id])
+    @suggested_councillor = @authority.suggested_councillors.build(suggested_councillor_params)
```
equivalentideas: Very nice 👌

PR #1183, opened 2017-06-20
Added before_action:check_if_feature_flag_is_on to to enable feature flag

hisayohorie:


equivalentideas:
Looking really good @hisayohorie . I'd like to see a bit of a description on the PR :)

We should add a note about this to the .env file like the other feature flags and also to the readme about how to turn this on in development and production.

Issue #1180, opened 2017-06-20
Add basic styles to the councillor contribution form


equivalentideas:
It currently appears as:

![screen shot 2017-07-27 at 8 14 14 pm](https://user-images.githubusercontent.com/1239550/28665801-3966dd96-7308-11e7-8b1d-8bc5878ca129.png)

Extracted from #1160

equivalentideas:
So I think the basic things to get this into shape for the first pass are:

* [ ] get the heading down to a size that doesn't dominate the form, it should draw attention so people know what the page is about and lead people into the next action.
* [ ] Style the form input elements to match the general app style that works well, bolder labels and clearer inputs.
* [ ] Differentiate the submit form and add a councillor buttons, in both colour and proximity. I think the 'add a councillor' button should align to the left hand side of the inputs so it's a clear line down the form (sitting it the place that the new councillor form will appear). The submit button should be on the right (conventionally signifying progressing through something), and be a bit lower than the form itself. The submit button should be the colour of the submit buttons across the site, the add a councillor button should be grey maybe? or blue?
* [ ] align the form inputs, and the submitted council records horizontally, so that the names and emails of all the councillors submitted line up. This helps you see if there's a typo in the email formatting, and quickly scan up and down the names.

equivalentideas:
I'm mid way through this, working on [branch `councillor_contribution_basic_styles`](https://github.com/openaustralia/planningalerts/tree/councillor_contribution_basic_styles):

![screen shot 2017-07-28 at 6 37 16 pm](https://user-images.githubusercontent.com/1239550/28709509-f235ee64-73c3-11e7-93c9-fed370323bb3.png)

Haven't done anything with the submitted councillors yet:

![screen shot 2017-07-28 at 6 39 47 pm](https://user-images.githubusercontent.com/1239550/28709567-32b7b918-73c4-11e7-82a5-2074c1c598a7.png)

@hisayohorie it would be great to get your thoughts and feedback on this. Going in the right direction?

hisayohorie:
@equivalentideas yes, I think it looks great!

equivalentideas:
> @equivalentideas yes, I think it looks great!

Thanks @hisayohorie I'll continue along that path then 🚠

equivalentideas:
I've done the bits on [the list above](https://github.com/openaustralia/planningalerts/issues/1180#issuecomment-318574099) and a couple of small other bits, I'll open a PR 🖼

[...]

Issue #1179, opened 2017-06-20
When someone contributes councillor information thank them and provide an experience that fits our principles


equivalentideas:
Here are our principles for this project https://github.com/openaustralia/australian_local_councillors_popolo#design-principles

PR #1185, opened 2017-06-22
Suggested councillor council name

### LINKED ISSUE #1181: Make sure people know which council they're adding councillors to when they're contributing
Currently the page to add councillors to an authority doesn't say which authority it is. This is important information :)

See http://localhost:3000/authorities/marrickville/suggested_councillors/new

hisayohorie:
This will display the name of council the suggested councillors belong. 

screen shot 2017-06-21 at 11 50 27 pm

Fixes #1181 .

### DIFF
--- a/app/views/suggested_councillors/new.html.haml
+++ b/app/views/suggested_councillors/new.html.haml
@@ -1,3 +1,6 @@
+%p
+  Council Name: #{@authority.full_name}
+
 = form_for [@authority, @suggested_councillor], url: authority_suggested_councillors_path do |f|
   %fieldset
     %legend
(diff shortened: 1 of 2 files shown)


app/views/suggested_councillors/new.html.haml
```diff
@@ -1,3 +1,6 @@
+%p
```
henare: Should this be a header instead of a paragraph?

app/views/suggested_councillors/new.html.haml
```diff
@@ -1,3 +1,6 @@
...
+  Council Name: #{@authority.full_name}
```
henare: Since this line is so short it could go on the same line as the tag, i.e.:

```haml
%p Council Name: #{@authority.full_name}
```

hisayohorie:
I think the styling and wording are addressed in different issues #1179 #1180

PR #1187, opened 2017-06-23
Added CONTRIBUTE_COUNCILLORS_ENABLED explanation in readme

hisayohorie:
To explain how to toggle feature flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ You will also need to [configure the app to accept replies from councillors](#ac
 
 #### Global feature flag
 
+##### Writing to councillors feature
 You can toggle the availability of the writing to councillors feature on or off for the entire site with the environment variable `COUNCILLORS_ENABLED`.
 The feature is globally enabled when the value of `ENV["COUNCILLORS_ENABLED"]` is `"true"`.
 This flag is useful if you need to turn the feature _off_ globally.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -110,6 +110,8 @@ We set this in the [`.env`](https://github.com/openaustralia/planningalerts/blob
 COUNCILLORS_ENABLED=true
 ```
 
+Similarly, you can toggle the availability of the contributing suggested councillors feature on or off for the entire site with the environment variable `CONTRIBUTE_COUNCILLORS_ENABLED`.
```
equivalentideas: Nice simple wording. Do you think that this section deserves a heading, and a short explanation of what the feature is, like the general councillors section or the donation section, and then the explanation of how to turn it on and off?

When I read this, I wonder what I should set the environment variable to be. Should it be `true` like the `COUNCILLORS_ENABLED` one? Consistency with this stuff is important, they should all work the same way and be explained consistently.

hisayohorie: Agreed! I pushed another PR to address the consistency

equivalentideas:
Real nice. It could seem repetitive, but I think that's actually useful for documentation because it makes it predictable.

This has exposed one difference in the implementation of the feature flags. This documentation now suggests that you can disable the feature by setting the variable to anything but "true". But that's not the case, see https://github.com/openaustralia/planningalerts/blob/person_contributes_councillor_info/app/controllers/suggested_councillors_controller.rb#L29 . It just calls 'present?'. Would it return true for the String "false"? That seems confusing.

Here with the other implementation it's explicitly looking for "true" https://github.com/openaustralia/planningalerts/blob/person_contributes_councillor_info/app/models/authority.rb#L256

I think what you're written here in the readme is best and we should change the implementation to match as part of this PR. What do you recon @hisayohorie ?

README.md
equivalentideas: I recon add those commits to this PR so we have it all in one place.

equivalentideas:
@hisayohorie That's not going to pass now because the two step thing isn't implemented on this branch, i.e. the Contributor model isn't used in this world. I forgot about that on our call.

The better thing to have done would be to make the change so it worked for the current situation and then make the same change on the new PR as well if necessary. Because it's such a small change that should be fine.

I think you should [revert that commit](https://git-scm.com/docs/git-revert) here, and then do that ☝️ because we shouldn't merge something that breaks tests.

Sorry I got confused about that!

PR #1186, opened 2017-06-23
Contributor provides their info in the suggested councillor form

hisayohorie:
Fixes #1177.
screen shot 2017-06-23 at 9 55 13 am

Styling and wording needs to be addressed in #1179 and #1180.
I haven't write a test for `contributor.rb`, since the contributor data is optional, but are there way to test it somehow? @henare

PR #1189, opened 2017-06-27
Two step contribution

hisayohorie:
this fixes #1186 

I used a hidden_field to track the suggested_councillor_id in contributor table, and not sure if this is the best practice. Please let me know otherwise.

### DIFF
--- a/app/controllers/contributors_controller.rb
+++ b/app/controllers/contributors_controller.rb
@@ -1,0 +1,2 @@
+class ContributorsController < ApplicationController
+end
--- a/app/controllers/contributors_controller.rb
+++ b/app/controllers/contributors_controller.rb
@@ -1,2 +1,6 @@
 class ContributorsController < ApplicationController
+  def new
+    @contributor = Contributor.new
+  end
+
 end
--- a/app/controllers/contributors_controller.rb
+++ b/app/controllers/contributors_controller.rb
@@ -1,4 +1,6 @@
 class ContributorsController < ApplicationController
+  before_action :check_if_feature_flag_is_on
+
   def new
     @contributor = Contributor.new
   end
@@ -8,6 +10,18 @@ class ContributorsController < ApplicationController
     if @contributor.save
       flash[:notice] = "Thank you"
       redirect_to root_url
+    end
+  end
+
+private
+
+  def contributor_params
+    params.require(:contributor).permit(:name, :email)
+  end
+
+  def check_if_feature_flag_is_on
+    unless ENV["CONTRIBUTE_COUNCILLORS_ENABLED"].present?
+      render "static/error_404", status: 404
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/contributors_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def new
+    @contributor = Contributor.new
+  end
+
+  def create
+    @contributor = Contributor.new(contributor_params)
+    @suggested_councillor = SuggestedCouncillor.find(params[:contributor][:suggested_councillor_id])
+    if @contributor.save
+      @suggested_councillor.contributor_id = @contributor.id
```
equivalentideas: Could this be done on one line with something like `@suggested_councillor.update(confributor: @contributor)`?

app/views/contributors/new.html.haml
```diff
@@ -0,0 +1,20 @@
+%h1 Thank you!
+%p
+  This data you contributed will be reviewd
+  by an administrator, and go live shortly.
+= form_for @contributor do |f|
+  = f.hidden_field :suggested_councillor_id, value: params[:suggested_councillor_id]
```
equivalentideas: 84 is fine—hound shouldn't be piping up about line length anyway.

app/views/contributors/new.html.haml
```diff
@@ -0,0 +1,20 @@
...
+  %fieldset
+    %legend
+      Please tell us about yourself, so we can send you a little note of appreciation
```
equivalentideas: This is fine as 85 characters 👍

config/routes.rb
```diff
@@ -107,6 +107,8 @@ def matches?(request)
       get :test_feed
     end
   end
+  resources :contributors, only:[:new, :create]
+
```
equivalentideas: Stray extra line?

equivalentideas:
Nice refactor 🎷 merging 🚀 

For the future, on commit ba2b5224bfad36dc64a96af14d904e9fbfb39347 , if you're ever writing "and" in a commit message, it should almost certainly be two separate commits ;-)

app/controllers/contributors_controller.rb
```diff
@@ -0,0 +1,29 @@
+class ContributorsController < ApplicationController
+  before_action :check_if_feature_flag_is_on
+
+  def new
+    @contributor = Contributor.new
+  end
+
+  def create
+    @contributor = Contributor.new(contributor_params)
+    @suggested_councillor = SuggestedCouncillor.find(params[:contributor][:suggested_councillor_id])
```
henare: What happens if the councillor can't be found?

If you use a bang method here (i.e. `.find!`) then it would raise a 404 if it couldn't find the councillor. This is usually really handy to use in situations like this however it's probably not the behaviour we want.

app/controllers/contributors_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    if @contributor.save
+      @suggested_councillor.update(contributor: @contributor)
```
henare: If this update doesn't happen for some reason then no one will know. It's better to use the bang method here, `#update!`, to fail fast and hard.

Issue #1178, opened 2017-06-20
Let me add all the councillors for an authority in one contribution


equivalentideas:
Currently on the `person_contributes_councillor_info` branch, people can add information for one councillor at a time.

They should be able to add details for all the councillors for an authority in one contribution.

Extracted from #1160

equivalentideas:
Next step, @equivalentideas write test.

equivalentideas:
> Next step, @equivalentideas write test.

I've just changed the existing test in https://github.com/openaustralia/planningalerts/compare/8db2c8482ebd...145699192f88

This is an interesting bit that we could probably write a few more tests for as we define how it really works, but I this feels like the next most basic step.

As we spoke about last week @hisayohorie I image you might be adding a new CouncillorContribution model or something that connects a set of SuggestedCouncillors to the Contributor.

I think don't worry too much about validations and stuff yet, because we can describe the specific behavior we want in issues and then write tests and then add it. If you have ideas for that, go ahead and make those issues :) Again that form book will have useful advice for that.

equivalentideas:
Since  #1192 was merged, you can now make a contribution with as many councillors as you want which solves this issues. 🔨 Great work @hisayohorie ✨

PR #1192, opened 2017-07-05
Multiple contribution

hisayohorie:
This is a first half step for the issue #1178.

I created councillor_contribution model, changed the association and moved the methods in `suggested_councillor_controller` to `councillor_contributions controller`.

To reflect the change in the association, I changed the paths in the `spec/user_contributies_new_coouncillor_spec.rb`

`new.html.haml` is moved from `views/suggested_councillor` to `views/councilor_contributions`

next step is create a view to enable multiple contributions

I will work on that and push it on this branch.


app/controllers/councillor_contributions_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+private
+
+def councillor_contribution_params
+  params.require(:councillor_contribution).permit(suggested_councillors_attributes: [:name, :email])
+end
+
+def suggested_councillor_params
+  params.require(:suggested_councillor).permit(:name, :email, :councillor_contribution_id)
+
```
hisayohorie: fixed

equivalentideas:
> next step is create a view to enable multiple contributions
> I will work on that and push it on this branch.

Sounds good 👍

equivalentideas:
As this PR is a work in progress, it's good to add the "[WIP] " to the beginning of the name https://github.com/blog/1943-how-to-write-the-perfect-pull-request

equivalentideas:
This is looking really good @hisayohorie but there’s lots of test failures https://travis-ci.org/openaustralia/planningalerts/builds/250398000?utm_source=github_status&utm_medium=notification they look very straightforward to fix though :)

Let me know when this is ready and I'll get it merged 🚀

app/views/councillor_contributions/new.html.haml
```diff
@@ -0,0 +1,21 @@
+%h1 Add a new councillor for #{@authority.full_name}
+
+= form_for [@authority, @councillor_contribution], url: authority_councillor_contributions_path do |f|
+  %fieldset{:class =>"fieldContainer"}
```
hisayohorie: I didn't push any of these files hound!

hisayohorie:
I added a validation to reject the submission of another councillor form when it's empty to prevent render the form. I tried and looked around how to use `#new` action in two different ways, but couldn't quite understand how it would work. I may not understanding your suggestion very well @equivalentideas, so I would very much interested to know more about it when we pair program!

equivalentideas:
In a pair programming session @hisayohorie and I did a big refactor in https://github.com/openaustralia/planningalerts/pull/1192/commits/41a27515ac921c1cf55b37a7abe55c3105c5a21e .

We made the change [suggested here](https://github.com/openaustralia/planningalerts/pull/1192#pullrequestreview-49804074):

> I think what's making this confusing is that conceptually, we're just trying to submit this new contribution, but in the controller we're bouncing over to the edit and update actions. I think you'll find it simpler if you stay on the #new action and don't actually save our contribution till all the councillors are added. Conceptually we're building up an Array of suggested_councillors, which we'll commit all at once with our contribution. To make that work we'll need to add a new route for push to the councillor_contribution#new route. That will allow us to reload that page with new data pushed from the form (e.g. the existing councillors). This is definitely a tricky one, not straightforward rails at all :) you're doing great.

@hisayohorie what are the next steps with this PR? It might be ready to go? You should do a review for the style of this as well. Rubocop doesn't seem to run on the haml files, but I can see we need to update the Hash style in app/views/councillor_contributions/_contribution_form_input.html.haml to the Ruby 1.9 hash syntax.

hisayohorie:
@equivalentideas I'll fix the style on the haml file, and try to understand how our refactor actually works!

As for next step, we could work on ajaxifying the multiple contribution form. But not sure if it's in the scope of this milestone. Otherwise I think it's ready to be merged.

app/controllers/councillor_contributions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @councillor_contribution = @authority.councillor_contributions.build(councillor_contribution_params)
+
+    if @councillor_contribution.save
+      redirect_to new_contributor_url(councillor_contribution_id: @councillor_contribution.id)
+    else
+      render :new
+    end
+  end
+
+private
```
equivalentideas: This should be indented to the level of the methods it applies to https://github.com/bbatsov/ruby-style-guide#indent-public-private-protected

app/views/councillor_contributions/_contribution_form_input.html.haml
```diff
@@ -0,0 +1,15 @@
+%fieldset
+  %legend
```
equivalentideas: Is the indentation here wrong? It appears to me like the form fields will be indented inside the `<legend>`, which they should not be.

app/views/councillor_contributions/_contribution_form_input.html.haml
```diff
@@ -0,0 +1,15 @@
...
+    Add a councillor
+    = f.fields_for :suggested_councillors, @councillor_contribution.suggested_councillors.last do |suggested_councillor_field|
+      %p{:class => "nameField"}
+        = suggested_councillor_field.label :name
+        = suggested_councillor_field.text_field :name
+      %p{:class => "emailField"}
+        = suggested_councillor_field.label :email
+        = suggested_councillor_field.text_field :email
+  %p.button{:id => "addMoreCouncillor"}
```
equivalentideas: These buttons are indented bellow the `<fieldset>` element, but they actually aren't part of the [fieldset](https://html.spec.whatwg.org/multipage/form-elements.html#the-fieldset-element), which groups input elements for adding the councillors details.

I recon bump them up out of the fieldset.

app/views/councillor_contributions/new.html.haml
```diff
@@ -0,0 +1,17 @@
+%h1 Add a new councillor for #{@authority.full_name}
+%h2 The councillor(s) you submitted:
+= form_for [@authority, @councillor_contribution], url: authority_councillor_contributions_path do |f|
+  %ul
+    - @councillor_contribution.suggested_councillors[0...-1].each do |s|
+      %li
+        %dt Name
+        %dd
```
equivalentideas: This could be on one line with `%dd= s.name`.

app/views/councillor_contributions/new.html.haml
```diff
@@ -0,0 +1,17 @@
+%h1 Add a new councillor for #{@authority.full_name}
+%h2 The councillor(s) you submitted:
+= form_for [@authority, @councillor_contribution], url: authority_councillor_contributions_path do |f|
+  %ul
+    - @councillor_contribution.suggested_councillors[0...-1].each do |s|
+      %li
+        %dt Name
```
equivalentideas: These `%dt` and `%dd` elements are currently missing their `%dl` parent, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt

app/views/councillor_contributions/_contribution_form_input.html.haml
equivalentideas: @hisayohorie what do you recon about moving the buttons out of the partial into app/views/councillor_contributions/new.html.haml ? That way this partial keeps a tight focus on just the HTML for inputing a councillor, and the buttons for other actions stay separate.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -66,5 +66,63 @@
 
       expect(page).to have_content "Thank you"
     end
+
+    it "successfully submit suggested councillor(s) information with a blank field of add another councillors" do
```
equivalentideas: Because the feature is "Contributing a new councillor for an authority" I think we can drop the 'submit ...' from this, to make it just "successfully with 3 completed and one blank councillor". That's a more detailed and more focused description.

For the final test maybe "successfully with three councillors".

With these tests descriptions make a change to get it to fail and see what the message is on the command line. It should be a clear, short sentence if possible, e.g. "Contributing councillors for an authority successfully with three councillors". The 'with three councillors' might be redundant at this point, but because we've got one with a blank let's keep the numbers to make it clear whats going on for now I recon.

app/views/councillor_contributions/new.html.haml
```diff
@@ -0,0 +1,17 @@
+%h1 Add a new councillor for #{@authority.full_name}
+%h2 The councillor(s) you submitted:
```
equivalentideas: This heading the `<ul>` of submitted councillors shouldn't render when the person hasn't submitted any councillors yet.

app/views/councillor_contributions/_contribution_form_input.html.haml
```diff
@@ -0,0 +1,15 @@
...
+    %button{:formaction => new_authority_councillor_contribution_path} "Add another councillor"
+
+  %p.button{:class => "contribute"}
```
equivalentideas: Why do these buttons need to be in `<p>` tags?

app/views/councillor_contributions/_contribution_form_input.html.haml
hisayohorie: @equivalentideas you mean not to indent `%legend` under `%fieldset`?  I think `%legend` needs to be one level deeper indented under `%fieldset` to achieve this style
https://developer.mozilla.org/en/docs/Web/HTML/Element/fieldset 
I may be wrong as i am not very familiar with haml style.

app/views/councillor_contributions/_contribution_form_input.html.haml
hisayohorie: @equivalentideas makes sense. I moved the buttons in a partial.

app/views/councillor_contributions/_contribution_form_input.html.haml
hisayohorie: @equivalentideas  no need!

app/views/councillor_contributions/new.html.haml
hisayohorie: @equivalentideas took that out of the form

app/views/councillor_contributions/_contribution_form_input.html.haml
hisayohorie: ah now I got it.

equivalentideas:
@hisayohorie and I have been through this together one last time and it looks good to us. Merging!

PR #1193, opened 2017-07-06
Multiple contribution cocoon gem

### LINKED ISSUE #1178: Let me add all the councillors for an authority in one contribution

hisayohorie:
fix #1178 
I used cocoon gem to dynamically append form field for suggested_councillors.
This is branched off from my multiple_contribution branch, so if it's better to not use the gem and write javascript, I'm happy to write it. Please let me know! @henare @equivalentideas


equivalentideas:
> so if it's better to not use the gem and write javascript, I'm happy to write it.

The gem looks fine, but it doesn't seem to provide for a non-javascript fallback. We use the progressive enhancement to make accessible and resilient sites, (here's some more from the GDS about that https://gdstechnology.blog.gov.uk/2016/09/19/why-we-use-progressive-enhancement-to-build-gov-uk/ and @adactio is always grand on this https://adactio.com/articles/11481 ).

I think it'll also be simpler to write this ourselves as we get into validations and stuff like that. So I recon the first step is making it work without JS, and then we can open up new issues for enhancing the interaction with stuff like adding another councillor without reloading the page. Does that sound good?

equivalentideas:
What do you think about https://github.com/openaustralia/planningalerts/pull/1193#issuecomment-314098297 @hisayohorie ?

Should we close this one and focus on the other PR https://github.com/openaustralia/planningalerts/pull/1192 ?

hisayohorie:
@equivalentideas sure,I'll close this PR and focus on #1192

Issue #1182, opened 2017-06-20
Help people collect councillor information by providing them some instructions on where to get it


equivalentideas:
Collecting councillor information can be tricky if you don't know where to find it.

hisayohorie:
I think this tutorial fro setlist.fm is a good example

http://www.setlist.fm/tutorial

hisayohorie:
go to the council website. To be sure, enter the province name where the council exist
(ex. "Cessnock City Council, New South Wales”, "George Town Council, Tasmania”)
[embed google search bar]
make sure it open the new tab so people can see the instruction

Typically, There are a navigation menu says something like 

- About Council
- Council
- Your Council
- Our Council
- About us

<img width="1246" alt="screen shot 2017-06-26 at 7 30 39 pm" src="https://user-images.githubusercontent.com/12241881/27592539-b7e475be-5b22-11e7-84f9-01c6af0f1089.png">
<img width="1254" alt="screen shot 2017-06-27 at 10 28 11 am" src="https://user-images.githubusercontent.com/12241881/27592876-9ff8dc82-5b23-11e7-8c87-8aeae5d49aab.png">
<img width="1160" alt="screen shot 2017-06-27 at 10 52 27 am" src="https://user-images.githubusercontent.com/12241881/27594089-e6d6f7f8-5b26-11e7-9618-9479830eba0b.png">

You may have to click the tab, or sometime you have to put your cursor on the tab and the submenu show up. 
<img width="1162" alt="screen shot 2017-06-27 at 11 13 06 am" src="https://user-images.githubusercontent.com/12241881/27595135-d1243ae4-5b29-11e7-83fa-eef3bab544fd.png">

<img width="1280" alt="screen shot 2017-06-26 at 7 31 02 pm" src="https://user-images.githubusercontent.com/12241881/27594532-0508769c-5b28-11e7-8745-6f9f7269325a.png">
<img width="1280" alt="screen shot 2017-06-26 at 7 31 20 pm" src="https://user-images.githubusercontent.com/12241881/27594535-075f102c-5b28-11e7-81e3-de2999586c99.png">

 
And under the page, typically there are categories like
- Councillor’s profile
- Councillors
- Mayer and Councillors
- Councillors and Wards
- Ellected Councillors

[...]

PR #1195, opened 2017-07-18
Council directory

hisayohorie:
This is a part of the #1182 to make it easier to search council website by creating a directly.

I added council_website column in the authorities table as well as the admin/authorities.rb
It seems like adding authorities to the date base is not an automated process?

Not sure of the next step. I'm guessing I can start making the tutorial page?  I very much appreciate your insight! @equivalentideas @henare

### DIFF
--- a/app/admin/authorities.rb
+++ b/app/admin/authorities.rb
@@ -11,6 +11,7 @@ ActiveAdmin.register Authority do
     column "Name", :full_name
     column :state
     column :email
+    column :council_website
     column(:applications) { |a| a.applications.count }
     column(:total_comments) { |a| a.comments.count }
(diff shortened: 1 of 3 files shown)


henare:
I think this makes sense but I assume @hisayohorie and @equivalentideas discussed this in their pairing session yesterday so I'm going to assign @equivalentideas as reviewer.

>It seems like adding authorities to the date base is not an automated process?

@hisayohorie that's correct, we currently manually add authorities as we create scrapers for them. We want to change this so that there's a page for every authority, whether we have them or not (see #559).

hisayohorie:
made renewed PR #1196  so closing this branch

Issue #1198, opened 2017-07-27
Help people contribute correct councillor email addresses by validating their input


equivalentideas:
We validate people’s email addresses when they sign up for alerts, we can do the same thing to help them contribute good councillor information.

The email field for councillors should not be blank, and should be a valid email address.

Extracted this idea from @hisayohorie comment https://github.com/openaustralia/planningalerts/issues/1182#issuecomment-313275951

hisayohorie:
I'm having a bit of issue here using `validates_email_format_of` gem, which is used for email validation in other models already. 
I have the `validates_email_format_of :email` in suggested controller model, but the validation only works on `councillor_contribution#create`, and contributions with invalid email format still display as suggested councillor list.
This email format is invalid(not following email@address.com format)
<img width="628" alt="screen shot 2017-07-29 at 7 50 31 pm" src="https://user-images.githubusercontent.com/12241881/28749552-2961a178-7499-11e7-830e-d6a3ce6f3503.png">
But still display as suggested councillor list, as it is not yet created
<img width="562" alt="screen shot 2017-07-29 at 7 51 01 pm" src="https://user-images.githubusercontent.com/12241881/28749558-5077dbb0-7499-11e7-8c90-56edf2cc4fb1.png">
And when i finally submit, the validation works
<img width="840" alt="screen shot 2017-07-29 at 8 07 27 pm" src="https://user-images.githubusercontent.com/12241881/28749567-94840aae-7499-11e7-9ee2-d3af4cd155bf.png">

But the validation should work before create to prevent displaying suggested councillor list. 
I'm wondering if this is one of the complication not using #edit and #update method. This is more complicated than I initially thought. any idea? @equivalentideas

equivalentideas:
@hisayohorie good explanation of what's going on 👍 If you've got any existing commits you could push up at "[WIP]" (work in progress) pull request to show how it's working too.

> I'm wondering if this is one of the complication not using #edit and #update method. 

[...]

PR #1206, opened 2017-08-01
Contribution validation email

hisayohorie:
This is a part of #1198 

I added `validates_email_format_of` gem in `suggested_councillor` controller.

As @equivalentideas suggested, I used `empty?` and `valid?` method in the `suggested_councillor_controller#new` to prevent refreshing the form/displaying the invalid contribution under "The councillor(s) you submitted" .

It does the job, but some more work is needed (test, error message, layout)

Try to submit invalid email format (without @email.com format)
screen shot 2017-07-31 at 10 58 15 pm

it does not accept the input, but it breaks the layout because of the error message div.
screen shot 2017-07-31 at 10 58 28 pm
screen shot 2017-07-31 at 10 59 04 pm

if the email format is right, it refresh the form and display the contribution.
screen shot 2017-07-31 at 10 59 28 pm

### DIFF
--- a/app/assets/stylesheets/partials/_form.scss
+++ b/app/assets/stylesheets/partials/_form.scss
@@ -55,6 +55,23 @@ textarea {
   }
 }
 
+// TODO: This form doesn't use the Semantic Form library (from formtastic gem) that our other forms do.
+//       We're actually writing over a lot of the css we get from that library for those forms,
+//       so for this new form we decided it would be less complex to use the vanilla rails form helper.
+//       The CSS from Semantic Form is technical debt at this point, and should be gradually removed as we touch
+//       each of those forms. These new styles could be the base for those styles.
+.councillor-contribution-form {
+  .field_with_errors {
+    input {
+      border-color: $form-error-color;
+    }
+
+    .councillor-contribution-error {
+      color: $form-error-color;
+    }
+  }
+}
+
 textarea {
   line-height: 1.5;
 }
--- a/app/controllers/councillor_contributions_controller.rb
+++ b/app/controllers/councillor_contributions_controller.rb
@@ -10,9 +10,7 @@ class CouncillorContributionsController < ApplicationController
       @councillor_contribution = CouncillorContribution.new
     end
 
-    if @councillor_contribution.suggested_councillors.empty? || @councillor_contribution.suggested_councillors.collect { |c| c.valid? }.exclude?(false)
-      @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
-    end
+    @councillor_contribution.suggested_councillors.build({email: nil, name: nil}) if new_suggested_councillor_required?
   end
 
   def create
@@ -40,5 +38,9 @@ class CouncillorContributionsController < ApplicationController
       unless ENV["CONTRIBUTE_COUNCILLORS_ENABLED"].present?
         render "static/error_404", status: 404
       end
+  end
+
+  def new_suggested_councillor_required?
+    @councillor_contribution.suggested_councillors.empty? || @councillor_contribution.suggested_councillors.collect { |c| c.valid? }.exclude?(false)
   end
 end
--- a/app/controllers/councillor_contributions_controller.rb
+++ b/app/controllers/councillor_contributions_controller.rb
@@ -18,13 +18,14 @@ class CouncillorContributionsController < ApplicationController
     @authority = Authority.find_by_short_name_encoded!(params[:authority_id])
     @councillor_contribution = @authority.councillor_contributions.build(councillor_contribution_params)
 
+    if @councillor_contribution.suggested_councillors.empty?
+      @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+    end
+
     if @councillor_contribution.save
       redirect_to new_contributor_url(councillor_contribution_id: @councillor_contribution.id)
     else
       flash[:error] = "There's a problem with the information you entered. See the messages below and resolve the issue before submitting your councillors."
-      if @councillor_contribution.suggested_councillors.empty?
-        @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
-      end
       render :new
     end
   end
--- a/app/controllers/councillor_contributions_controller.rb
+++ b/app/controllers/councillor_contributions_controller.rb
@@ -9,8 +9,9 @@ class CouncillorContributionsController < ApplicationController
     else
       @councillor_contribution = CouncillorContribution.new
     end
-
-    @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+    if @councillor_contribution.suggested_councillors.empty? || @councillor_contribution.suggested_councillors.collect { |c| c.valid? }.exclude?(false)
+      @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+    end
   end
 
   def create
@@ -20,6 +21,10 @@ class CouncillorContributionsController < ApplicationController
     if @councillor_contribution.save
       redirect_to new_contributor_url(councillor_contribution_id: @councillor_contribution.id)
     else
+      flash[:error] = "There's a problem with the information you entered. See the messages below and resolve the issue before submitting your councillors."
+      if @councillor_contribution.suggested_councillors.empty?
+        @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+      end
       render :new
     end
   end
--- a/app/controllers/councillor_contributions_controller.rb
+++ b/app/controllers/councillor_contributions_controller.rb
@@ -4,11 +4,12 @@ class CouncillorContributionsController < ApplicationController
   def new
     @authority = Authority.find_by_short_name_encoded!(params[:authority_id])
 
-    if params["councillor_contribution"]
-      @councillor_contribution = @authority.councillor_contributions.build(councillor_contribution_params)
-    else
-      @councillor_contribution = CouncillorContribution.new
-    end
+    @councillor_contribution =
+      if params["councillor_contribution"]
+        @authority.councillor_contributions.build(councillor_contribution_params)
+      else
+        CouncillorContribution.new
+      end
 
     @councillor_contribution.suggested_councillors.build({email: nil, name: nil}) if new_suggested_councillor_required?
   end
--- a/app/controllers/councillor_contributions_controller.rb
+++ b/app/controllers/councillor_contributions_controller.rb
@@ -9,6 +9,7 @@ class CouncillorContributionsController < ApplicationController
     else
       @councillor_contribution = CouncillorContribution.new
     end
+
     if @councillor_contribution.suggested_councillors.empty? || @councillor_contribution.suggested_councillors.collect { |c| c.valid? }.exclude?(false)
       @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
     end
--- a/app/models/suggested_councillor.rb
+++ b/app/models/suggested_councillor.rb
@@ -2,4 +2,5 @@ class SuggestedCouncillor < ActiveRecord::Base
   has_one :authority, through: :councillor_contribution
   belongs_to :councillor_contribution
   validates :councillor_contribution, :name, :email, presence: true
+  validates_email_format_of :email, message: "must be a valid email address, e.g. jane@example.com"
 end
--- a/app/views/councillor_contributions/_contribution_form_input.html.haml
+++ b/app/views/councillor_contributions/_contribution_form_input.html.haml
@@ -4,6 +4,12 @@
       .suggested-councillor-input-group
         = suggested_councillor_field.label :name, "Full name", class: "councillor-contribution-label"
         = suggested_councillor_field.text_field :name
+        - if suggested_councillor.errors[:name].any?
+          = suggested_councillor_field.label :name, class: "councillor-contribution-error" do
+            = suggested_councillor.errors.full_messages_for(:name).to_sentence
       .suggested-councillor-input-group
         = suggested_councillor_field.label :email, class: "councillor-contribution-label"
         = suggested_councillor_field.text_field :email
+        - if suggested_councillor.errors[:email].any?
+          = suggested_councillor_field.label :email, class: "councillor-contribution-error" do
+            = suggested_councillor.errors.full_messages_for(:email).to_sentence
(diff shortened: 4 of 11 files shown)


app/models/suggested_councillor.rb
```diff
@@ -1,5 +1,6 @@
 class SuggestedCouncillor < ActiveRecord::Base
   has_one :authority, through: :councillor_contribution
   belongs_to :councillor_contribution
-  validates :councillor_contribution, :name, :email, presence: true
```
equivalentideas: Why was the validation of  :councillor_contribution removed here?

hisayohorie:
I know why the test is breaking, SuggestedCouncillor does not have `authority_id` anymore, there is `councillor_contribution_id` instead. Should I make a separate PR for the change as it is not directly related to this issue? @equivalentideas

app/controllers/councillor_contributions_controller.rb
```diff
@@ -9,8 +9,9 @@ def new
     else
       @councillor_contribution = CouncillorContribution.new
     end
-
-    @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+    if @councillor_contribution.suggested_councillors.empty? || @councillor_contribution.suggested_councillors.last.valid?
```
equivalentideas: This looks like it does the job to me 👍  I think we need an integration test to describe this behavior change.

app/models/suggested_councillor.rb
```diff
@@ -1,5 +1,6 @@
...
+  validates :email, :name, presence: true
+  validates_email_format_of :email
```
equivalentideas: This should be tested on the model with the `validate_email_format_of` [method that the gem provides](https://github.com/validates-email-format-of/validates_email_format_of#usage). Something like:

```
  describe "must have a valid email attribute" do
    it { expect(SuggestedCouncillor.new).to validate_email_format_of :email }
  end
```

henare:
@hisayohorie and I worked on this during pairing today :pear: During the review we discovered that [point 1 in Luke's review](https://github.com/openaustralia/planningalerts/pull/1206#pullrequestreview-53398438) was still outstanding:

>This is quite a tricky form so I think we need:
>1. an integration test to describe this new behavior for when someone inputs a invalid email
>...

So we started to add an integration test. When we did we discovered there were no error messages being displayed to the user - not a great experience! :slightly_smiling_face: 

We dived in and started work to add that but there's still work to be done. I'm going to add these as a review requesting changes.

Hisayo asked to keep working on this so I'll assign to them. They asked me to fix the merge conflict, which I've just done :smile:

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,30 @@
...
+        end
+
+        click_button "Submit"
+      end
+
+      it "displays an error message" do
+        expect(page).to have_content("does not appear to be a valid e-mail address")
+      end
+
+      # FIXME: This test is passing but it's not testing what we want
```
henare: This test is supposed to make sure the entry with an invalid email address isn't added to the list of councillors we're about to submit. The test passes but it's not testing what we want it to test.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,30 @@
...
+      it "does not go into the list of the suggested councillors" do
+        expect(page).to_not have_content("Name: Mila Gilic")
+        expect(page).to_not have_content("Email: mglic.invalid")
+      end
+
```
henare: Remove extra new line.

app/views/councillor_contributions/_contribution_form_input.html.haml
```diff
@@ -7,3 +7,4 @@
       .suggested-councillor-input-group
         = suggested_councillor_field.label :email, class: "councillor-contribution-label"
         = suggested_councillor_field.text_field :email
+      = @councillor_contribution.suggested_councillors.last.errors.full_messages.to_sentence
```
equivalentideas: > This error display is pretty basic and it's not currently prominent and obvious to the contributor that something's gone wrong. At a minimum I think we need to move it somewhere more prominent and semantically appropriate.

One idea might be to have:

1. message at the top of the form like "There's a problem with the information you entered. See the messages below and resolve the issue before submitting your councillors." 
2. "We could also style it with form error styling we use elsewhere in the application." that plus the detailed input error message below the input.

Here's some stuff on accessible form error message html https://www.paciellogroup.com/blog/2016/01/simple-inline-error-message-pattern/ :)

Here's an example of how mailchimp do something like this, I think we can be more helpful and friendly :):

screen shot 2017-06-09 at 9 43 14 am

hisayohorie: I added `flash[:error]` in the controller new action to inform something is not right, and used `Proc.new` to display different error messages for each case of blank email and invalid email format.

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: I added for the case of blank email and invalid email format for 1) display appropriate error message 2)not to display the blank/invalid input

app/models/suggested_councillor.rb
```diff
@@ -2,4 +2,6 @@ class SuggestedCouncillor < ActiveRecord::Base
   has_one :authority, through: :councillor_contribution
   belongs_to :councillor_contribution
   validates :councillor_contribution, :name, :email, presence: true
+  validates_email_format_of :email, :message => "format seems not right.
```
equivalentideas: This should be in the newer Ruby has style `message: "value"`, not `:message => "value"`.

RuboCop should be picking this up. Are you running that before committing @hisayohorie ?

app/models/suggested_councillor.rb
```diff
@@ -2,4 +2,6 @@ class SuggestedCouncillor < ActiveRecord::Base
...
+  Does it follow the format of example@email.com?", :unless => Proc.new{|c| c.email.blank?}
```
equivalentideas: Does this `unless` option mean that this wont be passed as a message if the email is blank? Reading the reason for this on the [commit message](https://github.com/openaustralia/planningalerts/pull/1206/commits/4d6551b9c451e0189e4fd7e0bee72d8ae4f755fb):

> To display different error messages when the format is invalid and when email is blank

I'm interested in the thinking behind this @hisayohorie . What was happening before this change?

This isn't the right place to implement this behavior. It's a good hint that the commit message for this is about changing the 'display'. The model validations should focus on getting valid data to database/model. The display of error messages should be handled in the view and with helper methods (possibly on the model).

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,56 @@
       end
     end
 
+    scenario "on the contribution page" do
```
equivalentideas: Does this feel like a view spec rather than an integration test? https://robots.thoughtbot.com/how-we-test-rails-applications#view-specs

It doesn't really go through a user flow, and the scenario description is confusing to me. It implies that there's other places to make this contribution and doesn't refer to what this is actually testing: whether the name of the authority is on the page. What do you think @hisayohorie ?

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,56 @@
...
+      visit new_authority_councillor_contribution_path(authority.short_name_encoded)
+
+      expect(page).to have_content("Casey City Council")
+    end
+
+    context "when a person submit blank email" do
```
equivalentideas: typo: s/submit/submits/

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,56 @@
...
+        expect(page).to have_content("Email can't be blank")
+      end
+
+      it "does not go into the list of the suggested councillors" do
+        expect(page).to have_no_content("Full name: Mila Gilic")
+        expect(page).to have_no_content("Email:")
+      end
+    end
+
+    context "when a person submit an invalid email" do
```
equivalentideas: I really like the use of `context` and `before` here, makes it really easy to read. Nice one 👍

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -18,6 +18,56 @@
...
+        within ".councillor-contribution-councillors fieldset" do
+          fill_in "Full name", with: "Mila Gilic"
+          fill_in "Email", with: "mgilic.invalid"
+        end
+
+        click_button "Submit"
+      end
+
+      it "displays an error message" do
+        expect(page).to have_content("Email format seems not right. Does it follow the format of example@email.com?")
```
equivalentideas: I think that while it's nice that this error message is kinda friendly saying "seems" and asking 'does it follow?', I recon error messages should be more concrete. In this case we know that the email isn't in a valid format, so we should just tell the person that and help them fix it.

What do you think of something like "Email must be a valid email address, e.g. jane@example.com" @hisayohorie ? That tells them that what they entered is not right, what the requirement is to get it accepted, and keeps the helpful example you added, but human name in it to make it a little more human 🌳 . That's also a bit more succinct.

app/models/suggested_councillor.rb
hisayohorie: I forgot to run the rubocop. fixed them now!

app/models/suggested_councillor.rb
hisayohorie: @equivalentideas before this, it was displaying multiple error messages when the email was submitted blank ("Email cannot be blank" and ""format seems not right. Does it follow the format of example@email.com?").

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: I changed the description of the test and it made it clear that it is about the display logic. moved it to spec/views.

app/models/suggested_councillor.rb
equivalentideas: > My question now is, when I have multpiple validations for one field, how will I know which error message to display

I think for this first pass on validations it's fine to show multiple messages. We can refine the display as a follow up issue and see what the people using it find helpful.

Is that helpful?

spec/views/councillor_contributions/new_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+feature "Contributing new councillors for an authority" do
```
equivalentideas: Nice work moving this over. This spec is still really a feature spec [using Capybarra](https://github.com/teamcapybara/capybara#using-capybara-with-rspec) methods to interact with the app like visit. We just want to test what renders in the view file here when a specific Authority is assigned. We can use the rspec view spec methods for this. Here's the rspec documentation for that https://relishapp.com/rspec/rspec-rails/docs/view-specs/view-spec .

hisayohorie:
@equivalentideas and I made the change in validations. Next step is to make the error messages more humane and look good. can I assign this to you? @equivalentideas 
screen shot 2017-08-08 at 10 16 43 am

equivalentideas:
> can I assign this to you? @equivalentideas

Yes, my pleasure @hisayohorie 🏇

equivalentideas:
I've updated the error [html](https://github.com/hisayohorie/planningalerts/commit/5346a64d2a35f9beefcd44b08204bcb1978919b6) and [styles](), and [fixed a small bug in the validation message text](https://github.com/hisayohorie/planningalerts/commit/09b13d05b5424cba5b74b5e0e2b6d83c7e65bdb7). I also [changed the validation](https://github.com/hisayohorie/planningalerts/commit/fbfc3decf10bd96014ef84f0f6007e5af4043b6a) introduced in https://github.com/hisayohorie/planningalerts/commit/705f1279908ee7f43ff46398f161b75bdca1ed05 to something a bit more flexible. Please have a read through those commits and let me know what you think @hisayohorie 😄 

screen shot 2017-08-10 at 11 28 40 am
screen shot 2017-08-10 at 11 28 49 am
screen shot 2017-08-10 at 11 29 02 am

equivalentideas:
@hisayohorie anything more to do on this to achieve—

> This error display is pretty basic and it's not currently prominent and obvious to the contributor that something's gone wrong. At a minimum I think we need to move it somewhere more prominent and semantically appropriate.

> We could also style it with form error styling we use elsewhere in the application.

from https://github.com/openaustralia/planningalerts/pull/1206#discussion_r131311951

hisayohorie:
@equivalentideas I checked the validation in browser, and it is letting a blank name and a blank email contribution go through. I think what we want to see is this.
screen shot 2017-08-10 at 12 57 26 pm
I added tests to check if all blank contribution is rejected, and for now it fails

equivalentideas:
>  I checked the validation in browser, and it is letting a blank name and a blank email contribution go through. 

@hisayohorie I agree we shouldn't let people submit a blank councillor, it's too confusing. But, at the moment there's no way to remove a councillor if you hit the 'add another button too many times'. If you did that then the only way to submit your councillors would be to add a fake one on the end.

But I think adding the ability to remove councillors from the list is too much for this PR. @hisayohorie could you please open a new issue for that change?

I think we change that last test you added to be [`pending`](https://relishapp.com/rspec/rspec-core/v/3-5/docs/pending-and-skipped-examples/pending-examples) and then do a last check before merging this.

app/assets/stylesheets/partials/_form.scss
```diff
@@ -55,6 +55,19 @@ textarea {
   }
 }
 
+// TODO: Is this wrapping form class necessary?
```
equivalentideas: This TODO I left is a bit vague, I'll try and make the intention clearer.

hisayohorie:
@equivalentideas sounds good! will open a new issue

equivalentideas:
@henare this PR still has requested changes from you https://github.com/openaustralia/planningalerts/pull/1206#issuecomment-320155223 👍 

> So we started to add an integration test. When we did we discovered there were no error messages being displayed to the user - not a great experience! 🙂

> We dived in and started work to add that but there's still work to be done. I'm going to add these as a review requesting changes.

> Hisayo asked to keep working on this so I'll assign to them. They asked me to fix the merge conflict, which I've just done 😄

This has integration tests for messages and them displaying on the page now, so I think these requested changes are done. Don't want to 'dismiss' your review through so could you check that what you were after is done here and clear that ❌ on your review :)

equivalentideas:
I'm resolving the merge conflict.

equivalentideas:
> Don't want to 'dismiss' your review through so could you check that what you were after is done here and clear that ❌ on your review 

Spoke to @henare and he's happy for me to dismiss his review :) Ta!

equivalentideas:
@hisayohorie well done with this, it's really complex stuff. Check out in those last group of commits how I merged this with the multiple input stuff and refactored a bit.

I'm going to merge and deploy this 🚢

PR #1208, opened 2017-08-01
Add form for a person to contribute suggested councillor info

hisayohorie:
This is related to this issue https://github.com/openaustralia/australian_local_councillors_popolo/issues/98#issuecomment-305383236

This PR is the most basic first version of how people contribute councillor info, so that we can load the information to the site.
People can go to this page from each authorities, and put the names and emails of the councillors they want to contribute. After they submit their contribution, optionally they can give us their contact information(name and email), so that we can contact them to thank them.

screen shot 2017-08-01 at 8 45 03 am

The suggested councillors appear on admin page, and admin can copy them in the existing process of adding new councillors.

screen shot 2017-08-01 at 8 53 23 am

When this is deployed, migration needs to run to add new database tables.


app/controllers/contributors_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    @contributor = Contributor.new(contributor_params)
+    @councillor_contribution = CouncillorContribution.find(params[:contributor][:councillor_contribution_id])
+    if @contributor.save
+      @councillor_contribution.update(contributor: @contributor)
+      flash[:notice] = "Thank you"
+      redirect_to root_url
+    end
+  end
+
+private
```
equivalentideas: This should be indented to the same level as the methods it covers below.

app/models/contributor.rb
```diff
@@ -0,0 +1,4 @@
+class Contributor < ActiveRecord::Base
+  has_many :suggested_councillors, through: :councillor_contributions
```
equivalentideas: Is this association necessary? If we're not using it we should remove that line.

app/models/authority.rb
```diff
@@ -21,6 +21,9 @@ class Authority < ActiveRecord::Base
   has_many :applications
   has_many :councillors
   has_many :comments, through: :applications
+  has_many :councillor_contributions
+  has_many :suggested_councillors, through: :councillor_contributions
```
equivalentideas: Is this association necessary? If we're not using it we should remove that line.

app/views/contributors/new.html.haml
```diff
@@ -0,0 +1,20 @@
+%h1 Thank you!
+%p
+  This data you contributed will be reviewd
```
equivalentideas: Should be "reviewed".

app/views/contributors/new.html.haml
```diff
@@ -0,0 +1,20 @@
...
+  by an administrator, and go live shortly.
```
equivalentideas: I think we can improve this text a bit but that's not a problem for this first pass being merged. We should open an issue for that.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  context "when the feature flag is on" do
+    around do |test|
+      with_modified_env CONTRIBUTE_COUNCILLORS_ENABLED: "true" do
+        test.run
+      end
+    end
+
+    scenario "on the contribution page" do
```
equivalentideas: This spec file should either use `scenario` or `it` not both.

equivalentideas: Also the description of this test doesn't really tell you what is being tested :)

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      visit new_authority_councillor_contribution_path(authority.short_name_encoded)
+
+      expect(page).to have_content("Casey City Council")
```
equivalentideas: This might fit better in a view spec or as a step in one of the specs below.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    it "works successfully when the contributor provides their information" do
+      visit new_authority_councillor_contribution_path(authority.short_name_encoded)
+
+      within ".councillor-contribution-councillors fieldset" do
+        fill_in "Full name", with: "Mila Gilic"
+        fill_in "Email", with: "mgilic@casey.vic.gov.au"
+      end
+
+      click_button "Submit"
```
equivalentideas: I think this should use the full button text like the other tests do for consistency.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    it "works successfully without contributor information" do
+      visit new_authority_councillor_contribution_path(authority.short_name_encoded)
+
+      within ".councillor-contribution-councillors fieldset" do
+        fill_in "Full name", with: "Mila Gilic"
+        fill_in "Email", with: "mgilic@casey.vic.gov.au"
+      end
+
+      click_button "Add another councillor"
```
equivalentideas: @hisayohorie and I talked about this, and we recon that this test could be more focused if the person only adds one councillor. That keeps the focus on them not adding there contributor information.

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        fill_in "Name", with: "Jane Contributes"
+        fill_in "Email", with: "jane@contributor.com"
+      end
+
+      click_button "Submit"
+
+      expect(page).to have_content "Thank you"
+    end
+
+    it "works successfully without contributor information" do
```
equivalentideas: To be consistent with the test above, I recon the description here should be something like "works successfully when the contributor does not provide their information"

spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      within ".councillor-contribution-councillors fieldset" do
+        fill_in "Full name", with: "Rosalie Crestani"
+        fill_in "Email", with: "rcrestani@casey.vic.gov.au"
+      end
+
+      click_button "Submit 3 new councillors"
+
+      expect(page).to have_content "Thank you"
+
```
equivalentideas: Stray empty line.

spec/models/contributor_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+RSpec.describe Contributor, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
equivalentideas: Let's remove this file if we're not adding tests yet.

app/models/contributor.rb
hisayohorie: `suggested_councillor_spec.rb`and`user_contributes_new_councillor_spec.rb`pass without this association. Currently we don't have the `councillor_contribution_controller_spec.rb`. so if we get to the stage to have the test, we may need the association? but for now I deleted it.

app/views/contributors/new.html.haml
hisayohorie: fixed

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: Fixed

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: moved in the spec below.

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: fixed

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: fixed

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: I came to realize that this test is to check it `works successfully when the contributor does not provide their information`, and the test is not what it's described. I changed the test to check what it described. I will make a change in the view page to pass the test in separate PR

spec/features/user_contributes_new_councillor_spec.rb
hisayohorie: fixed

spec/models/contributor_spec.rb
hisayohorie: removed

spec/features/user_contributes_new_councillor_spec.rb
equivalentideas: We decided to change this so that all the times this button is hit, the full text is used "Submit 1 new councillor".

spec/features/user_contributes_new_councillor_spec.rb
equivalentideas: @hisayohorie and I decided to try and make these tests more focused by splitting the options for the contributor to tell us their contact information into a new feature spec file https://github.com/openaustralia/planningalerts/pull/1208/commits/74466eab98fde6bce0839258e34c9e24da917ef3

config/routes.rb
```diff
@@ -102,11 +102,17 @@ def matches?(request)
...
+    resources :councillor_contributions, only:[:new, :create]
     collection do
       get :test_feed
     end
   end
 
+  post "/authorities/:authority_id/councillor_contributions/new", to: "councillor_contributions#new"
+
+  resources :contributors, only:[:new, :create]
+    get "/contributors/no_info", to: "contributors#no_contributor_info"
```
equivalentideas: I think this should change to something a little more specific like "contributors/anonymous", but we don't need to do that before deploying.

README.md
```diff
@@ -109,6 +110,17 @@ We set this in the [`.env`](https://github.com/openaustralia/planningalerts/blob
 ```
 COUNCILLORS_ENABLED=true
 ```
+##### Contributing suggested councillors feature
+Similarly, you can toggle the availability of the contributing suggested councillors feature on or off for the entire site with the environment variable `CONTRIBUTE_COUNCILLORS_ENABLED`.
+The feature is globally enabled when the value of `ENV["CONTRIBUTE_COUNCILLORS_ENABLED"]` is `"true"`.
+This flag is useful if you need to turn the feature _off_ globally.
+
+We set this in the [`.env`](https://github.com/openaustralia/planningalerts/blob/master/.env) file in production.  You can control setting in development by creating your own `.env.development` file which includes:
```
equivalentideas: Actually we set it in production in `.env.production`. We should update that.

PR #1214, opened 2017-08-08
Multiple form fields

### LINKED ISSUE #1210: If I make a mistake in one of the councillors I'm contributing, I have to abandon all my work to fix it
I was adding councillors, and I made a typo in one, writing "Tpm" not "Tim". I didn't notice this until I'd requested to add another councillor. There's no way to edit the councillors I've added already.

screen shot 2017-08-04 at 12 03 23 pm

It seems like the only way to correct it is to close the window and start again. That's really frustrating.

hisayohorie:
**  Please merge #1206 before this PR **

This is to fix #1210

I refactored the hidden_fields to text_field so that user can edit their contribution anytime.
I also refactored the form to partial.


equivalentideas:
@hisayohorie nice work, should this be against the `master` branch? Now that we've merged `person_contributes_councillor_info` in there we should actually delete that branch.

This is quite a big change to the way this form works, I think we should update/add an integration test to cover the situation described in #1210 . What do you recon @hisayohorie ?

equivalentideas:
Sorry I accidentally closed this when I deleted the old `person_contributes_councillor_info` branch. Changed the base branch to `master` and reopening.

hisayohorie:
https://github.com/openaustralia/planningalerts/pull/1214/commits/e2b4d0943f5aee6d7fa38f44a028ebb308af99a6

I added this test, but I feel unsure if this is the right implementation. I tried to find better and more explicit way to target the test and couldn't quite come to how. @equivalentideas thoughts?

equivalentideas:
> I added this test, but I feel unsure if this is the right implementation. I tried to find better and more explicit way to target the test and couldn't quite come to how. @equivalentideas thoughts?

That's looking really good @hisayohorie 👍 I recon if the test could include changing a value that was already inputed that would really capture the key point we're aiming for on the issue #1210 . Maybe the test ends up reading "Contributing new councillors for an authority when the feature flag is on successfully with councillors being edited after they're first added" ? Does that help?

hisayohorie:
@equivalentideas Added the test for the changing value in the suggested councillor in the form and to submit ! ^^

equivalentideas:
That did the job @hisayohorie 👍 I saw an issue in the HTML and made some changes to the tests to make them a bit clearer. See:

* 67eccbe Display all the councillor inputs in one block
* 86a6432 Make method for selecting inputs consistent through test
* 422f275 Use a simpler method to select input rather than long class
* 3b2f19f Change the input to match the purpose of the test
* fde5341 Use the same selector to find the same fieldset each time
* f969313 Confirm that the original councillor wasn't saved
* 478a8ce Remove redundant test

What do you think?

hisayohorie:
@equivalentideas that makes so much sense!

equivalentideas:
> @equivalentideas that makes so much sense!

Yay! :) @hisayohorie Is this ready to merge then?

hisayohorie:
@equivalentideas I think so!

equivalentideas:
@hisayohorie merged 🌴 great work, I'll deploy this :shipit:

PR #1226, opened 2017-08-17
added contributor email in admin councillor contribution#show page 

hisayohorie:
This is to fix #1218 

I updated the admin councillor contribution #show page to look like this, so admin can immediately see the contact information of the contributor without going to contributor#show page 

screen shot 2017-08-16 at 11 31 55 pm

There can be some layout polishing. Please let me know what you think @henare @equivalentideas

Issue #1229, opened 2017-08-18
People contributing councillor details that prefer not to provide their contact details shouldn't need to click a link


henare:
Currently if you don't wish to supply your contact information after submitting councillor data you click a link and then you're sent back to the home page with a message that says, "Thank you":

![selection_001](https://user-images.githubusercontent.com/48945/29442850-8ca7e962-8417-11e7-86f5-c5aefecf4efa.png)

![selection_002](https://user-images.githubusercontent.com/48945/29442861-9c7046aa-8417-11e7-8bcd-12d0db22886d.png)

Isn't the step of clicking the link just extra work for the user that they don't really need to do?

If there's no reason for it then I think we should remove it and just include in the page text that people don't need to fill this page in. We currently say, "This field is optional." but what we really mean is that this page is optional. We should say that and make people feel good about not submitting their personal data if they don't want to and that they've got nothing else to do :smile:

hisayohorie:
@henare I think it goes back to the discussion we had here https://github.com/openaustralia/planningalerts/issues/1177#issuecomment-311240817

I think the larger message of the page is to say thank you to the contributor and let them know the impact they are making and what's the next step. So even if we move the optional contributor info form in the first page, I think it is still important to have this page. And if we would to have this page, the contributor has to click something to get to somewhere. 
But if you have ideas, I'm very curious to know if you have other suggestions on how we can thank them on spot other than having the page! 🚀

henare:
>But if you have ideas, I'm very curious to know if you have other suggestions on how we can thank them on spot other than having the page! :rocket:

On our weekly status call today we had quite a few ideas. @equivalentideas is going to try and bring them together into some issues.

equivalentideas:
So what we talked about in our call last week was the idea that this page is a bit confusing because it has a few purposes:

1. thanking the person for their contribution
2. telling them what will happen next with their data
3. explaining why they might like to provide their contact info
4. providing the mechanism for them to input information, or to indicate that they're not into that.

Because of the way people read a page like this, jumping up and down rather than from top to bottom, it makes the whole thing over-all more confusing. In testing this we saw that the person made their contribution, saw the big 'Thank you!', and said 'great that's done'. I then had to prompt them to re-look at the page and consider the second question about adding their contact info.

On our call, we talked about switching things around so the decision about adding your contact info comes before the final big thanks, and moving that question to its own page. The advantages here are:

[...]

Issue #1177, opened 2017-06-20
Councillor info contributor provides their details so the admin can contact them about their contribution and thank them


equivalentideas:
To contact them, we'll need people’s email address. To be able to thank them, recognise their contribution and communicate with them in a nice friendly way, we need their name too.

We should get these details at the end of the process, so they can get straight into adding councillors.

Extracted from https://github.com/openaustralia/planningalerts/issues/1160

hisayohorie:
in relation to #1179 and the design principle, I think the information of the user should be optional. 
Also, for the user contribution flow, I think it is a nicer/less work experience for user if the form of the suggested councillor and their (optional) user info can be submitted all at once with one click(flow 2 in the chart
![user_contribution_flow](https://user-images.githubusercontent.com/12241881/27352991-91890260-55d0-11e7-8700-b55acf0092e2.png)
).

hisayohorie:
So the main issues here are:
1) Does the user information need to be optional or mandatory?
2) Does the submission of data need to be 1 step or 2 steps?

1) Does the user information need to be optional or mandatory?
 mirroring to our principle, I think users need to have control of where and when they give up their information. so I still think it needs to be optional.

For 2), After a bit of research, I started to lean toward for 2 steps submission. 
1 step submission
**pros**
- Less work for users.

**cons**
-  Presenting optional input with mandatory input may send a mixed message

2 steps submission
**pros**
- It is obvious that the user information input is optional by the way it presents itself as a follow-up
- It makes the input more conversational

**cons**
- More work for users

From a bit of research, I am leaning toward 2 steps submissions. It feels to me that there are more pros than cons.
![user_contribution_flow_2](https://user-images.githubusercontent.com/12241881/27399381-8fae5f3e-568a-11e7-8678-8d45cd7c6bfd.png)

henare:
>is this the right direction? @equivalentideas @henare

Hey @hisayohorie, as we've discussed in other channels this is looking good and is all headed in the right direction :+1: 

>Does the submission of data need to be 1 step or 2 steps?

We've all [discussed this a bit in another thread](https://github.com/openaustralia/planningalerts/pull/1186#issuecomment-311117290). I think I might have disrupted things by suggesting we keep it simple on the 1 page for now. Both @hisayohorie and @equivalentideas have said they think 2 step is the way to go so if we want to change this now then I'm up for that.

equivalentideas:
[...]

PR #1232, opened 2017-08-18
Added source_url in councillor contribution table and created a form in councillor_contribution#new

### LINKED ISSUE #1221: Ask contributors to optionally specify a source for their councillor data contributions
This allows admins to check the accuracy of a contribution.

hisayohorie:
** please marge #1226 before this PR**

this is to fix #1221 

So far, I did:
* added `source_url` in councillor contribution model,
* added `source_url` in the strong params in the councillor_contributions controller. 
* created a form field for the source_url in councillor_contribution#new
* added `source_url` in admin/councillor_contribution#show

screen shot 2017-08-18 at 6 45 40 pm
screen shot 2017-08-18 at 6 45 24 pm

However, the admin display is planned to be look like below in #1226 for the case without contributor info
screen shot 2017-08-18 at 5 00 26 pm

And in case with contributor info
screen shot 2017-08-17 at 1 09 22 pm

So I think we should work on merging #1226, and then we can change the admin page. 
@henare @equivalentideas

### DIFF
--- a/app/admin/councillor_contributions.rb
+++ b/app/admin/councillor_contributions.rb
@@ -14,6 +14,8 @@ ActiveAdmin.register CouncillorContribution do
       row :contributor
       row :id
       row :authority
+      row :source_url
+
     end
 
     h3 "Suggested Councillors"
--- a/app/admin/councillor_contributions.rb
+++ b/app/admin/councillor_contributions.rb
@@ -15,7 +15,6 @@ ActiveAdmin.register CouncillorContribution do
       row :id
       row :authority
       row :source_url
-
     end
 
     h3 "Suggested Councillors"
--- a/app/views/councillor_contributions/new.html.haml
+++ b/app/views/councillor_contributions/new.html.haml
@@ -6,8 +6,8 @@
         = render 'contribution_form_input', f: f, suggested_councillor: s
       %button{formaction: new_authority_councillor_contribution_path(@authority.short_name_encoded), class: "button"} Add another councillor
     .councillor-contribtution-source
-      %p Where did you find this information? please include a link if possible
-      = f.label :name, "Source URL"
-      = f.text_field :source_url
+      %p If you found this information online, please provide the URL where we can check it.
+      = f.label :name, "Source"
+      = f.text_area :source
     .councillor-contribution-actions
       = f.submit "Submit #{pluralize(@councillor_contribution.suggested_councillors.length, "new councillor")}", class: "button-action"
--- a/app/views/councillor_contributions/new.html.haml
+++ b/app/views/councillor_contributions/new.html.haml
@@ -5,5 +5,9 @@
       - @councillor_contribution.suggested_councillors.each do |s|
         = render 'contribution_form_input', f: f, suggested_councillor: s
       %button{formaction: new_authority_councillor_contribution_path(@authority.short_name_encoded), class: "button"} Add another councillor
+    .councillor-contribtution-source
+      %p Where did you find this information? please include a link if possible
+      = f.label :name, "Source URL"
+      = f.text_field :source_url
     .councillor-contribution-actions
       = f.submit "Submit #{pluralize(@councillor_contribution.suggested_councillors.length, "new councillor")}", class: "button-action"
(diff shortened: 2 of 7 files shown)


app/admin/councillor_contributions.rb
```diff
@@ -14,6 +14,8 @@
       row :contributor
       row :id
       row :authority
+      row :source_url
+
```
equivalentideas: Stray extra line here :)

app/views/councillor_contributions/new.html.haml
```diff
@@ -5,5 +5,9 @@
       - @councillor_contribution.suggested_councillors.each do |s|
         = render 'contribution_form_input', f: f, suggested_councillor: s
       %button{formaction: new_authority_councillor_contribution_path(@authority.short_name_encoded), class: "button"} Add another councillor
+    .councillor-contribtution-source
+      %p Where did you find this information? please include a link if possible
```
equivalentideas: "please include a link if possible" should be "Please include a link if possible".

But maybe something a little more conversational? "If you found this information online, please provide the URL where we can check it." or something that hints at why we want the data?

app/admin/councillor_contributions.rb
```diff
@@ -14,6 +14,8 @@
       row :contributor
       row :id
       row :authority
+      row :source_url
```
equivalentideas: The source information might not actually be a URL, so I recon we should just call this attribute "source".

db/schema.rb
```diff
@@ -141,8 +141,9 @@
   create_table "councillor_contributions", force: :cascade do |t|
     t.integer  "contributor_id", limit: 4
     t.integer  "authority_id",   limit: 4
-    t.datetime "created_at",               null: false
-    t.datetime "updated_at",               null: false
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+    t.string   "source_url",     limit: 255
```
equivalentideas: If we change this attribute name to just be 'source' we should change it here too.

Also, because people might want to tell a bit of a story here, and we welcome that, we should use the column data type 'text' rather than 'string'. Text is generally for larger blocks and defaults to having a larger limit.

app/views/councillor_contributions/new.html.haml
```diff
@@ -5,5 +5,9 @@
...
+      = f.label :name, "Source URL"
+      = f.text_field :source_url
```
equivalentideas: @hisayohorie Do you recon we should make a bigger text area here to encourage people to add more information if they want? 

Following on from what @henare wrote https://github.com/openaustralia/planningalerts/issues/1221#issuecomment-323242217 I recon that would encourage people a bit more.

> This also allows for input we're not expecting, i.e. not a URL, that might be something like, "I found the information in my local newspaper".

> I believe this permissive design approach is generally a good idea but especially with ideas like this that are at their early stages

If we do want to make a more spacious input for people, like our comment form, we could use `text_area` rather than `text_field` here.

screen shot 2017-08-22 at 12 43 47 pm

hisayohorie:
@equivalentideas I think there is an heated argument about clicking to make the form more conversational or saving the users from optional clicking here #1229 and there https://github.com/openaustralia/planningalerts/issues/1177#issuecomment-311240817 🔥 
Probably it is a good idea for us to be on the same page.

I like the idea of having interactive and conversational form. Although In this case, I think it is ok to keep the source text area in the same page as the councillor_contribution for 2 reasons:
* source of the councillors information are optional but different from asking for user's personal information. So it still feels like it belongs to the same category of the questions on the first page.
* As Pip was doing(and I would do that too), the user are likely copy and pasting from the council website, and it is a lot of flicking of the tab. I think having the source text area in the same page save potential one more flicking.
* If we have the source in the same page as councillor contribution, we can use the existing route of POST. If we decide to have the source in the different page, I would think we use update (PATCH), but not sure if that's what we saved update for.  

Or should we use the same method of having hidden_field in the separate page for the source? what do you think @equivalentideas ?

equivalentideas:
This has been superseded by https://github.com/openaustralia/planningalerts/pull/1247

PR #1241, opened 2017-08-30
Created mailer to notify admin when new councillor contribution is created

### LINKED ISSUE #1217: Administrators not notified of new councillor data contributions
So they know contributions are awaiting their review and merge.

hisayohorie:
This is a first step to fix #1217 

I added a mailer action in the councillor_contribution#create, so the admin will be notified every time new councillor contribution is made.

In `admin_notice_mailer`, I used `example.email.com` for `:from`, as I don't know if there is specific email address the admin wish to be notified from. 

Also the email template is very minimal right now.  Please let me know what you like to see in it. @henare @equivalentideas


app/mailers/admin_notice_mailer.rb
```diff
@@ -0,0 +1,8 @@
+class AdminNoticeMailer < ActionMailer::Base
+  def notice_for_councillor_contribution
+    @admins =  User.where(admin: true)
```
equivalentideas: Rather than going out to all the admins here, I think we can just send this to the central site owners email address, like we do with abuse reports https://github.com/openaustralia/planningalerts/blob/master/app/mailers/report_notifier.rb#L6

app/views/admin_notice_mailer/notice_for_councillor_contribution.html.haml
```diff
@@ -0,0 +1,6 @@
+%p
+  Hi Admin,
+
+%p
+  There is a new submission of councillor contribution. Please check the
```
equivalentideas: I think this should link directly to the contribution admin show view, rather than the index. So it's a bit clearer what's actually come in.

It would also be nice to have a bit of info about it in the email, like:

> Hi there humans,
> 
> You've got a wonderful [new councillor contribution from anonymous for Bayside Council](link).
> You can view and export the contribution in the PlanningAlerts Admin area:
> 
> link
> 
> Don't forget to thank the lovely human who made this contribution!
> 
> Have a great day,
> 
> PlanningAlerts

app/controllers/councillor_contributions_controller.rb
```diff
@@ -23,6 +23,7 @@ def create
     end
 
     if @councillor_contribution.save
+      AdminNoticeMailer.notice_for_councillor_contribution.deliver_later
```
equivalentideas: The other way to do this is with a callback on 'create' for the contribution—but I *much* prefer this approach of directly making the call to email where we want it sent, because it's much easier to follow and test because it's more predictable ✨

app/views/admin_notice_mailer/notice_for_councillor_contribution.html.haml
hisayohorie: @equivalentideas I update the haml as 
>Hi There humans,

>You've got a wonderful new councillor contribution in Marrickville Council. Here is the[link.](link).

>Don't forget to thank the lovely human who made this contribution if you know who they are!

>Have a great day,

>PlanningAlerts

I did not include the name of the contributor, because as we have the `AdminNoticeMailer` under `create` method of the councillor contribution, we don't have the information of the contributor at this point.

hisayohorie:
@equivalentideas I updated the haml and created a test. Ready for your review!

spec/mailers/admin_notice_mailer_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+
+describe AdminNoticeMailer do
```
equivalentideas: These are good tests that cover all the important stuff 👍

I think we should use the `let` pattern, rather than this pattern of assigning instance variables here. Checkout spec/mailers/reply_notifier_spec.rb for an example of this.

equivalentideas:
Thanks for making those changes @hisayohorie ✨ I'll merge and deploy this.

equivalentideas:
This is deployed :shipit:

PR #1246, opened 2017-09-06
Moved methods from contributors_controller to councillor_contributions_controller

### LINKED ISSUE #1229: People contributing councillor details that prefer not to provide their contact details shouldn't need to click a link

hisayohorie:
This is to the first step to fix #1229

Using the test @equivalentideas wrote as a guide, I moved the methods in `contributors_controller` to `councillor_contributions_controller`.

Now the user flow would look like
                                                
                                                                                                                                     
            councillor_contribution#new
                                 |
     Great work! would you like to give us your info?
                    /                               \
           give us info                  not give us info
                     \                         /
                   Thank you for this great contribution of 
                   #{number of suggested councillors} in #{the authority name}

screen shot 2017-09-06 at 4 19 15 pm
screen shot 2017-09-06 at 4 19 25 pm

either button you click, the user will land on this page:
screen shot 2017-09-06 at 4 19 48 pm

thank you page is bare bone just to pass the test at this moment, and need more polishing to fit our design principle!


equivalentideas:
@hisayohorie and I had a big pairing session this morning working through the process to make this flow like:

1.  #new /authorities/foo_council/councillor_contributions/new
2. #add_contributor /authorities/foo_council/councillor_contributions/add_contributor
3. #create /councillor_contributions

It got quick late and so we stopped half way through a big change, getting a bit stuck on how the strong params work on the #create step. The Contribution's suggested councillors weren't getting through our permitted params (or maybe the params hash isn't being constructed correctly from what's coming through from the view?—something to check). Here's the docs on `.permit`: http://api.rubyonrails.org/v4.2.7.1/classes/ActionController/Parameters.html#method-i-permit

equivalentideas:
Fantastic small commits @hisayohorie ! This is looking real good 💯 Just those two test failures to debug, and some clean up some of the old contributor controller code (we should remove everything that isn't getting used now).

equivalentideas:
In chat, @hisayohorie also raised an important architecture point:

>  it came to my mind, if we want to wait to save suggessted councillors till contributor input page for users to click “submit” or “no thanks” of their personal info, I worry that users doesn’t even bother clicking either buttoms and close the window, then nothing is ever get saved?

So in the situation where someone builds up a contribution, they get to the page to add their details, and then something else comes up and they forget to come back (e.g. they abandon the process).

We'd lose the whole contribution.

Could one way to avoid this to actually save the contribution to the database rather than just validate it in the #add_contributor action? Then we also don't need to pass all the suggested councillor details through the #add_contributor template, and can just pass through the id of the Councillor Contribution. If we do that we should tighten up which parameters we permit so we just permit what we expect at each stage.

app/controllers/councillor_contributions_controller.rb
```diff
@@ -31,7 +31,7 @@ def add_contributor
     end
   end
 
-  def create
+  def update
```
equivalentideas: @hisayohorie This isn't really a conventional #update action, but I agree that it's not a conventional #create action now either. Why not make this the #thank_you action and make it accept the data posted from the form in the #add_contributor form?

hisayohorie:
@equivalentideas I changed the method `update` to `thank_you` as well as route and view files accordingly.

the other change I made is that I removed the hidden_field of the suggested_councillors from add_contributor form. To do that I changed the way we retrive the councillor contribution from the database(since it is saved at the point) instead of building it new.

equivalentideas:
@hisayohorie works great! I made a few small changes to make it a little tighter and got a little bit carried away late at night fiddling with the page content :)

I changed it so that the contribution id is passed through the page rather than the url, which feels cleaner (and the urls look cleaner).

Testing this out I found a bug where the flash message persists and fixed it, see https://github.com/openaustralia/planningalerts/pull/1246/commits/4c81341bb451ec0ad5579304f726d2c71a21b664

I also made it so that if you put stuff in the form and then hit 'No thanks' the content you entered is discarded https://github.com/openaustralia/planningalerts/pull/1246/commits/3cecb5cee6f79085c9f636025c44fcb668cdd507

Merging! :shipit:

PR #1247, opened 2017-09-11
Created a source input flow in the councillor contribution

### LINKED ISSUE #1221: Ask contributors to optionally specify a source for their councillor data contributions

hisayohorie:
This is to fix #1221 

In this PR, I created
*  a method `source` in the councillor contribution controller
*  a route `post /authorities/:authority_id/councillor_contributions/source`
*  a view file for source input form
*  updated the feature test to incorporate the flow

Now the flow looks like this on browser:
<br>
submit the suggested councillor info
<br>
screen shot 2017-09-11 at 1 19 30 pm
<br>
land on 
screen shot 2017-09-11 at 1 20 05 pm
source information page
<br>
To incorporate diverse data format, it is a text field 
screen shot 2017-09-11 at 1 22 01 pm
<br>
and land on the contributor info input page
screen shot 2017-09-11 at 1 22 12 pm

Please let me know what you think @equivalentideas!


spec/features/user_contributes_new_councillor_spec.rb
```diff
@@ -59,15 +75,15 @@
           expect(CouncillorContribution.first.contributor.name).to eq "Jane Contributes"
         end
       end
+        context "and skiping contributor details" do
```
equivalentideas: Is this indentation correct?

app/controllers/councillor_contributions_controller.rb
```diff
@@ -17,6 +17,23 @@ def new
...
+
+    @councillor_contribution = @authority.councillor_contributions.build(
+      councillor_contribution_with_suggested_councillors_params
+    )
+
+    if @councillor_contribution.suggested_councillors.empty?
+      @councillor_contribution.suggested_councillors.build({email: nil, name: nil})
+    end
+
+    unless @councillor_contribution.valid?
```
equivalentideas: Given our discussion here @hisayohorie https://github.com/openaustralia/planningalerts/pull/1246#issuecomment-327966708 should we be saving the contribution at this point instead of just validating, in case the person has to abandon ship at this point?

hisayohorie: @equivalentideas I gave some thoughts about it and now I'm not sure. Maybe because the way I set it up it feels like the submission is in the middle.

equivalentideas:
I've pushed some styling for this new page in the form and made the 'add_contributor' page consistent with it as well.

screen shot 2017-09-18 at 3 54 12 pm
screen shot 2017-09-18 at 3 54 19 pm

app/controllers/councillor_contributions_controller.rb
equivalentideas: > @equivalentideas I gave some thoughts about it and now I'm not sure. Maybe because the way I set it up it feels like the submission is in the middle.

@hisayohorie and I spoke about this on a call last week and decided that the safest option at the moment is to save the contribution on when the councillors are submitted, before adding the source and contributor.

There are many reasons why someone might halt the process before hitting 'submit' on the 'add source' page (e.g. they get distracted by kids, laptop runs out of battery) and we wouldn't want to have wasted their time by throwing away all their work.

I'll implement that change now 😄

equivalentideas:
I've made that [change in the spot where the contribution is saved](https://github.com/openaustralia/planningalerts/pull/1247#discussion_r138784445). All great otherwise, so merging!

equivalentideas:
Deployed :)

PR #1248, opened 2017-09-12
Created boolean column of :reviewed in admin/councillor_contribution.rb

### LINKED ISSUE #1240: When an admin views the list of councillor contributions they don't know which ones have been processed already or need to be reviewed
Looking at the admin councillor contributions index page, it's not clear which contributions need attention.

screen shot 2017-08-30 at 2 40 59 pm

hisayohorie:
This is to fix #1240

In this PR I created
- a column of  `t.boolean, reviewed default: false` in councillor_contributions table
- `:edit` and `update` action in `admin/councillor_contributions`
- added strong params in `admin/councillor_contributions` to enable the edit & update action

admin/councillor_contributions/index

screen shot 2017-09-12 at 5 17 14 pm

admin/councillor_contribution/show

screen shot 2017-09-12 at 5 20 15 pm

admin/councillor_contributions/:id/edit

screen shot 2017-09-12 at 5 18 12 pm

when you hit "Update councillor contribution" , you land on admin/councillor_contribution/show page with flash notice
screen shot 2017-09-12 at 5 18 22 pm

No on admin/councillor_contributions/index, we can see it which contribution is reviewed 

screen shot 2017-09-12 at 5 18 32 pm

Please let me know what you think @equivalentideas !


hisayohorie:
Also I added tests under `feature/admin_views_councillor_contribution_spec.rb`

db/schema.rb
```diff
@@ -142,8 +142,10 @@
   create_table "councillor_contributions", force: :cascade do |t|
     t.integer  "contributor_id", limit: 4
     t.integer  "authority_id",   limit: 4
-    t.datetime "created_at",               null: false
-    t.datetime "updated_at",               null: false
+    t.datetime "created_at",                                   null: false
+    t.datetime "updated_at",                                   null: false
+    t.text     "source",         limit: 65535
```
equivalentideas: This 'source' column is isn't in the migration on this PR, I think that's leaked in from another PR—got to be careful when running migrations and working across multiple branches to rollback any unrelated migrations before committing.

app/admin/councillor_contributions.rb
```diff
@@ -21,7 +22,25 @@
     end
   end
 
-  action_item :download,:only => [:show] do
+  form title: "edit" do |f|
+    inputs "Councillor contribution" do
+      input :contributor
```
equivalentideas: Given that we have no need to edit the contributor or source at the moment, I think we shouldn't have these attributes available to edit.

Would another way to do this be to have just an action button on the admin page that says 'mark as reviewed' if it's not reviewed, and 'mark as not reviewed' if it's been reviewed so the admin can toggle it back if they need (I can imaging that being useful) ? Would that be simpler?

spec/features/admin_views_councillor_contribution_spec.rb
```diff
@@ -58,4 +58,45 @@
     expect(page).to have_content "Susan Serey sserey@casey.vic.gov.au"
     expect(page).to have_content "Rosalie Crestani rcrestani@casey.vic.gov.au"
   end
+
+  it "with reviewed column default false" do
+    sign_in_as_admin
+
+    click_link "Councillor Contributions"
```
equivalentideas: I think we can extract the:

```
sign_in_as_admin

click_link "Councillor Contributions"
```

—to a shared setup now.

```
before :each do
  sign_in_as_admin

  click_link "Councillor Contributions"
end
```

equivalentideas:
@hisayohorie Looks like there's merge conflicts since merging another PR. Could you have a squiz at that and resolve them? Try using the GitHub 'resolve conflicts' button below, or merge master into this branch local to deal with it there.

This is what this currently looks like for admins:

screen shot 2017-09-18 at 5 48 11 pm

I think it does the job functionality wise, but it's not very obvious, looks a bit messy on the page and it requests a few steps. Check 'reviewed' and then hit 'update'. It's a bit confusing conceptually too, cause as an admin I'm not seeking to 'update' the contribution, just let people know I've reviewed it.

Another way to do this would be to have a button like our download button that said 'Mark as reviewed' or 'Mark as not reviewed'. That would be one click and the status would be clear by the button text. That also would follow the pattern we use with other admin pages where all the main actions are buttons like that, e.g. loading councillors to authority etc. . Something like:

screen shot 2017-09-18 at 5 50 54 pm

I recon that might be a bit simpler and clearer.

What do you think about that approach @hisayohorie ?

hisayohorie:
@equivalentideas The toggle button looks pretty sleek! I looked around similar solutions, and it seems to require javascript. Maybe you don't want javascript for additional dependancy, also it is not likely admins looking at the page from somewhere javascript is disabled? So I think I'll try implement javascript version.

app/admin/councillor_contributions.rb
```diff
@@ -15,14 +16,18 @@
...
+    column :name
+    column :email
   end
+end
 
-  action_item :download,:only => [:show] do
+  action_item :download, only: [:show] do
     link_to "Download the suggested councillors CSV", authority_councillor_contribution_path(resource.authority, resource.id, format: :csv)
   end
+  action_item :mark_as_reviewed, only: [:show] do
```
equivalentideas: While this is a nice straight-forward implementation @hisayohorie I think we should follow the pattern used in admin/authorities, and use the Active Admin DSL for adding controller actions. That way we save adding a route, and controller action for this small feature.

Checkout https://github.com/openaustralia/planningalerts/blob/master/app/admin/authorities.rb#L88-L98 and https://activeadmin.info/8-custom-actions.html .

It's a bit magical but basically use declare the controller action as a `member_action` in the actual admin file.

And then rather than having a link you get `button_to` which is more semantic. Active Admin creates the appropriate route by itself, because https://github.com/openaustralia/planningalerts/blob/master/config/routes.rb#L26 .

app/controllers/councillor_contributions_controller.rb
```diff
@@ -69,6 +69,16 @@ def show
     end
   end
 
+  def mark_as_reviewed
+    @councillor_contribution = CouncillorContribution.find(params[:id])
+    if @councillor_contribution.reviewed?
```
equivalentideas: This logic can be replaced by [the handy `toggle` method](http://api.rubyonrails.org/v4.2.7.1/classes/ActiveRecord/Persistence.html#method-i-toggle).

```
@councillor_contribution.toggle('reviewed')
@councillor_contribution.save!
```

app/controllers/councillor_contributions_controller.rb
```diff
@@ -69,6 +69,16 @@ def show
     end
   end
 
+  def mark_as_reviewed
```
equivalentideas: This action should actually be called 'toggle_reviewed' or something, because it actually does the opposite of the current label sometimes.

equivalentideas:
Nice work @hisayohorie looks good and works a treat ✨ 

Once you get tests parsing you should do a careful review of your new code to see if there's any refactors or bits that can be simplified. I removed some test setup, and other stuff that wasn't being used and fixed some indentation that was off.

PR #1265, opened 2017-10-17
Add instructions on how to add a new authority to PlanningAlerts

henare:


LoveMyData:
Is it worth while to add a template message into the doc like "Thank you for @xxx contribution, we are now covering @council with over xx,xxx people are benefit from it." ?

henare:
@LoveMyData has reviewed this and we've incorporated that feedback. He's also done a more-or-less complete run-through and it seems to be working so I'll :shipit:

PR #1327, opened 2019-01-08
Seperate web/email API keys

jamezpolley:
Google's API dashboard lets us break down usage by API key, but not by anything else that helps us track where the usage is coming from.

The intent here is to allow us to differentiate between straight web traffic (which will use the new key) and links in alert emails (which will continue to use the old key).

Why this way around? Because lots of emails have already been sent out with the old key. We can't change those easily; but we *can* easily change what we display on the website. Once this change is live we can be fairly confident that all web traffic will switch to the new key immediately, while anything using the old credential will be links from emails.

This change connects to openaustralia/infrastructure/pull/119 (although due to the default values provided here, this change should be fine to push without the matching change in infrastructure).


jamezpolley:
okay; tests are passing (yay). I'll push this live in the morning

app/helpers/applications_helper.rb
```diff
@@ -82,49 +83,52 @@ def authority_applications_json_url_for_current_user(authority)
...
-  def google_static_streetview(application, size: "350x200", fov: 90)
-    image_tag(google_static_streetview_url(application, size: size, fov: fov), size: size, alt: "Streetview of #{application.address}")
+  def google_static_streetview(application, size: "350x200", fov: 90, key: "GOOGLE_MAPS_API_KEY")
+    image_tag(google_static_streetview_url(application, size: size, fov: fov, key: key), size: size, alt: "Streetview of #{application.address}")
   end
 
   private
 
   def google_signed_url(domain, path, query)
+    key = query.delete(:key) { "GOOGLE_MAPS_API_KEY" }
```
mlandauer: The `key` parameter is being used quite differently than the rest of the `query` parameters. I think this is a bit confusing. It would be clearer if the key was a separate parameter to the function rather than being mixed in with `query`.

jamezpolley: .env updated.

PR #1347, opened 2019-02-11
Add beginning of bare bootstrap theme

mlandauer:
This adds a theming engine where admins can switch back and forth between the current theme (standard) and a new experimental theme (bootstrap).

The bootstrap theme only currently implements the landing page. Furthermore it is completely bare bones html with no styling (other than that which bootstrap provides out of the box).

This PR is intended as the basis of further work to move the front end over to bootstrap (and redesign and simplify in the process) just as we did with theyvoteforyou.

By having a theming engine we can safely deploy this to production in a half-baked state and be fairly confident that we're not breaking other things while avoiding the merge nightmare that would be caused by a very long-running branch.


mlandauer:
I deployed this to production. It appears to have somehow broken the admin login. So, rolling back.

mlandauer:
All fixed again.

Issue #1351, opened 2019-02-20
Record in user table whether user is a commercial api user (and log to elasticsearch)


mlandauer:
We can then use this to do our rate limiting rather than the way we do it currently which is more work on our part.

Also, if we record the new user data in elasticsearch with api requests we can more easily find api usage that is breaking our terms of use.

mlandauer:
@jamezpolley @katska @kat Is this the most recent document with all the current commercial customers? https://docs.google.com/spreadsheets/d/1yEDp2nsXxo9f8_yUTwk1H4Dm3aYUNkJvGHrvCAB5b8U/

mlandauer:
Maybe what we should do instead is record when we know that a user is commercial (they're paying), when we know that it's non-commercial use and when we don't know.

mlandauer:
Closed by #1358

PR #1358, opened 2019-03-05
Api throttling by key

### LINKED ISSUE #1354: Rejig API throttling to use api key rather than ip address
Currently we're doing our throttling using the `rack-throttle` gem based on originating ip address. This means that for commercial users we have to increase their quotas by adding their server ip addresses. This is obviously a hassle for all concerned and definitely a problem if they don't have fixed ip addresses. Furthermore we have no way to limit usage based on user. So, for instance a user could increase their overall quota by spreading out the load among many servers.

So, let's switch over to something way more sensible. Let's do the throttling based on the api key.

Then, we can also handle commercial users and non-commercial & low-volume users differently automatically. I propose that we also simplify the throttling - we just move to a simple two-tier setup. By default people can do 1000 api requests per day. Commercial users (and high volume non-commercial users that we setup) simply have unlimited api calls. This goes along with the move to standardised pricing for commercial users.

So, I propose:
* Add an "unlimited_api_usage" field to the user which by default is false (This replaces the idea in #1351)
* Log this new field to elasticsearch too with the api requests
* If the field is false throttle to 1000 api requests per day
* If the field is false also throttle to 100 api requests per hour

mlandauer:
This fixes #1354 - note however that it doesn't implement the hourly throttling. It just turned out it was a little bit messy. There was more value in getting this pushed out to production quickly without that feature.


mlandauer:
This also closes #1351 as we now record in the database whether api limits are imposed on the api user.

PR #1380, opened 2019-05-07
Document the overall goals of the project as an aid to planning

jamezpolley:
While thinking about changes to the comment form, I was finding it hard to balance different options. In order to help figure out what was most important, I started thinking about the overall goals of PlanningAlerts.

I figured this would be worth adding to the docs as it would help us with future decisions as well.

I've started with documenting what I think are the three primary goals, and now I'm expanding on those.

I don't know if what I've come up with as a list of what I feel are the primary goals is going to match what anyone else thinks - but I figure that time spent coming up with a consensus on the primary  goals will be repaid in time we don't have to spend on minor decisions later

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,25 +12,41 @@ PlanningAlerts is brought to you by the [OpenAustralia Foundation](http://www.op
 
 ## Table of Contents
 
-* [Development](#development)
-  * [Scraping and sending emails in development](#scraping-and-sending-emails-in-development)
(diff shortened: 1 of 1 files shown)


jamezpolley:
* We can only work with the data councils provide
* If residents want better data from us they should talk to their council about providing better data

jamezpolley:
Replaced by 028ace67

PR #1393, opened 2019-06-10
Switch donation form to donorbox

jamezpolley:


mlandauer:
@james did you give up on this approach in favour of #1394? If so, could you please close this PR and explain what the issue was. Thanks!

jamezpolley:
I haven't made a decision yet - I've discussed the options a bit more at https://github.com/openaustralia/oaf/issues/276#issuecomment-500355626

jamezpolley:
But I think #1394 is going to be the way forward for now, so I'll go ahead and close this off for clarity. It's trivial to re-open if things change.

jamezpolley:
The problems I ran into were around the size of the form - the old form is very small, the new one is longer than the background image. It extends down into the green "Because of backers like you, here’s our week on PlanningAlerts" section. In my testing, when the screen gets narrow (eg, on mobiles), it drops *under* that section, which hides the bottom part of the form.

To use this we'd have to change the layout to adapt to the longer form, and I'm not sure how to do that.

Alternatively, we could use the popup form from donorbox, which would mean a "click here to donate" button within the area of the old form, which would trigger a floating form that is always on top of other content.

PR #1401, opened 2019-07-12
Don't promise that an application will be considered.

### LINKED ISSUE #1400: More accurately describe what PlanningAlerts can do with a comment when we don't know a notification period
### Summary

Improve commenter and authority experience by not making the commenter expect that a comment *will* be reviewed

````diff
- Your comment and details will be sent to Newcastle City Council. They’ll consider your submission when
- they decide whether to approve this application. Your name and comment will be posted publicly above.
+ Your comment and details will be sent to Newcastle City Council, and may be considered by them as
+ part of the decision making process for this application. Your name and comment will be posted publicly above
````

### Details

For planning applications where there is no known notification period, the message we display isn't accurate. For instance, looking at https://www.planningalerts.org.au/applications/1295427, tell commenters:

> Your comment and details will be sent to Newcastle City Council. They’ll consider your submission when they decide whether to approve this application. Your name and comment will be posted publicly above.

"They'll consider your submission" isn't always accurate:
 - Some councils list Complying Developments in their feeds, which never have any decision made and never list feedback
 - Some applications will never get to a decision stage
 - It's possible that a decision has already been made

Setting the commenter's expectation to be that the council *will* consider their comment sets them and the council up for an unpleasant experience in cases where that's not accurate.

As well as being a bad experience for the commenter, this is a bad experience for the council too: we've had feedback from councils that this leads to commenters being angry at the council for not considering their response after we had told them it would be considered.

I'd like suggest that we could more accurately set the commenter's expectation - without being alarming, bureaucratic, or confusing, with a small change along the lines of:

> Your comment and details will be sent to Newcastle City Council, and may be considered by them as part of the decision making process for this application. Your name and comment will be posted publicly above.

### PR DESCRIPTION
**Pull request by jamezpolley**
Fixes #1400

Front logo [Front conversations](https://app.frontapp.com/open/top_11iyt)

### DIFF
--- a/app/themes/standard/views/comments/_comment_form_introduction.html.haml
+++ b/app/themes/standard/views/comments/_comment_form_introduction.html.haml
@@ -7,15 +7,7 @@
     or one of your elected local councillors.
   - else
     Your comment and details will be sent to
-<<<<<<< HEAD
-    %strong #{application.authority.full_name}, 
-    - if application.official_submission_period_expired?
-      but cannot be formally considered by them as the notification period for this application has expired.
-    - else
-      and may be considered by them as part of the decision making process for this application.
-=======
     %strong #{application.authority.full_name}.
     - unless application.official_submission_period_expired?
       They may consider your submission when they decide whether to approve this application.
->>>>>>> 40b68ee6... Simply change wording from "They\'ll" to "They may"
   Your&nbsp;name and comment will be posted publicly above.
--- a/app/themes/standard/views/comments/_comment_form_introduction.html.haml
+++ b/app/themes/standard/views/comments/_comment_form_introduction.html.haml
@@ -7,7 +7,9 @@
     or one of your elected local councillors.
   - else
     Your comment and details will be sent to
-    %strong #{application.authority.full_name}.
-    - unless application.official_submission_period_expired?
-      They’ll consider your submission when they decide whether to approve this application.
+    %strong #{application.authority.full_name},
+    - if application.official_submission_period_expired?
+      but cannot be formally considered by them as the notification period for this application has expired.
+    - else
+      for them to consider when they decide whether to approve this application.
   Your&nbsp;name and comment will be posted publicly above.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mlandauer** [comment]
@jamezpolley I'm going to take at the wording now. In the meantime could you please fix up this PR so the tests pass. Thanks!

File: app/themes/standard/views/comments/_comment_form_introduction.html.haml [hunk 1]
```diff
@@ -7,7 +7,9 @@
     or one of your elected local councillors.
   - else
     Your comment and details will be sent to
-    %strong #{application.authority.full_name}.
-    - unless application.official_submission_period_expired?
-      They’ll consider your submission when they decide whether to approve this application.
+    %strong #{application.authority.full_name}, 
+    - if application.official_submission_period_expired?
+      but cannot be formally considered by them as the notification period for this application has expired.
```
mlandauer: According to https://github.com/openaustralia/planningalerts/blob/master/spec/helpers/applications_helper_spec.rb#L160 the `on_notice_text` helper will already say "...but it will not be officially considered by the planning authority.". So, this bit seems on the face of it to be unnecessary.

app/themes/standard/views/comments/_comment_form_introduction.html.haml
```diff
@@ -7,7 +7,9 @@
...
+    - else
+       and may be considered by them as part of the decision making process for this application.
```
mlandauer: Why not simply replace "They'll consider" with "They may consider"? I don't like the use "decision making process" because it's passive bureaucratic language. The authority is deciding whether to approve the application. That's more active language.

jamezpolley:
I believe the build errors are fixed... but Travis is having some temporary issues with the mysql-5.7 repo, so it can't build. I'm going to leave this until this afternoon or tomorrow and try travis again

jamezpolley:
@mlandauer Tests are passing now :)

PR #1422, opened 2020-02-20
Add "Be nice" and standard agreements to terms of use

jamezpolley:


jamezpolley:
I don't know if this works, but I'm merging to master because that's the only way I can deploy to test.

PR #1927, opened 2025-07-07
Add vendor/bundle to .gitignore

Br3nda:


CloCkWeRX:
Oh, hello here too :) if your head is in property/planning/etc space would love to compare notes across AU/NZ models some time!

PR #1944, opened 2025-10-07
Add codeowners

Br3nda:


jamezpolley (changes requested):
Looking at that team, I think it's too broad - it includes lots of people who have been connected to PA in various ways but not as developers. 

I've just created team-planningalerts-dev  which is a tighter group and probably more appropriate for this

.github/CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# This file is used to define code owners for this repository.
+# See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+* @openaustralia/team-planningalerts
```
Br3nda, line 4:
```suggestion
* @openaustralia/staff-devs

```

PR #1945, opened 2025-10-07
Add dependabot config

Br3nda:


jamezpolley (approved):
Sure. Dependabot has usually annoyed me because we've never had time to work on the issues it creates. But we should, and having it running on this repo too won't make things any worse.

Br3nda:
i'd like to get percy working again too

PR #1968, opened 2025-11-27
upgrade webrick

Br3nda:


ianheggie-oaf:
@Br3nda - do you want to close this, or is there something still worthwhile in it?
(With a title and no description its not selling itself as a keeper even in draft status)

PR #1978, opened 2025-12-05
this pr contains the changes we rolled back

Br3nda:
Last week we rollbacked back recent changes, while diagnosing errors. 

The cause was a difference between rbenv and rvm on the server.

This PR puts back the changes we rolled back.


jamezpolley (approved):
Changes have already been pushed and appeared to work - but were rolled back as a precaution when we encountered issues. We've confirmed that the issues were unrelated to this change, so rolling this forward again makes sense.

Issue #1982, opened 2025-12-17
Fix issues found whilst setting up development environment


ianheggie-oaf:
**Description**

Fix issues found whilst setting up development environment

* Add "Resetting Your Environment" README.md section
* Add "Resetting Your Environment" README.md section 
* Update docker-compose.yml to make web container wait till the other services where up and running, not just started to fix rake db:setup crashing if web container was already built
* Enhance db:seed so production DB is protected, it makes admin user an actual admin user you can login to, and makes db:seed repeatable (idempotent)

**Additional context**

It's what I had to do to get the development environment up for myself, so I documented it.

I usually only use docker in a production rather than development environment, so I regard myself as a good test case for whats needed to help other developers new to the application.

Unsure if I should add in these bash aliases I have found useful in the past (I only used dcbe this time):
```
alias dcbe='docker compose exec web bundle exec'
alias dcd='docker compose down'
alias dce='docker compose exec'
alias dcr='docker compose exec web bundle exec rails'
alias dcrk='docker compose run --rm web bundle exec rake'
alias dcsh='docker compose run --rm web bash'
alias dcu='docker compose up --build'
``` 

*Work done as a donation to OAF*

PR #1983, opened 2025-12-17
Chore/update development setup and doc

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/planningalerts/issues/1982

## What does this do?

* Adds further explanation how to use docker in README.md
* Adds health checks so web container waits for database services to be available, not just start to fix db:setup failing if postgres was still coming up

## Why was this needed?

I was investigating the development environment out of curiosity for the first time

## Notes to reviewer (Optional)

Removed changes to db/seed.rb as it broken the web browser checks. My intent was: Enhance and protect db:seed rake task
    
* Protect production from being seeded with development data
* Make db:seed repeatable (checks if records already exist)
* Confirms admin user so its immediately usable (can be done by clicking link in email)
* Add all roles to admin user (not sure how this is done normally for setting up dev)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,32 @@ git clone https://github.com/openaustralia/planningalerts.git
 cd planningalerts
 ```
 
+### Resetting Your Environment
+
+To force a rebuild from scratch, stopping and removing containers and volumes (including the databases):
+
+```sh
+docker compose down -v --rmi all
+# Check if this volume is "still in use" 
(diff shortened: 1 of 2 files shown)


benrfairless:
@ianheggie-oaf tests are now passing. Did you want this reviewed?

benrfairless (approved):

PR #1989, opened 2026-02-05
Add Authority json API call

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/planningalerts/issues/1996 which is needed before the following is done
  * https://github.com/openaustralia/infrastructure/issues/312 
* No longer needed for: https://github.com/planningalerts-scrapers/issues/issues/1075 - dropped that part of the design and using a hard coded list instead

## What does this do?

Adds an API to list authorities, eg: `http://localhost:3000/authorities.json?key=mykey` returns
```json
[
  {
    "authority": {
      "id": 1,
      "state": "NSW",
      "full_name": "Marrickville Council",
      "short_name": "Marrickville",
      "slug": "marrickville",
      "morph_name": "planningalerts-scrapers/marrickville",
      "scraper_authority_label": null,
      "population_2021": 81489,
      "broken": false
    }
  }
]
```

## Why was this needed?

1. nsw_planning_portal scraper will use this list to 
  a. avoid duplicate entries from existing scrapers and 
  b. list records it finds vs the current scrapers so we can make an informed choice to swap over;
2. We will be putting cloudflare in front of planning alerts so scraping the index page will cease being available
3. This will make the transition simple - when we want to swap an authority from the existing scraper to nsw_planning_portal, we will disable the authority in PA, the nsw_planning_portal scraper will grab a fresh copy of the authorities list and will cease excluding that authority. PA will then pick up the applications from nsw_planning_portal
4. My Plannies mate sitte will also need this

## Implementation/Deploy Steps (Optional)

Standard capistrano deploy

## Notes to reviewer (Optional)

slug is authority.short_name_encoded as used in pa's html and api paths for the authority


app/controllers/api_controller.rb
```diff
@@ -14,16 +14,27 @@ class ApiController < ApplicationController
   # GET requests for JSONP instead of using XHR
   # TODO: Remove this line to re-enable CSRF protection on API actions
   skip_before_action :verify_authenticity_token,
-                     only: %i[authority suburb_postcode point area date_scraped all]
+                     only: %i[authorities authority suburb_postcode point area date_scraped all]
```
Br3nda: 👍

spec/controllers/api_controller_spec.rb
```diff
@@ -59,6 +59,34 @@
 
   let(:key) { create(:api_key) }
 
+  describe "#authorities" do
+    it_behaves_like "an authenticated API" do
```
ianheggie-oaf: > @Br3nda said: i see this adds a new route to the api. Is there any new checks that the user is allowed to to this call?

@Br3nda This spec checks user authentication works on authorities endpoint

ianheggie-oaf:
@benrfairless / @Br3nda - Is there anything holding this up from being accepted?

Br3nda (approved):
approved.

PR #2032, opened 2026-04-17
Increase default trial daily limit from 1 to 10

### LINKED ISSUE #2030: New API Trials are currently set to "1" call per day
**Describe the bug**
If you request a trial API key, the default API rate limit is set to "1".

**To Reproduce**
1. Create an account
2. Request a Trial Key
3. Check the admin portal to see what the API limit should be

**Expected behavior**
The user should have enough API calls to test the service, but not enough to abuse the service.

**Screenshots/Output**
N/A at this stage

**Additional context**
This was changed in [this commit](https://github.com/openaustralia/planningalerts/commit/1c17a005c02a4359148057fcd9b0b95676752ca8). The code in question is located [here](https://github.com/openaustralia/planningalerts/blob/89151a33bb4ec817d53aa1d9cc9187ef4f41d254/app/models/api_key.rb)

benrfairless:
## Relevant issue(s)
Fixes #2030

## What does this do?
Increases the default daily API call limit for trial users from 1 to 10.

## Why was this needed?
The previous limit of 1 call per day was insufficient for users to effectively test the service without risking abuse.

## Implementation/Deploy Steps (Optional)
- Deployment using existing processes

### DIFF
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -71,7 +71,7 @@ class ApiKey < ApplicationRecord
 
   sig { returns(Integer) }
   def self.default_daily_limit_trial
-    1
+    10
   end
 
   # TODO: Should this be longer (like 28 days)?


ianheggie-oaf (approved):
Makes sense to me, and the user is informed by app/views/api_keys/_create_key_box.html.erb what the new limit is (for new keys).

@benrfairless - I am assuming we don't need to inform people who have just started a trial - they can be pleasantly surprised they can do more than the 1 / day (unless this was in response to a specific complaint).

PR #2043, opened 2026-04-28
Add Dependabot Tapioca workflow

benrfairless:
## What does this do?
Adds a GitHub Actions workflow for Dependabot to automatically update Tapioca RBI files.

## Why was this needed?
CI will always fail when Dependabot creates a PR, as the Tapioca gems are not updated. This new Github action ensures that when Dependabot creates or updates a PR, the Tapioca RBI files are updated.

PR #2048, opened 2026-05-07
Revert "Add Dependabot Tapioca workflow"

benrfairless:
Reverts openaustralia/planningalerts#2043

When this workflow runs, no other checks are run. 

See https://github.com/openaustralia/planningalerts/pull/2047 as an example 

8315f7c544a3382517a101bb009c19fec3509bd0 performs 6 checks but 0b7e9898c4900a77e465eb16b54e51f045934a3f only performs SonarCube


ianheggie-oaf (approved):
Approved - I am trusting your conclusion is correct, because at a glance its a bit confusing why adding the tapioca workflow should have changed other checks

Issue #2014, opened 2026-03-25
Migrate Google Maps from client ID to API key before May 2026


benrfairless:
We need to Migrate from client ID to API key before May 2026. Refer to [this slack thread](https://oaf.slack.com/archives/C4D3LGYVA/p1774245625190919) for further details.

ianheggie-oaf:
@jamezpolley gave me a replacement production.yml.enc to rotate keys

https://oaf.slack.com/archives/C2U7XUPTR/p1779428176138659

benrfairless:
I have discussed this with @jamezpolley, and reviewed Client ID usage in the Google Cloud console. We are not using a Client ID in the Planning Alerts organisation, so this is a non-issue - no further investigation is required.

jamezpolley:
I've just had a closer look at the email about this, and noticed a detail that's really, really important that I'd previously not noticed.

> 
> | Client ID | Associated Cloud Project | API | % of client ID traffic |
> |---|---|---|---|
> | gme-openaustraliafoundation | map-for-work-ext-191635714 | Maps Static API | 1 |
> | gme-openaustraliafoundation | map-for-work-ext-191635714 | Street View Static API | 1 |

I think the reason the project ID is really jumping out at me now is that I saw it last week when I was updating the biling account, and I dug in to see what is using it. The answer I found is - no idea at all what's using it, but it's nothing to do with PlanningAlerts. I've removed that from the billing account. Whatever it is, it's not related to planningalerts, and the only way we'll ever figure out what it is is in the unlikely event that we find something broken. 

There's nothing to do here, sorry for wasting eveyone's time.

PR #2070, opened 2026-06-10
Allow credentials:edit of production on a dev system

ianheggie-oaf:
## Relevant issue(s)

Was found when working on the following issue, but is useful stand alone:
* https://github.com/openaustralia/planningalerts/issues/2014

## What does this do?

You can edit credentials with RAILS_ENV=production on a dev system (that is missing a config file cap sets up for production and was crashing due to the assumption that file exists)

## Why was this needed?

I wanted to see what the values where on my own system.


benrfairless (approved):
@ianheggie-oaf I do not know why this is failing CI. Can you review and advise?

Issue #229, opened 2013-12-29
Create a public mailing list to have scraper errors sent to (PA-214)


oajirabot:
<p>...so that anyone can debug and resolve issues with the scrapers.</p>

<p>So maybe just setup a Google Group, then change the address errors get sent to<span class="error">&#91;0&#93;</span>.</p>

<p><span class="error">&#91;0&#93;</span> <a href="http://github.com/openaustralia/planningalerts-parsers/blob/master/config.ru#L13" class="external-link">http://github.com/openaustralia/planningalerts-parsers/blob/master/config.ru#L13</a></p>

---

**Issue details imported from Jira:**

| Detail | Value |
| --- | --- |
| Assignee | Unassigned |
| Reporter | Henare Degan |
| Created | 2010-10-06 16:03:35 |
| Updated | 2011-11-23 11:53:39 |

---
### Imported Jira Comments

**Henare Degan** - 2011-11-23 11:53:39

> <p>We're slowly moving to ScraperWiki which means this is less important. WONTFIX.</p>
>

PR #2114, opened 2026-07-29
Update gems to fix security vulnerabilities (non-breaking)

ianheggie-oaf:
## Description

Bumps a set of direct and transitive gem dependencies to fix vulnerabilities flagged by `bundle-audit`, without touching gems that are pinned to specific versions for documented reasons (`rails`, `devise`, `zeitwerk`).

Direct dependency changes in `Gemfile`:
- `httparty` -> `>= 0.24.0` (CVE-2025-68696, SSRF leading to API key leakage)
- `nokogiri` -> `>= 1.19.4` (GHSA-5prr-v3j2-97mh, out-of-bounds read)
- `selenium-webdriver` (test group) -> `>= 4.14.0` (CVE-2023-5590, NULL pointer dereference)
- Removed `webdrivers` gem (test group) - unmaintained, and its version constraint (`< 4.11`) was blocking the `selenium-webdriver` fix above. `selenium-webdriver` >= 4.6 ships Selenium Manager, which replaces what `webdrivers` did (automatic driver download/version matching).

Added explicit floors for transitive dependencies flagged by `bundle-audit`, so a future lockfile regen can't silently resolve them back down:
- `bcrypt` -> `>= 3.1.22` (CVE-2026-33306, via devise)
- `msgpack` -> `>= 1.8.2` (CVE-2026-54522, via bootsnap)
- `rails-html-sanitizer` -> `>= 1.7.1` (GHSA-cj75-f6xr-r4g7, via actionview)
- `webrick` -> `>= 1.8.2` (CVE-2025-6442, via rackup)

Also removed a dead `require "webdrivers/geckodriver"` from `spec/spec_helper.rb`, left over from the `webdrivers` gem removal. Confirmed no `spec/support/*.rb` file registers a custom Firefox/geckodriver Capybara driver, the suite runs on Capybara's built-in headless Chrome driver, so this require was unused.

**Not included in this PR** (need a dedicated follow-up, as they require major version bumps):
- Rails CVEs requiring Rails `~> 7.2` (currently pinned `~> 7.1.0`)
- Devise CVEs requiring devise `>= 5.0.3` (currently pinned `~> 4.2`)
- The Rails bump also intersects with the existing `zeitwerk ~> 2.6.0` pin, which is there to work around a tapioca compatibility issue - that needs resolving as part of the Rails upgrade, not as a side effect of this PR.

## Motivation and Context

Routine dependency security maintenance, output of a `bundle-audit` run flagging the CVEs/GHSAs listed above. Scoped deliberately to changes unlikely to break anything, existing version pins with documented reasons were left alone.

## How Has This Been Tested?

- [x] Ran automated tests on my own system (`./,bundle-test`, full rspec suite via docker compose)

Test run: 571 examples, 0 failures, 6 pending (5 pre-existing "not yet implemented" pending specs unrelated to this change, 1 pre-existing accessibility pending spec unrelated to this change).

Developed and tested on Ubuntu 24.04.4 LTS x86_64 with ruby installed by mise, mysql 8.0, redis 5:7.0, make 4.3.

## Screenshots (if appropriate)

N/A - dependency/config only, no UI change.

## Types of Changes

- [ ] Documentation

## Checklist

---

Signed-off-by: Ian Heggie (OAF) <[ian@oaf.org.au](mailto:ian@oaf.org.au)>
Assisted-by: Claude:claude-sonnet-5


ianheggie-oaf:
See also internal: Security & robustness updates for Planning Alerts (PA) [#229](https://github.com/openaustralia/oaf-internal/issues/229)

benrfairless (approved):
Nice work, this is carefully scoped. I checked all seven cited advisories against `rubysec/ruby-advisory-db` and every CVE/GHSA ID, title and `patched_versions` floor matches what's in the Gemfile, with no floor set below the patched version. Leaving `rails`, `devise` and `zeitwerk` alone with a documented follow-up is the right call.

Approving, since the lockfile resolves clean and nothing here blocks. A few things worth picking up.

## One worth fixing in this PR

`gem "erb", "~> 6.0"` is doing more than its comment suggests. erb has a live advisory: CVE-2026-41316 (CVSS 8.1, `@_init` deserialisation guard bypass via `def_module` / `def_method` / `def_class`), patched in `>= 6.0.4`, or `~> 4.0.3.1` / `~> 4.0.4.1` / `~> 6.0.1.1`.

`~> 6.0` permits 6.0.0 through 6.0.3, which are all vulnerable. The lockfile resolves 6.0.6 so `bundle-audit` is happy today, but that is exactly the silent walk-back the transitive floors in this PR exist to prevent. Suggest:

```ruby
gem "erb", ">= 6.0.4" # fix CVE-2026-41316 (ERB deserialisation guard bypass)
```

`psych "~> 5.1"` has no advisory in ruby-advisory-db at all, and currently resolves to the same 5.1.2 that Ruby 3.3.4 ships, so it is a no-op with an upper bound attached. Happy either way, but worth a line in the description on why it is there. Same for `bundler-audit` and `ruby_audit`, which aren't mentioned in the description at the moment.

## Description correction

The suite runs headless **Firefox**, not Chrome. `Capybara.javascript_driver = :selenium_headless` maps to `browser: :firefox` with `-headless` in capybara 3.40.0 (`lib/capybara/registrations/drivers.rb`); `:selenium_chrome_headless` is the Chrome one. So `require "webdrivers/geckodriver"` was matched to the configured driver rather than left over from something else.

It also isn't exercised either way. Every `js: true` in the suite is commented out:

```
spec/spec_helper.rb:113
spec/features/search_for_applications_spec.rb:204
spec/features/sign_up_for_alerts_spec.rb:377
```

So the 571 green examples don't actually demonstrate Selenium Manager working here, because nothing instantiates a browser. The removal is still fine, but the reasoning as written would mislead whoever next enables one of those specs.

Small related thing: the rewritten VCR comment says the `ignore_hosts "github.com", ...` lines now benefit Selenium Manager. Selenium Manager is a native binary invoked as a subprocess, so its downloads never pass through Ruby's HTTP stack and WebMock can't see them. That config is dead now rather than repurposed. Harmless, just inaccurate.

## Follow-ups, not for this PR

- `bundler-audit` and `ruby_audit` are good additions, but there's no CI job, so they only run when someone remembers. Worth a job alongside `lint`.
- Heads up that `ruby_audit` will go red immediately against the pinned Ruby 3.3.4: CVE-2025-58767 and CVE-2025-61594 both need `~> 3.3.10`. Legitimate findings rather than noise, so probably deserves its own issue.

## Mine, not yours

The `arm64-darwin` lockfile commit and the sorbet `.claude/` ignore are both mine and unrelated to your security work. `arm64-darwin-25` in particular is macOS-version-specific and will churn on the next `bundle lock` from a different machine. I'll strip that back to plain `arm64-darwin` and split the sorbet change out.

Also a tiny typo in the selenium-webdriver comment: "repalces".

---

*Reviewed with Claude Code (claude-opus-5). The advisory checks were run against rubysec/ruby-advisory-db; the conclusions are mine.*

Issue #2116, opened 2026-07-31
Capistrano Deployment stalls at puma:reload


benrfairless:
**Describe the bug**
When I attempt to perform a capistrano deployment, the deployment stalls at the `puma:reload` stage.

**To Reproduce**

1. Given I have access to the server via SSH
2. When I go to the planningalerts repo
3. And run `bundle exec cap production deploy`
5. I expected the deployment to complete
6. But Instead the deployment stalls on the following section:
```
00:40 puma:reload
      01 /bin/systemctl --user status planningalerts_puma_production > /dev/null
    ✔ 01 deploy@ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com 0.206s
      02 /bin/systemctl --user reload planningalerts_puma_production
    ✔ 01 deploy@ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com 0.208s
```
I have to Ctrl+C out of the deployment. It does not fail. It just stalls.

**Expected behavior**
It shouldn't take more than a few minutes to reload Puma on the server. I note that the server does not go offline when this issue happens.

**Screenshots/Output**
https://oaf.slack.com/archives/C098U38GY8P/p1785474109176439

**Additional context**
I've already asked @ianheggie-oaf to have a look at this as a priority, and to deploy himself if he can.

ianheggie-oaf:
On a freshly booted system:

Given I can `ssh pa1` and `ssh pa2` and I have the entries that set my ssh key:
```
Host pa1 planningalerts1 ip-172-31-2-63 ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com 
        User deploy # ubuntu
        Hostname ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com 
        IdentityFile ~/.ssh/id_rsa-OAF
        # IdentityFile ~/.ssh/id_ed25519-OAF
        IdentitiesOnly yes

Host pa2 planningalerts2 ip-172-31-18-88 ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com
        User deploy # ubuntu
        Hostname ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com
        IdentityFile ~/.ssh/id_rsa-OAF
        # IdentityFile ~/.ssh/id_ed25519-OAF
        IdentitiesOnly yes
```
And pa1 has:
```
deploy@ip-172-31-2-63:~$ egrep ian .ssh/authorized_keys 
[...]

PR #2118, opened 2026-07-31
Update puma service on deployment with up to date service config

ianheggie-oaf:
## Description

Update puma service to reflect recent capistrano puma gem

* Use latest template and fixed template error from gem
* notify systemd rather than simple type (so it will restart if puma process freezes rather than just dies)
* Check linger is set but doesn't try and fix (we dont have password less sudo)
* Always updates systemd service config

Based on (rather than stack as 3 PRs depend directly on it):
* https://github.com/openaustralia/planningalerts/pull/2137

## Motivation and Context

Fixes service config was not being updated in normal deploy (and puma:enable calls `sudo` by default which deploy user doesn't have access to):
* https://github.com/openaustralia/planningalerts/issues/2116

## How Has This Been Tested?

Developed and tested on Ubuntu 24.04.4 LTS x86_64 with ruby installed by mise, mysql 8.0, redis 5:7.0, make 4.3

Deployed using updated capistrano config.

Checked systemd was being alerted to status (each 5 seconds) and that it restarts a process that dies:
```
deploy@ip-172-31-2-63:~$ sleep 3 ; systemctl --user show planningalerts_puma_production -p WatchdogTimestamp ; date ; echo === ; sleep 3 ; systemctl --user show planningalerts_puma_production -p WatchdogTimestamp ; date ; echo === ; sleep 3 ; systemctl --user show planningalerts_puma_production -p WatchdogTimestamp ; date ; echo === ; sleep 3 ; systemctl --user show planningalerts_puma_production -p WatchdogTimestamp ; date ; echo === ;
WatchdogTimestamp=Fri 2026-07-31 22:31:11 AEST
Fri Jul 31 22:31:14 AEST 2026
===
WatchdogTimestamp=Fri 2026-07-31 22:31:16 AEST
Fri Jul 31 22:31:17 AEST 2026
===
WatchdogTimestamp=Fri 2026-07-31 22:31:16 AEST
Fri Jul 31 22:31:20 AEST 2026
===
WatchdogTimestamp=Fri 2026-07-31 22:31:21 AEST
Fri Jul 31 22:31:23 AEST 2026
===
deploy@ip-172-31-2-63:~$ systemctl --user show planningalerts_puma_production -p MainPID
MainPID=37771
deploy@ip-172-31-2-63:~$ (sleep 10; kill -STOP "$(systemctl --user show planningalerts_puma_production -p MainPID --value)") & \
journalctl --user -u planningalerts_puma_production -f
[1] 64609
Jul 31 21:12:44 ip-172-31-2-63 systemd[593]: Starting Puma HTTP Server for planningalerts (production)...
Jul 31 21:12:54 ip-172-31-2-63 systemd[593]: Started Puma HTTP Server for planningalerts (production).
Jul 31 21:49:22 ip-172-31-2-63 systemd[593]: Reloading Puma HTTP Server for planningalerts (production)...
Jul 31 21:49:22 ip-172-31-2-63 systemd[593]: Reloaded Puma HTTP Server for planningalerts (production).
Jul 31 21:52:08 ip-172-31-2-63 systemd[593]: Reloading Puma HTTP Server for planningalerts (production)...
Jul 31 21:52:52 ip-172-31-2-63 systemd[593]: Reloaded Puma HTTP Server for planningalerts (production).
Jul 31 21:53:40 ip-172-31-2-63 systemd[593]: Reloading Puma HTTP Server for planningalerts (production)...
Jul 31 21:53:57 ip-172-31-2-63 systemd[593]: Reloaded Puma HTTP Server for planningalerts (production).
Jul 31 22:04:15 ip-172-31-2-63 systemd[593]: Reloading Puma HTTP Server for planningalerts (production)...
Jul 31 22:04:37 ip-172-31-2-63 systemd[593]: Reloaded Puma HTTP Server for planningalerts (production).

Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Watchdog timeout (limit 10s)!
Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Killing process 37771 (ruby) with signal SIGABRT.
Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Killing process 60721 (ruby) with signal SIGABRT.
Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Killing process 60758 (ruby) with signal SIGABRT.
Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Killing process 60788 (skylightd) with signal SIGABRT.
Jul 31 22:33:36 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Killing process 60856 (ruby) with signal SIGABRT.
Jul 31 22:33:37 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Main process exited, code=dumped, status=6/ABRT
Jul 31 22:33:37 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Failed with result 'watchdog'.
Jul 31 22:33:37 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Consumed 22min 12.656s CPU time.
Jul 31 22:33:38 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Scheduled restart job, restart counter is at 1.
Jul 31 22:33:38 ip-172-31-2-63 systemd[593]: Stopped Puma HTTP Server for planningalerts (production).
Jul 31 22:33:38 ip-172-31-2-63 systemd[593]: planningalerts_puma_production.service: Consumed 22min 12.656s CPU time.
Jul 31 22:33:38 ip-172-31-2-63 systemd[593]: Starting Puma HTTP Server for planningalerts (production)...
Jul 31 22:33:48 ip-172-31-2-63 systemd[593]: Started Puma HTTP Server for planningalerts (production).
```

## Types of Changes

- [x] New feature (non-breaking change that adds functionality) - makes sure service config is up to date and loaded, and will persist over reboot
- [ ] Documentation

## Checklist:


config/deploy.rb
benrfairless, line 1: Four things here, two of them blocking.

### 1. MemoryMax has gone backwards from 1680M to 1500M, but the comment still says 1,680 (blocking)

```ruby
# Total limit = 3 workers x 550 MB + 30 MB master = 1,680 MB
set :puma_service_unit_props, %w[MemoryMax=1500M TimeoutStopSec=300]
```

1500M is the value from *before* 311c52d99 ("Fine tune memory limits to JUST under NewRelic alert at 75%"), which deliberately raised it to 1680M and rewrote the comment to match. The new comment has come across in this PR but the old number has too, so it looks like an accidental revert while restructuring into `puma_service_unit_props`.

`MemoryMax` is a hard cgroup limit, so if 1680M was the right number we'd be inviting kernel OOM kills under load. It's also the one thing manual testing on staging wouldn't surface. Which of the two is correct?

### 2. `puma_systemctl_user` is a mode flag, not a username (blocking)

The gem only ever compares this setting against `:system`, in six places, to pick system-unit vs user-unit mode, and its default is `:user`. Setting it to the string `"deploy"` works today purely because `"deploy" != :system` behaves the same as `:user`.

The trouble is it's now doing double duty as the username for `check_lingering`, and that coupling isn't visible from either file. Suggest dropping this line and reading the user off the role instead, see my comment on the rake file.

### 3. `puma:install` runs after Puma has already been restarted

The gem registers `after 'deploy:published', 'puma:smart_restart'`, and `deploy:published` comes before `deploy:finishing`. With `puma_phased_restart` true, `smart_restart` resolves to `puma:reload`, which is `systemctl --user reload`, i.e. SIGUSR1 to the running master.

`daemon-reload` in `puma:install` does reapply cgroup properties like `MemoryMax` to a running unit, so that part lands fine. But `Type=` and `WatchdogSec=` are only read when the unit starts, and a phased reload doesn't re-exec it. So notify and the watchdog activate at the next reboot or manual restart, unsupervised, rather than during the deploy.

That also explains the testing in the description: those watchdog timestamps follow a full restart, not a deploy.

Suggest `before "deploy:publishing", "puma:install"`, plus a note in the PR that each host needs a one-off `systemctl --user restart planningalerts_puma_production` to pick up the type change this time round.

### 4. `check_lingering` should run at `deploy:check`

It's a precondition check wired to run after publish, after restart and after install, then raise before `deploy:cleanup` and `deploy:log_revision`. If linger isn't set you get a half-finished deploy plus a failure, when the check costs one `capture`. `before "deploy:check", "puma:check_lingering"` would catch it before anything has changed.

### Minor, take or leave

- A few of these settings match the gem defaults already, so they aren't doing anything: `puma_systemd_watchdog_sec` is 10 by default, and `service_unit_type` already returns `"notify"` for Puma 6.1+. Worth dropping so it's clearer which settings are actually carrying weight.
- The hardcoded log paths equal what the gem derives from `shared_path` given `deploy_to = /srv/www/production`, so they'd silently diverge if `deploy_to` ever changed. `-> { File.join(shared_path, "log", "puma.log") }` keeps it derived.
- `puma_error_log` now points at the same file as `puma_access_log`, so `puma_error.log` stops being written. Two append handles on one file is safe, just checking it's deliberate and that nothing on the box tails the old path.
- Stray bare `#` on line 171, looks like a placeholder that lost its text.

config/deploy/templates/puma.service.erb
benrfairless, line 1: Suggest deleting this file rather than updating it.

Strip the attribution line added at the top and it's byte-identical to `capistrano3-puma-8.0.0/lib/capistrano/templates/puma.service.erb`. I diffed it both ways to be sure.

`compiled_template_puma` falls through to the gem's copy when the local file is absent, so removing it changes nothing about the generated unit. What it does remove is the copy that'll silently go stale at the next gem upgrade, which is exactly the bug this PR is fixing: the old one was pinned at 6.0.0.beta.1 and drifted.

All the customisation now lives in `puma_service_unit_props` and friends, which is a much better place for it.

One thing worth confirming either way: the description mentions "fixed template error from gem", but the file matches the gem exactly. Did that fix land upstream in 8.0.0, or has it been lost in the copy?

lib/capistrano/tasks/puma_check_lingering.rake
benrfairless, line 1: The task itself reads well. The warning message is genuinely useful, `2>/dev/null | grep ... || true` correctly avoids a spurious non-zero exit, and `# typed: strict` matches what's in `lib/tasks/data.rake` and `db_stats.rake`.

Two changes though.

### 1. Don't take the username from `puma_systemctl_user` (blocking)

`fetch(:puma_systemctl_user, fetch(:user))` only returns "deploy" because deploy.rb sets it to that, and that setting is really the gem's system-vs-user mode flag. If anyone ever sets it to the legitimate `:system`, this becomes `loginctl show-user system`. Remove the line from deploy.rb and it falls back to the gem default, giving `loginctl show-user user`.

Worth flagging a trap in the obvious fix: `fetch(:puma_lingering_user)` resolves to `fetch(:lingering_user, fetch(:user))`, and `aws_ec2_register(user: "deploy")` passes `user:` through to `server ip, ...` as a per-server property, not the global `:user`. Nothing in config/ or Capfile calls `set :user`, so `fetch(:user)` is nil and you'd end up running `loginctl show-user` with an empty argument.

Reading it off the role sidesteps all of that:

```ruby
task :check_lingering do
  on roles(fetch(:puma_role)) do |role|
    puma_user = role.user
    ...
```

### 2. Wire it to `deploy:check` instead

Covered in my comment on deploy.rb, but in short: raising here means the deploy has already published and restarted before we find out. `before "deploy:check", "puma:check_lingering"` catches it before anything has changed.

benrfairless (dismissed):
Good change overall, and the testing writeup is exactly right: SIGSTOPping the master to prove the watchdog actually fires is much better evidence than assuming `Type=notify` works. Moving the memory and timeout tuning out of the template and into `puma_service_unit_props` is the right direction too.

Requesting changes on three things:

1. `MemoryMax` has gone from 1680M back to 1500M while the comment above it still says 1,680. Looks like an accidental revert of 311c52d99.
2. `puma_systemctl_user` is being used as a username, but the gem treats it as a `:system`/`:user` mode flag. It works by luck today, and the coupling with `check_lingering` isn't visible from either file.
3. The template file is now byte-identical to the gem's, so it'd be better deleted than kept in sync.

Two more worth a look before merge: `puma:install` runs after `puma:smart_restart`, so `Type=` and `WatchdogSec=` won't actually take effect until the next reboot or manual restart, and `check_lingering` raises after the deploy has already published and restarted.

One framing note that might save some time. A few of the new settings aren't doing what the description implies. `Type=<%= service_unit_type %>` is an unchanged context line in the template diff and the gem's helper already returns "notify" for Puma 6.1+, `puma_systemd_watchdog_sec` matches the gem default of 10, and the hardcoded log paths equal what the gem derives from `shared_path`. The real payload here is `puma_enable_lingering, false`, the `puma:install` hook, and `puma_service_unit_props`. The simple to notify switch comes from finally regenerating a unit file that had been frozen at 6.0.0.beta.1, not from the new settings.

Detail is in the three file comments. Nothing security-related found, and I'm not raising a test coverage gap given there's no practical harness for Capistrano hooks in this repo.

config/deploy.rb
ianheggie-oaf: RE:
1. MemoryMax has gone backwards from 1680M to 1500M, but the comment still says 1,680 (blocking)  -Fixed
2. puma_systemctl_user is a mode flag, not a username (blocking) - Fixed
3. puma:install runs after Puma has already been restarted - Fixed
4. check_lingering should run at deploy:check - Fixed
Minor, take or leave:
* commented out puma_systemd_watchdog_sec which is 10 by default
* commented out service_unit_type already returns "notify" for Puma 6.1+
* shared_path is based on deploy_path with an override of the subdir name, so shared_path should be used not deploy_path to keep DRY
* puma_error_log pointed at the same file as puma_access_log until the gem was updated, so I chose to continue using just one log file - production does not have any puma error logs.
* Stray bare # on line 171, looks like a placeholder that lost its text - fixed

lib/capistrano/tasks/puma_check_lingering.rake
ianheggie-oaf: done - see above

config/deploy/templates/puma.service.erb
ianheggie-oaf: I have removed the file - the fix was adding a missing newline, but when I rechecked just now the gem file was correct, so somehow I fixed something that I accidentaly introduced myself ??

ianheggie-oaf:
> Requesting changes on three things

@benrfairless - I have updated these, rebased and its ready for reviewing again

lib/capistrano/tasks/puma_check_lingering.rake
```diff
@@ -0,0 +1,28 @@
...
+namespace :puma do
+  desc "Check lingering is already set"
+  task :check_lingering do
+    on roles(fetch(:puma_role)) do |role|
+      puma_user = role.user
+
+      linger_status = capture("loginctl show-user #{puma_user} 2>/dev/null | grep Linger || true")
+
+      if linger_status.include?("Linger=yes")
+        info "OK: Linger is enabled for #{puma_user} on #{host}."
```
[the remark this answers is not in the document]
ianheggie-oaf: Fixed and tested locally:
```bash
ianh@ben:.../planningalerts (feature/2116-update-p)$ be cap production puma:check_lingering 
...
00:00 puma:check_lingering
      OK: Linger is enabled for deploy@ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com.
      OK: Linger is enabled for deploy@ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com.
```

ianheggie-oaf:
@benrfairless back to you for review pls

benrfairless (commented):
See comments from sentry that need review

ianheggie-oaf:
Fix tested on my dev system:

```
ianh@ben:.../planningalerts (feature/2116-update-p)$ be cap production puma:check_lingering
rvm 1.29.12-next (master) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]
rvm 1.29.12-next (master) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]
ruby-3.3.4
ruby-3.3.4
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
00:00 puma:check_lingering
      OK: Linger is enabled for deploy@ec2-3-27-58-54.ap-southeast-2.compute.amazonaws.com.
      OK: Linger is enabled for deploy@ec2-3-27-189-101.ap-southeast-2.compute.amazonaws.com.
```

ianheggie-oaf:
@benrfairless - over to you for re-review

benrfairless (approved):

PR #2124, opened 2026-08-03
Remove misleading request count from API trial box

### LINKED ISSUE #2122: API trial box displays "(14 requests)" - it's showing the trial duration in days, not a request count
**Describe the bug**

The "Instant access 14 day trial" box on the API page ends with a parenthetical "(14 requests)". That figure is wrong: it renders `ApiKey.default_trial_duration_days`, which is the trial length in days, not a number of requests. The real limit is 10 requests per day, or 140 across the whole trial. Nothing in the system caps a trial key at 14 requests.

It also contradicts the sentence it sits inside, so anyone reading carefully gets two different numbers for the same thing.

**To Reproduce**

1. Given I am not signed in, or am signed in without an existing API key
2. When I go to the Planning Alerts API page (`/api/howto`)
3. And scroll down to the "Instant access 14 day trial" box
4. Then I expected the box to describe the trial limits consistently
5. But instead I see "Enjoy up to 10 requests a day (returning up to a total of 1,000 applications per day) for 14 days. **(14 requests)**", where the trailing "(14 requests)" is the trial duration mislabelled as a request count

**Expected behavior**

The box should state the trial limits once, without a contradictory figure. The parenthetical should be deleted, since the sentence already gives the request limit:

> Enjoy up to 10 requests a day (returning up to a total of 1,000 applications per day) for 14 days.

For comparison, `app/views/api_keys/_create_key_box.html.erb` already describes the same trial correctly and without a parenthetical:

> Build, test and explore with live data capped at 10 calls per day for 14 days.

**Screenshots/Output**

Rendered text:

> Instant access 14 day trial
> Build, test and explore with live API data
> Enjoy up to 10 requests a day (returning up to a total of 1,000 applications per day) for 14 days. **(14 requests)**

Source, in `app/views/documentation/api_howto.html.erb`:

```erb
Enjoy up to
<%= ApiKey.default_daily_limit_trial %>
requests a day (returning up to a total of
<%= number_with_delimiter(ApiKey.default_daily_limit_trial * Application.max_per_page) %>
applications per day) for
<%= pluralize(ApiKey.default_trial_duration_days, "day") %>. (<%= number_with_delimiter(ApiKey.default_trial_duration_days) %> requests)
```

**Additional context**

*How it probably happened*

The trial box mirrors the Community and Standard plan cards further up the same file, where the parenthetical correctly converts applications into requests:

```erb
Return up to
<%= number_with_delimiter(ApiKey.default_daily_limit_community * Application.max_per_page) %>
applications per day. (<%= number_with_delimiter(ApiKey.default_daily_limit_community) %> requests)
```

In the plan cards the parenthetical adds information, because the headline figure there is applications. In the trial box the headline figure is already requests, so the parenthetical has nothing to add - and the wrong constant got dropped into it.

*The rest of the message is accurate*

Checked against the code while investigating:

- **"14 day trial"** - `ApiKey.default_trial_duration_days` returns 14, and `ApiKeysController` creates the key with `expires_at: 14.days.from_now`. Correct.
- **"up to 10 requests a day"** - `ApiKey.default_daily_limit_trial` returns 10, set as the key's `daily_limit` and enforced by the Rack::Attack throttle in `config/initializers/rack_attack.rb`. Correct.
- **"up to a total of 1,000 applications per day"** - `10 × Application.max_per_page` (`max_paginates_per 100`). Correct as a theoretical maximum, ie. only if every request asks for and fills a 100-result page.

*Two related things, out of scope here*

Noticed while checking the above. Happy to split these into their own issues if they're worth pursuing:

1. **"Per day" means per UTC day.** The Rack::Attack throttle uses a `period: 1.day` window, so a trial key's quota resets at UTC midnight rather than Australian midnight. `ApiController#update_api_usage` already carries a TODO acknowledging this for usage stats, but the public wording says "per day" unqualified.
2. **Expired keys get the generic throttle limit.** `ApiKey.daily_limit_with_default` falls back to `default_daily_limit` (1,000) when the key is invalid or expired, so an expired trial key is throttled at 1,000/day rather than 10. No data exposure, since `require_api_key` returns 401 for expired keys, but it does mean an expired key can generate 1,000 error responses a day instead of 10.

---

Investigated with the help of Claude (Opus 5), reviewed and filed by me.

benrfairless:
## Description

Removes the trailing "(14 requests)" from the "Instant access 14 day trial" box on the API page, in `app/views/documentation/api_howto.html.erb`. That figure rendered `ApiKey.default_trial_duration_days`, which is the trial length in days, dressed up as a request count.

The box now reads:

> Enjoy up to 10 requests a day (returning up to a total of 1,000 applications per day) for 14 days.

Also adds a spec to `spec/features/documentation_pages_spec.rb` asserting the sentence is there and "(14 requests)" isn't, so the wording can't regress quietly. It passes `normalize_ws: true` because the view breaks that sentence across several lines and Capybara's rack_test driver keeps the newlines in page text.

The Community and Standard plan cards higher up the same file keep their parentheticals. There the headline figure is applications, so converting to requests adds something.

## Motivation and Context

Fixes #2122.

Nothing in the system caps a trial key at 14 requests. `ApiKey.default_daily_limit_trial` is 10 a day, enforced by the Rack::Attack throttle, across the 14 days `ApiKeysController` sets on the key. So the parenthetical was both wrong and a contradiction of the sentence it sat inside: anyone reading carefully got two different numbers for the same limit.

It looks like it was copied from the plan cards, where the same pattern is correct, with the wrong constant dropped in. `app/views/api_keys/_create_key_box.html.erb` already describes this trial correctly, so this brings the two descriptions into line.

## How Has This Been Tested?

The first two aren't ticked, and I'd rather say so than tick them. The change was written in a container with no working app environment (no Docker daemon, and the `pg` gem wouldn't build), so the suite couldn't be started and the page couldn't be loaded.

Verification rests on the Actions run for this PR, where `lint`, `type_check` and `test` have all passed.

Worth a second pair of eyes on the new spec in particular, since it has only ever run in CI.

Text-only change.

Before: `... for 14 days. (14 requests)`
After: `... for 14 days.`

## Types of Changes

- [ ] Documentation

## Checklist:

The change is itself a wording fix on a documentation page, and nothing else documents the trial limits, so there's no further documentation to update.

### DIFF
--- a/app/views/documentation/api_howto.html.erb
+++ b/app/views/documentation/api_howto.html.erb
@@ -180,7 +180,7 @@
         requests a day (returning up to a total of
         <%= number_with_delimiter(ApiKey.default_daily_limit_trial * Application.max_per_page) %>
         applications per day) for
-        <%= pluralize(ApiKey.default_trial_duration_days, "day") %>. (<%= number_with_delimiter(ApiKey.default_trial_duration_days) %> requests)
+        <%= pluralize(ApiKey.default_trial_duration_days, "day") %>.
       </p>
       <%= image_tag "illustration/key.svg", alt: "", size: "237x155", class: "my-8" %>
       <% if current_user %>
--- a/spec/features/documentation_pages_spec.rb
+++ b/spec/features/documentation_pages_spec.rb
@@ -83,6 +83,12 @@ describe "Browsing basic documentation pages" do
       end
     end
 
+    it "describes the trial limits without a contradictory request count" do
+      # The view breaks this sentence across several lines, so normalise whitespace
+      expect(page).to have_content("Enjoy up to 10 requests a day", normalize_ws: true)
+      expect(page).to have_no_content("(14 requests)", normalize_ws: true)
+    end
+
     # rubocop:disable RSpec/NoExpectationExample
     it "renders a snapshot for a visual diff", :js do
       page.percy_snapshot("API")


ianheggie-oaf (approved):
Approved - LGTM

PR #2125, opened 2026-08-03
Work out the years on the API how-to page instead of hard-coding it

### LINKED ISSUE #2121: API how-to page says "14 yrs" but we've been collecting data since 2009
**Describe the bug**

The API how-to page (https://www.planningalerts.org.au/api/howto) says:

> **14 yrs**
> We've been doing this for more than 14 years with an extremely experienced team. Leverage that for your own needs rather than trying to reinvent the wheel.

The claim is technically true but stale and undersells us by two to three years. The production database shows:

| Milestone | Earliest record | Age as at 3 August 2026 |
|---|---|---|
| First application scraped | 13 Dec 2009 | 16 years, 8 months |
| First email alert signup | 2 Feb 2011 | 15 years, 6 months |
| First public comment | 7 Jun 2011 | 15 years, 2 months |

Application volumes ramp from 2,352 in Dec 2009 to 59,954 in 2010 and 76,905 in 2011, so late 2009 is a genuine start of continuous collection rather than a stray backfilled record.

The number appears to be hard-coded, so it will keep going stale.

**To Reproduce**

1. Given I am not signed in
2. When I go to https://www.planningalerts.org.au/api/howto
3. Then I expected the stated number of years to match how long we've actually been collecting planning data
4. But instead I see "14 yrs", which was accurate around 2023-24 and is now two to three years out of date

**Expected behavior**

The page reflects the real figure, and ideally does so without needing a manual edit each year. Two options:

1. Reword to something self-updating, e.g. "We've been collecting planning data since 2009." No number to maintain.
2. Keep the "N yrs" stat block but calculate it from a fixed start year (2009) at render time.

If we keep a number, "16 yrs" is right for data collection, or "15 yrs" if we'd rather date it from when the email alerts service started in Feb 2011.

**Screenshots/Output**

Verified against the Planning Alerts production Postgres database via Metabase:

```sql
SELECT
  (SELECT MIN(first_date_scraped) FROM applications) AS app_first_scraped,
  (SELECT MIN(created_at) FROM alerts)               AS first_alert,
  (SELECT MIN(created_at) FROM comments)             AS first_comment;
```

```
app_first_scraped        | first_alert             | first_comment
2009-12-13 18:06:57+11   | 2011-02-02 04:46:15+11  | 2011-06-07 00:57:52+10
```

**Additional context**

Two caveats on the data:

- The 2009-2011 dates were inherited from the MySQL to Postgres migration, so they're only as reliable as that migration. They're internally consistent and consistent with the volume ramp, which is reassuring, but they haven't been checked against an independent source.
- The old MySQL Planning Alerts database in Metabase (`PlanningAlerts (mysql) OLD`, database ID 2) no longer connects: `Unknown database 'pa-production'`. Worth cleaning up as a separate issue if it's genuinely decommissioned.

Also worth a look: whether the same hard-coded "14 years" figure appears anywhere else on the site or in other services' copy.

---

*Written with the help of Claude (Opus 5), then reviewed by me. The figures above came from queries I ran against our own Metabase.*

benrfairless:
## Description

The API how-to page said "14 years" in three places. All three now call a new `years_collecting_data` helper that works the number out at render time, so it reads 16 today and won't go stale again.

- `config/application.rb`: adds `config.planningalerts_first_application_scraped_on = Date.new(2009, 12, 13)`, alongside the existing `planningalerts_*` settings.
- `app/helpers/application_helper.rb`: adds `years_collecting_data`, which counts complete years from that date. It rounds down, so the surrounding "over" and "more than" wording stays true.
- `app/views/documentation/api_howto.html.erb`: the intro bullet, the big number, and the paragraph beside it all use the helper. Tailwind classes and spacing are untouched.

The helper went in `ApplicationHelper` rather than `ApiHowtoHelper` because `ApiHowtoHelper` is entirely about building example API URLs, and this is general site copy that other pages might want later.

## Motivation and Context

Fixes #2121.

The claim was accurate around 2023-24 and has been stale since. We've been collecting planning data since the first application was scraped on 13 December 2009, so the real figure is 16 years. Because it was hard-coded it would have kept going stale every December.

One thing to check before merging: the 13 December 2009 date now drives the whole page. It's the minimum `applications.first_date_scraped` in production, which came through the MySQL to Postgres migration and hasn't been verified against an independent source. It is consistent with the application volume ramp (2,352 in December 2009, 59,954 in 2010), which is reassuring.

## How Has This Been Tested?

New tests:

- `spec/helpers/application_helper_spec.rb`: covers the figure today and the anniversary boundary, 15 on 12 December 2025 and 16 on 13 December 2025. Uses `Timecop.freeze`, matching the pattern in `spec/features/authorities_spec.rb`.
- `spec/features/documentation_pages_spec.rb`: checks the rendered page inside the existing "api landing page" block, with an `around` hook freezing the date so it doesn't drift.

Run locally against Postgres 16 with PostGIS and Redis:

- `bin/rubocop --parallel`: 243 files, no offences
- `bin/erblint --lint-all`: 189 files, no errors
- `bin/srb tc`: no errors
- `bin/rspec spec/helpers/application_helper_spec.rb`: 11 examples, 0 failures
- `bin/rspec spec/features/documentation_pages_spec.rb -e "api landing page"`: 3 examples, 0 failures, including the axe accessibility check

I haven't loaded the page in a browser by hand.

The first CI run failed on `spec/features/search_for_applications_spec.rb:140` with a Selenium `ElementClickInterceptedError`, a `<td>` obscuring the Search button on the address search page. Nothing in this change touches that page, and the whole `no results in new design` block including the `:js` example passes locally on this branch. Re-ran the job and it went green, so it looks like a browser-test flake.

No screenshot from me, but the Percy "API" snapshot will show a real visual diff as the big number goes from 14 to 16. Worth an eye on that. Note the snapshot renders at the real current time, so it will legitimately change again on 13 December 2026.

## Types of Changes

- [ ] Documentation

## Checklist:

---

Assisted-by: Claude Code

https://claude.ai/code/session_019r94foyNrCpxt95SBKMEuD


ianheggie-oaf (approved):
Approved - LGTM

Issue #2132, opened 2026-08-04
A failed GitHub issue sync re-imports the whole authority


benrfairless:
**Describe the bug**

`ImportApplicationsService#call` finishes an import and then calls the broken-scraper issue
sync as its last statement:

```ruby
# app/services/import_applications_service.rb:26-31
def call
  time = Benchmark.ms { import_applications_date_range }
  logger.info "Took #{(time / 1000).to_i} s to import applications from #{authority.full_name_and_state}"

  SyncGithubIssueForAuthorityService.call(logger:, authority:)
end
```

That call is not rescued. The only `rescue StandardError` in the file is at line 91, inside
the per-record loop in `import_applications_date_range`, so it does not cover line 30.
`ImportApplicationsJob#perform` doesn't rescue either, and unlike
`queue_up_import_applications_job.rb:15` it doesn't set `sidekiq_options retry:`, so it gets
Sidekiq's default of 25 retries.

The effect is that anything going wrong while talking to GitHub fails the whole job, and
Sidekiq then re-runs `ImportApplicationsService` from the top, which means re-fetching the
full morph.io result set for that authority. The applications from the first attempt were
already imported and logged by the time the sync ran, so the re-import is wasted work. A
persistent GitHub problem turns into up to 25 full re-imports for every affected authority.

Ways this can currently fail:

- GitHub is down or rate limiting
- the credential is wrong or expired
- `attach_issue_to_project` raises a bare `RuntimeError` at
  `sync_github_issue_for_authority_service.rb:162` when the project's custom fields don't
  resolve, which happens if someone renames a field on the project board

**Expected behavior**

A failure to file or label a scraper issue shouldn't invalidate an import that already
succeeded. The import should be treated as done, the GitHub failure reported to Honeybadger,
and the job should not retry.

**Additional context**

Noticed while working on #2131, which replaces the personal access token in this code path
with a GitHub App. That change doesn't cause this, but it does widen the surface: JWT signing
against a malformed private key, the installation token exchange, and the new dependency on
organisation-level Projects permission all become new ways to trigger the re-import loop.
Kept out of scope there deliberately so the auth change stays reviewable.

Worth deciding as part of a fix: whether the sync belongs inside `ImportApplicationsService`
at all, or whether it should be its own job queued after a successful import. Splitting it
would make the failure boundary obvious rather than relying on a rescue in the right place.

PR #2134, opened 2026-08-04
Create broken-scraper issues from an org-owned GitHub App

### LINKED ISSUE #2131: Create broken-scraper GitHub issues from an org-owned identity instead of a personal account
**Describe the feature**

When a council's scraper stops returning data, Planning Alerts automatically opens an issue in `planningalerts-scrapers/issues` and adds it to the org project so the work of fixing scrapers can be prioritised. That happens in `SyncGithubIssueForAuthorityService`, which authenticates with a single GitHub personal access token stored in Rails credentials as `github_personal_access_token`.

That token belongs to a personal GitHub account (@mlandauer), so every one of those issues is authored by that account. Recent examples: planningalerts-scrapers/issues#1418, #1417, #1416.

Two problems with that:

- It ties a core automated process to one person's account. If the account rotates 2FA, revokes the token, or that person stops being involved, scraper issues quietly stop being created and we probably don't notice until a broken council goes untracked.
- It reads as a person filing hundreds of near-identical issues rather than as a service, which is confusing for anyone looking at the repo.

**Describe the solution you'd like**

Move the automation to an identity owned by the organisation rather than by a person. A GitHub App looks like the right fit:

- Issues would be authored by something like `planningalerts[bot]`, which reads correctly.
- Credentials are org-owned. Any org admin can rotate the private key without touching a personal account.
- Permissions can be scoped to exactly what's needed: repository `Issues` (write) plus organisation `Projects` (write) for the project field updates. Note the access needs to be on `planningalerts-scrapers`, not `openaustralia`, since that's where both the issues repo and project #3 live.
- Installation access tokens do work with the Projects GraphQL API, so `addProjectV2ItemById` and `updateProjectV2ItemFieldValue` are supported: https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects

What would need to change in this repo:

- `app/services/sync_github_issue_for_authority_service.rb:22` (the GraphQL `headers` method) and `:128` (the Octokit client) both read `Rails.application.credentials[:github_personal_access_token]`. Both are evaluated per call, so a token that's generated and refreshed on demand can be swapped in without restructuring the class.
- Installation access tokens expire after an hour, so we'd mint one per run (or cache it with its expiry) from the App's private key rather than storing a long-lived secret.
- New credentials to add: App ID, installation ID and private key. `github_personal_access_token` goes away.
- The non-production path writes to `planningalerts-scrapers/test-issues` and project #4 (`:111-117`), so the App needs to cover those too, or we need a separate installation for development.
- `SCHEMA` at `:31-37` downloads the GraphQL schema over HTTP at boot when `schema.json` isn't present, which happens before any per-call token exists. Worth confirming that path still works, or committing the schema so it never needs fetching.

**Describe alternatives you've considered**

- **A dedicated machine account with its own personal access token.** Much less work: create the account, add it to the org, generate a fine-grained token, swap the credential. Issues would be authored by that account rather than a bot. The downsides are that it's still a user account someone has to own, it needs its own mailbox and 2FA, and fine-grained tokens expire and need renewing by hand.
- **Leave it as is and just document it.** Cheapest option, but it doesn't fix the dependency on one person's account.
- **An org-owned personal access token.** Not possible. Personal access tokens are always owned by a user, even where an org has to approve them.

**Additional context**

The only caller is `ImportApplicationsService` (`app/services/import_applications_service.rb:30`), so the blast radius of this change is limited to the broken-scraper issue sync.

Nothing needs migrating. Existing issues keep their current author, and `github_issues` rows only store the repo and issue number, so they stay valid.

Open question: should the App be owned by `openaustralia` and installed on `planningalerts-scrapers`, or owned by `planningalerts-scrapers` directly?

benrfairless:
## Description

Broken-scraper issues are now created by a GitHub App owned by the `planningalerts-scrapers`
organisation instead of a personal access token.

- New `GithubAppTokenService` signs a short-lived JWT with the App's private key and swaps it
  for an installation access token, holding on to it until five minutes before it expires.
  One run of the sync makes about a dozen requests, so minting per request would be wasteful.
- Both credential reads in `SyncGithubIssueForAuthorityService` now go through it. Both were
  already evaluated per call, so this is a substitution rather than a restructure.
- Adds the `jwt` gem, since Octokit doesn't generate JWTs itself.
- Fixes the two `# TODO: Check for errors` comments in `attach_issue_to_project`. Nothing
  looked at what the seven field-update mutations returned, so an error on any of them was
  discarded and the issue would end up on the project with most of its fields blank and
  nothing in the logs. A null project is now handled explicitly too, because that is what a
  missing organisation Projects permission looks like from the API: no errors, just a null.
- Adds the first spec for `SyncGithubIssueForAuthorityService`, which had none.

### Credentials

`github_app.id`, `github_app.installation_id` and `github_app.private_key` are **included in
this PR** for production, so it is safe to deploy from this branch. Verified by decrypting
`production.yml.enc` and using the stored credential to mint a real installation access token
from github: it authenticates as `PlanningAlerts Bot` and comes back with
`issues: write`, `organization_projects: write`, `metadata: read`.

Without them every import job would raise, and because the sync call isn't rescued, Sidekiq
would retry the whole morph.io import each time (see #2132). The service now fails with the
missing key name and the command to set it, rather than a Sorbet type error.

**Development credentials are not included, because no `config/credentials/development.key`
exists anywhere.** It isn't in the infrastructure repo, which only manages `production.key`,
and it isn't in 1Password. So `development.yml.enc` (which holds Google Maps keys for
development, added in ba578de36) can't currently be edited by anyone here. Nothing sets
`require_master_key`, so the app doesn't complain: `Rails.application.credentials` just
returns nil in development.

That's pre-existing and doesn't block this deploy, but it does mean local verification against
`test-issues` and project #4 isn't possible, so verification rests on the manual run below
plus watching the first production import cycle. Worth chasing the key separately and storing
it in 1Password so it stops being one person's.

`github_personal_access_token` is not referenced anywhere in the code any more, but is
deliberately still present in `production.yml.enc` so there's something to fall back to if the
first import cycle doesn't go as expected. Removing it should be one more commit on this
branch once a real issue has been created by the bot, so that #2131 closes when the work is
genuinely complete.

## Motivation and Context

Resolves #2131.

Every broken-scraper issue was authored by one person's personal account (@mlandauer), which
tied a core automated process to that account staying available and read as an individual
filing hundreds of near-identical issues. Credentials are now org-owned and rotatable by any
org admin, and permissions are scoped to `Issues: write` plus organisation `Projects: write`
and nothing else.

Related: #2132, which covers the unrescued sync call and the full re-import on retry. This PR
makes previously-silent GraphQL failures raise, so it interacts with that: a failure that used
to pass unnoticed will now fail the import job. That is the right trade, since silent data
loss is worse, but it makes #2132 more worth doing.

## How Has This Been Tested?

Manually, against the non-production targets only, before writing any of the service code.
The load-bearing assumption was that an *installation* access token can read and mutate an
org-level ProjectV2, and if that failed the whole approach would have collapsed back to the
machine-account alternative described in the issue. It works: the token read org project #4
with all seven custom fields resolving, attached an issue to it and set a field. The issue it
created is planningalerts-scrapers/test-issues#128, authored by `planningalerts-bot[bot]`.

Automated: 585 examples, 0 failures, plus `bin/srb tc` and `rubocop` clean. Run in the
container, since Sorbet is amd64-only.

The tests were checked by breaking the code rather than only by watching them pass. Reverting
either call site to the old personal access token fails all four wiring examples; before this
PR that same revert left the suite green. Removing the GraphQL error check, the null-project
guard, or the token expiry margin each fails its own example and only its own.

Worth knowing for anyone testing locally: `Issues` was disabled on
`planningalerts-scrapers/test-issues`, so the non-production path of this service could never
have worked in development. I've enabled it. That is independent of this change.

n/a

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

Flagged as breaking because it needs new credentials before it deploys, and because issues
change author. Nothing about the public site changes.

## Checklist:

No documentation in the repo mentions GitHub credentials, so there was nothing to update.


ianheggie-oaf (approved):
Approved - LGTM and a welcomed change

PR #2154, opened 2026-08-17
Remove stale TODOs from migration and initializer

benrfairless:
## Description

Removes 2 config/db TODO comments that are verifiably resolved. Comment-only change; no behaviour affected.

| TODO removed | Why it's stale |
|---|---|
| `db/migrate/20230206002921_add_first_date_scraped_to_applications.rb:3` — "make this null: false in a later migration once data is loaded" | Done two days later: `20230208012842_make_first_date_scraped_not_null.rb` exists and `db/schema.rb:151` shows `first_date_scraped ... null: false` |
| `config/initializers/elasticsearch.rb:5` — "With the line above can I get rid of the stuff below?" | The question is answered: no. `ElasticSearchClient` is still used directly by `LogApiCallService` and `ElasticsearchSnapshotJob` |

Note: the other TODO in `elasticsearch.rb` (line 2, "store this in a more sensibly named/namespaced variable") is still relevant and is deliberately left in place — it is tracked in #2164.

## Motivation and Context

Part of a repo-wide TODO audit: all 279 `TODO`/`FIXME` comments in the codebase were classified against the current code and the issue tracker. This is one of four PRs (#2152, #2153, #2154, #2156) removing the 19 verifiably-stale TODOs; the still-relevant ones are grouped into tracking issues #2157–#2168 (plus bugs #2169, #2170). Stale TODOs are noise that misleads readers about outstanding work — each removal above documents the evidence that the work is already done.

## How Has This Been Tested?

- [ ] Checked affected area manually on my own / staging system
- [ ] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

This is a comment-only change (deleting 2 Ruby comment lines); it produces no change to rendered output or behaviour, so the CI suite (tests, lint, type_check) is the relevant verification.

## Screenshots (if appropriate):

Not applicable — comment-only change.

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

None of the above apply: this only deletes stale comments from the source.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2153, opened 2026-08-17
Remove stale TODOs from controllers and services

benrfairless:
## Description

Removes 3 controller/service TODO comments whose work has verifiably already been done. Comment-only change; no behaviour affected.

| TODO removed | Why it's stale |
|---|---|
| `app/controllers/alerts/registrations_controller.rb:15` — "Make link in email point to the right place" | Fixed the same day the TODO was added: 83171c4c8 added `after_confirmation_path_for` (users/confirmations_controller.rb), redirecting confirmed users with alerts to `alerts_path` |
| `app/controllers/applications_controller.rb:18` — "Handle the situation where the authority name isn't found" | Handled: `Authority.find_short_name_encoded!` raises `ActiveRecord::RecordNotFound`, which renders the styled 404 page for this HTML controller |
| `app/services/process_alert_and_record_stats_service.rb:27` — "Figure out the impact on caching updating this so regularly" | Moot: `Stat.emails_sent` / `Stat.applications_sent` have no readers anywhere in app/lib/config (only their own specs) — no cached page displays these stats |

## Motivation and Context

Part of a repo-wide TODO audit: all 279 `TODO`/`FIXME` comments in the codebase were classified against the current code and the issue tracker. This is one of four PRs (#2152, #2153, #2154, #2156) removing the 19 verifiably-stale TODOs; the still-relevant ones are grouped into tracking issues #2157–#2168 (plus bugs #2169, #2170). Stale TODOs are noise that misleads readers about outstanding work — each removal above documents the evidence that the work is already done.

## How Has This Been Tested?

- [ ] Checked affected area manually on my own / staging system
- [ ] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

This is a comment-only change (deleting 3 Ruby comment lines); it produces no change to rendered output or behaviour, so the CI suite (tests, lint, type_check) is the relevant verification.

## Screenshots (if appropriate):

Not applicable — comment-only change.

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

None of the above apply: this only deletes stale comments from the source.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #2152, opened 2026-08-17
Remove stale TODOs from views and components

benrfairless:
## Description

Removes 10 view/component TODO comments whose work has verifiably already been done. Comment-only change; no rendered output or behaviour is affected.

| TODO removed | Why it's stale |
|---|---|
| `app/views/application/_html_head.html.erb:22` — "#1769 Only load the font weights that we're actually using" | #1769 was closed as completed (July 2024); the stylesheet link now loads exactly the weights in use (Fira Sans 300–700, Merriweather 700/900) |
| `app/views/application/_welcome_banner.html.erb:14` — "button doesn't do anything without javascript" | The partial is dead code — its render call in `layouts/application.html.erb:13` is commented out (it was the June 2024 redesign announcement) |
| `app/views/applications/show.html.erb:74–75` — "link in notified text should go to the new alert sign up page" / "Should we have the link at all now?" | `_notified_text.html.erb` no longer contains any link — the whole partial is plain text |
| `app/views/applications/show.html.erb:80` — "What to do if there are 0 comments?" | Handled: `_comments_text.html.erb` renders a "Be the first to comment" link for the zero case |
| `app/views/documentation/api_howto.html.erb:137` — "Add documentation for bulk data API" | Bulk API docs now exist in `api_developer.html.erb:127–162` (heading, description, example URL) |
| `app/views/contact_messages/_form.html.erb:41` — "Make sure recaptcha errors are shown here" | Verified end-to-end: `verify_recaptcha(model:)` adds `:base` errors in the controller and the view renders `f.error :base` directly below |
| `app/views/api_keys/confirm.html.erb:8–9` — "don't use the word dickhead" / "update the wording about it being a trial" | Both done in 10ef8255e ("Updated wording on the interstitial") — the copy is now "We trust you. Don't abuse that trust." with rewritten trial/plan bullets |
| `app/components/share_button_component.html.erb:2` — "Use web share API" | Implemented: `navigator.canShare`/`navigator.share` with clipboard fallback in this same component |

## Motivation and Context

Part of a repo-wide TODO audit: all 279 `TODO`/`FIXME` comments in the codebase were classified against the current code and the issue tracker. This is one of four PRs (#2152, #2153, #2154, #2156) removing the 19 verifiably-stale TODOs; the still-relevant ones are grouped into tracking issues #2157–#2168 (plus bugs #2169, #2170). Stale TODOs are noise that misleads readers about outstanding work — each removal above documents the evidence that the work is already done.

## How Has This Been Tested?

[...]

PR #2156, opened 2026-08-17
Remove stale TODO questions from specs

benrfairless:
## Description

Removes 4 spec TODO comments that are questions the current code already answers. Comment-only change; no behaviour affected.

| TODO removed | Why it's stale |
|---|---|
| `spec/controllers/applications_controller_spec.rb:38–39` — "Can this line be removed? It seems to be a duplicate of expectation on final line" | No, it can't: the `allow(Application).to receive(:find)` stub is what returns the instance carrying the `location: nil` stubs. The final line is an `expect(...)` assertion, not a stub — they're different things |
| `spec/features/search_for_applications_spec.rb:152` — "Would it be better to only enable the feature temporarily?" | Already effectively temporary: `spec_helper.rb` deletes `tmp/flipper.pstore` before every example, so `Flipper.enable` can't leak between examples |
| `spec/features/authorities_spec.rb:39` and `:86` — "I suspect we'll need to lock down the date of the application so that percy snapshots are consistent" | Already done in both blocks: fixed `date_scraped` values (2020-01-01/08/15) plus `Timecop.freeze` around each `visit` |

## Motivation and Context

Part of a repo-wide TODO audit: all 279 `TODO`/`FIXME` comments in the codebase were classified against the current code and the issue tracker. This is one of four PRs (#2152, #2153, #2154, #2156) removing the 19 verifiably-stale TODOs; the still-relevant ones are grouped into tracking issues #2157–#2168 (plus bugs #2169, #2170). Stale TODOs are noise that misleads readers about outstanding work — each removal above documents the evidence that the work is already done.

## How Has This Been Tested?

This is a comment-only change (deleting 5 Ruby comment lines); it produces no change to rendered output or behaviour, so the CI suite (tests, lint, type_check) is the relevant verification.

Not applicable — comment-only change.

## Types of Changes

- [ ] Documentation

None of the above apply: this only deletes stale comments from the source.

## Checklist:

### DIFF
--- a/spec/controllers/applications_controller_spec.rb
+++ b/spec/controllers/applications_controller_spec.rb
@@ -35,8 +35,6 @@ describe ApplicationsController do
 
       allow(application).to receive_messages(location: nil, find_all_nearest_or_recent: [])
 
-      # TODO: Can this line be removed? It seems to be a duplicate of
-      # expectation on final line.
       allow(Application).to receive(:find).with("1").and_return(application)
 
       get :show, params: { id: 1 }
--- a/spec/features/authorities_spec.rb
+++ b/spec/features/authorities_spec.rb
@@ -36,7 +36,6 @@ describe "Authorities" do
       sign_in create(:confirmed_user, name: "Jane Ng")
       authority = create(:authority, full_name: "Byron Shire Council", morph_name: "planningalerts-scrapers/byron")
       # We need it to have at least one application so it's not "broken"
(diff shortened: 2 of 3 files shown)


ianheggie-oaf (approved):
Approved, though I wonder if it would be better to change to a comment on decision and terse reason why its not needed so future someone doesn’t go through the “maybe we should do…” loop

PR #2182, opened 2026-08-18
Support Sentry CLI v4 and v3, and skip release recording when neither is usable

### LINKED ISSUE #2181: Deploy fails with Errno::ENOENT when sentry-cli is not installed locally
## What happened

Running `cap production deploy` on a machine without `sentry-cli` installed aborts at the very end of the deploy:

```
(Backtrace restricted to imported tasks)
cap aborted!
Errno::ENOENT: No such file or directory - sentry-cli (Errno::ENOENT)
lib/capistrano/tasks/sentry.rake:15:in `block (3 levels) in <top (required)>'
lib/capistrano/tasks/sentry.rake:14:in `block (2 levels) in <top (required)>'
Tasks: TOP => sentry:release
** DEPLOY FAILED
```

The deploy itself has actually completed by this point (the task runs `after "deploy:finished"`), but Capistrano reports `DEPLOY FAILED`, which is alarming and hides the real state.

## Why

`sentry:release` guards against a missing/unauthenticated `sentry-cli` with:

```ruby
unless test("sentry-cli info")
```

That guard was written on the assumption that `test` returns false when the command fails. Over SSH that's true, but this task uses `run_locally`, and SSHKit's local backend execs the command directly rather than through a shell. When the binary doesn't exist at all, the exec raises `Errno::ENOENT` instead of returning a non-zero exit, so the warn-and-skip path never runs and the whole deploy aborts.

The guard currently only works for the "installed but not authenticated" case, not the "not installed" case it was also meant to cover.

## Fix

Rescue `Errno::ENOENT` around the `test("sentry-cli info")` call and treat it the same as an unauthenticated CLI: print the existing warning and skip recording the release, letting the deploy finish successfully.

benrfairless:
## Description

Two changes to the `sentry:release` Capistrano hook in `lib/capistrano/tasks/sentry.rake`:

1. Rescue `Errno::ENOENT` around the CLI detection so a missing binary is treated the same as an unauthenticated one: print the existing warning, skip recording the release, and let the deploy finish successfully.
2. Support both Sentry CLI v4 (binary `sentry`) and v3 (binary `sentry-cli`), preferring v4. v4 is a ground-up rewrite ([migration guide](https://cli.sentry.dev/migrating-from-v3/)) that renames the binary, makes command groups singular (`releases` -> `release`), and replaces `deploys new --release X -e env` with positional arguments (`release deploy <version> <environment>`). The hook now detects which CLI is installed and authenticated and issues the matching command shapes.

The README "Sentry release tracking" setup instructions now recommend v4 while documenting that v3 keeps working.

## Motivation and Context

Fixes #2181.

Today's production deploy aborted with `DEPLOY FAILED` at the very last step:

```
cap aborted!
Errno::ENOENT: No such file or directory - sentry-cli (Errno::ENOENT)
lib/capistrano/tasks/sentry.rake:15
```

The guard assumed `test` returns false when the command fails. Over SSH that holds, but this task uses `run_locally`, and SSHKit's local backend execs the command directly rather than through a shell, so a missing binary raises `Errno::ENOENT` instead of returning a non-zero exit. The warn-and-skip path the task was written for never ran, and the deploy reported failure even though it had already completed.

The v4 rename is also a likely cause of the binary being missing in the first place: a deployer who installs the current CLI gets `sentry`, not `sentry-cli`, so without v4 support the hook would silently skip recording releases forever.

## How Has This Been Tested?

Verified the failure mode and the detection logic directly against SSHKit's local backend:

- missing binary raises `Errno::ENOENT` from `test` (reproduces the crash)
- neither CLI installed -> detection returns nil (warn and skip)
- v4 installed -> picked ahead of v3
- v3 only and unauthenticated -> nil (warn and skip)
- v3 only and working -> picked

`bundle exec rubocop lib/capistrano/tasks/sentry.rake` (no offenses) and `bundle exec srb tc` (no errors). There are no specs covering Capistrano tasks.

N/A

## Types of Changes

- [x] Documentation

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

---

AI assistance disclosure: this change was made with the help of OpenCode running anthropic.claude-fable-5, as noted in the commit trailers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,17 +184,21 @@ bundle exec cap --tasks
 
 ### Sentry release tracking
 
-Deploys automatically record a release and deploy in Sentry (org `oaf-org-au`, project `planning-alerts`) so issues can be tied to the deploy that introduced them. This runs via a Capistrano hook using `sentry-cli` on your machine (part of [#2049](https://github.com/openaustralia/planningalerts/issues/2049)).
+Deploys automatically record a release and deploy in Sentry (org `oaf-org-au`, project `planning-alerts`) so issues can be tied to the deploy that introduced them. This runs via a Capistrano hook using the Sentry CLI on your machine (part of [#2049](https://github.com/openaustralia/planningalerts/issues/2049)).
 
-One-time setup per deployer machine:
+The hook supports both CLI v4 (binary `sentry`) and the legacy v3 (binary `sentry-cli`), preferring v4 when both are installed. See the [v4 migration guide](https://cli.sentry.dev/migrating-from-v3/).
 
-1. Install sentry-cli: `brew install getsentry/tools/sentry-cli` on macOS, or `curl -sL https://sentry.io/get-cli/ | sh` on Linux
-2. Authenticate: run `sentry-cli login` — it opens Sentry to create a personal User Auth Token (needs the `project:releases` and `org:read` scopes) and stores it in `~/.sentryclirc`
-3. Verify: `sentry-cli info` should show you are authenticated against the `oaf-org-au` org
+One-time setup per deployer machine (v4, recommended):
+
+1. Install the CLI: `brew install getsentry/tools/sentry` on macOS, or `curl https://cli.sentry.dev/install -fsS | bash` on Linux
+2. Authenticate: run `sentry auth login` — it opens Sentry in a browser to authenticate (the token needs the `project:releases` and `org:read` scopes)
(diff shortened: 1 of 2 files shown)


ianheggie-oaf (approved):
Approved - LGTM at a quick glance

PR #2190, opened 2026-08-19
Guard copy and share buttons for missing clipboard API

### LINKED ISSUE #2186: Bug: Copy and share buttons silently do nothing where the clipboard API is unavailable

Planning Alerts

### What happened?

Two components call `navigator.clipboard.writeText` without first checking that `navigator.clipboard` exists:

- `app/components/share_button_component.html.erb:23`
- `app/components/click_to_copy_component.html.erb:7`

Where `navigator.clipboard` is undefined the call throws a `TypeError`. Because the exception happens *before* `copied = true` runs, the button gives no feedback whatsoever: no "Link copied", no error, nothing. From the person's point of view the button is simply broken.

`navigator.clipboard` is undefined in a non-secure context and in some in-app browsers and older Firefox versions.

Reported by Sentry: [PLANNING-ALERTS-H](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-H), 2 events, 1 person, Firefox, on an application's external link page.

Worth noting how this compounds in `share_button_component`: the clipboard call is the fallback branch, taken only when `navigator.canShare` is unavailable. So the people most likely to depend on the fallback are exactly the ones getting nothing.

### Steps to reproduce

1. Open Planning Alerts in a browser or context where `navigator.clipboard` is unavailable (a non-secure context is the easiest way, or override `navigator.clipboard` to `undefined` in the console)
2. Click "Share this application", or the Copy button next to an API key
3. Nothing happens, and a `TypeError` is thrown to the console

### What did you expect to happen?

The button should either copy the value and confirm it, or tell the person it could not. It should not look like it worked, and it should not look dead.

### Browser, device, or OS (if relevant)

Recorded on Firefox. Applies to any browser or context without the async clipboard API.

### Screenshots or error messages

```
TypeError: can't access property "writeText", navigator.clipboard is undefined
```

### Anything else?

#1746 ("The Share this application button doesn't do anything!") was the same user-visible symptom from a different cause, closed in July 2024. Worth a look for context on what was decided then.

A commit or PR containing `Fixes PLANNING-ALERTS-H` will close the Sentry issue.

---

Raised from Sentry with AI assistance (Claude, model `claude-opus-5`).

benrfairless:
## Description

Two Alpine.js click handlers called `navigator.clipboard.writeText` without checking that `navigator.clipboard` exists:

- `app/components/share_button_component.html.erb` (the clipboard fallback branch, used when the Web Share API is unavailable)
- `app/components/click_to_copy_component.html.erb`

Where the async clipboard API is undefined (non-secure contexts, some in-app browsers, older Firefox) the call threw a `TypeError` before `copied = true` ran, so the button gave no feedback at all and looked dead.

This change:

- Keeps both buttons hidden where they could never work. They already render `invisible` and are only revealed once Alpine boots, so the reveal binding now also checks for the needed APIs (`navigator.clipboard`, or for the share button `navigator.canShare || navigator.clipboard`). The click-to-copy value remains `select-all` so people can still copy it by hand.
- Awaits the `writeText` promise inside `try/catch`. A rejected or throwing write now shows a brief failure state ("Failed" / "Couldn't copy link") instead of falsely confirming the copy.
- Uses `shareData.url` for the clipboard write in the share button rather than duplicating the raw (unescaped) URL interpolation.

One subtlety worth knowing for review: Alpine only wraps an inline handler as statements (in an async IIFE) when the expression starts with `if (`, so both handlers are structured that way. A handler starting with `try {` compiles to invalid JavaScript and fails silently.

## Motivation and Context

Fixes #2186
Fixes [PLANNING-ALERTS-H](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-H) (2 events, 1 person, Firefox, on an application's external link page)

The people most likely to need the share button's clipboard fallback (no Web Share API) were exactly the ones hitting the crash.

## How Has This Been Tested?

Written test-first: the new feature spec was added and confirmed failing (4 of 6 examples) before the components were touched.

New feature spec `spec/features/copy_and_share_buttons_spec.rb` (headless Firefox, which has no Web Share API, so the share button specs exercise the same clipboard fallback branch people hit in the wild) covers, for each button:

- happy path: click copies and confirms ("Copied" / "Link copied")
- `writeText` rejects: failure state shown
- `navigator.clipboard` undefined at click time: failure state shown, no `TypeError`

All three CI gates pass locally via Docker Compose: `bin/rake` (647 examples, 0 failures), `bin/rake ci:type`, `bin/rake ci:lint`.

Captured from the feature spec run (headless Firefox). The states of each button:

| Normal | Copied | Failed |
|---|---|---|
| Copy button | "Copied" | "Failed" |
| Share this application | tick + "Link copied" | "Couldn't copy link" |

(Screenshot files attached in a comment below, as they can't be uploaded via `gh` directly.)

## Types of Changes

- [ ] Documentation

## Checklist:

---

This pull request was prepared with AI assistance (OpenCode, model `anthropic.claude-fable-5`).

### DIFF
--- a/app/components/click_to_copy_component.html.erb
+++ b/app/components/click_to_copy_component.html.erb
@@ -1,15 +1,27 @@
-<div class="flex items-baseline" x-data="{ copied: false }">
+<div class="flex items-baseline" x-data="{ copied: false, failed: false }">
   <div class="px-8 py-2 text-xl select-all text-navy bg-light-grey" x-ref="value"><%= content %></div>
   <%# We only want the button to appear once javascript (and alpine) are going %>
   <%# Using invisible here so that we avoid content reflow when javascript is running %>
+  <%# The clipboard API is unavailable in some contexts (e.g. non-secure ones). %>
+  <%# Keep the button hidden there because it could never work. The value above %>
+  <%# is still selectable so people can copy it by hand. See #2186. %>
   <button class="flex items-baseline invisible gap-2 px-4 py-2 text-xl font-semibold text-white w-28 bg-warm-grey"
-          x-on:click="
-            navigator.clipboard.writeText($refs.value.innerText);
-            copied = true;
-            setTimeout(() => { copied = false }, 750)"
-          x-bind:disabled="copied"
-          x-bind:class="{ 'invisible': false }">
-    <%= image_tag "copy-to-clipboard.svg", "x-show": "!copied" %>
-    <div x-text="copied ? 'Copied' : 'Copy' ">Copy</div>
+          x-on:click="if (navigator.clipboard) {
+            try {
+              await navigator.clipboard.writeText($refs.value.innerText);
+              copied = true;
+              setTimeout(() => { copied = false }, 750);
+            } catch {
+              failed = true;
+              setTimeout(() => { failed = false }, 2000);
+            }
+          } else {
+            failed = true;
+            setTimeout(() => { failed = false }, 2000);
+          }"
+          x-bind:disabled="copied || failed"
+          x-bind:class="{ 'invisible': !navigator.clipboard }">
+    <%= image_tag "copy-to-clipboard.svg", "x-show": "!copied && !failed" %>
+    <div x-text="copied ? 'Copied' : (failed ? 'Failed' : 'Copy')">Copy</div>
   </button>
 </div>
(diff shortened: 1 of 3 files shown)


benrfairless:
Screenshots of the button states, captured from the feature spec run in headless Firefox ([gist](https://gist.github.com/benrfairless/1b509740a794af7cde44ed049653099f)):

**Click-to-copy (external link page)**

| Normal | Copied | Write failed / API unavailable |
|---|---|---|
| Copy button | Copied | Failed |

**Share button (application page)**

| Normal | Link copied | Write failed / API unavailable |
|---|---|---|
| Share this application | Link copied | Couldn't copy link |

Where the needed APIs are missing at page load the buttons stay hidden entirely (no screenshot to show, by design). The failure state above is for when the API exists but the write is refused, or the API disappears between load and click.

ianheggie-oaf (approved):
Approved - LGTM and with tests!

Issue #2155, opened 2026-08-17
Feature: Create a Bulk Deletion Task based on Prefix and Authority ID


benrfairless:
### Which service is this about?

Planning Alerts

### What problem are you trying to solve?

We have a business need to remove a signfiicant number of requests that start with PA1, PA2, PA3 from a specific Authority

### What would you like to happen?

Create a rails/rake task that can take a partial `council_reference` and a full `authority_id` and remove (or hide when #2140 is deployed) the application from Planning Alerts.

The task must operate in dry-run mode and provide a report to the user of what applications have been removed.

My preference is to land #2140 first, and then just hide these requests.

### Alternatives you've considered

We could remove the applications manually - but that is going to take days.

Issue #2140, opened 2026-08-11
Add ability to hide an application without deleting it


benrfairless:
**Describe the feature**

Is your feature request related to a problem?
At the moment the only way to remove an application from public view on the site is to delete it outright. There's no option to hide an application while keeping its record intact.

Is your feature request related to something the service doesn't provide?
The service doesn't currently provide a way to hide an application from the site without deleting it.

**Describe the solution you'd like**

A "hide"/"unhide" action for an application that removes it from public listings, search results and any subscriber alerts, while keeping the underlying record so it isn't permanently lost and can be restored later if needed.

**Describe alternatives you've considered**

Deleting the application, but this permanently loses the data, which isn't appropriate when the record just needs to be hidden rather than removed.

**Additional context**

None.

PR #2192, opened 2026-08-19
Extend bulk application task to hide as well as delete

### LINKED ISSUE #2191: Feature: Extend the bulk application task to hide as well as delete

Planning Alerts

#2155 shipped `BulkDeleteApplicationsService` and the `planningalerts:bulk_delete_applications` rake task, which can only delete. #2140 has since landed, so applications can now be hidden (`hidden` and `hidden_reason` on `applications`, plus the `Application.visible` scope) without losing the record.

As flagged in #2155, hiding is the preferred outcome for the PA1/PA2/PA3 cleanup at that authority. Deleting is irreversible and throws away records we may want back. At the moment the only way to hide in bulk is to click through the admin one application at a time.

Extend the bulk task so an operator can choose to hide rather than delete:

- Add a hide mode that sets `hidden = true` and a supplied `hidden_reason` on every matching application, plus an unhide mode that reverses it.
- Keep the existing behaviour: match on `authority_id` and a partial `council_reference`, dry run by default, and report what was (or would be) changed.
- Drop the delete-mode guards when hiding. Applications with comments, and applications that are the target of an `ApplicationRedirect`, can be hidden safely and should not be skipped. The report should still show how many comments are affected.
- Update the Elasticsearch index. `Application#should_index?` returns false when hidden, so hidden records need to leave the index. A bare `update_all` skips the callbacks, which would leave applications hidden on the site but still returned by full text search.
- Require a reason when hiding. `hidden_reason` is published word for word on the public application page, so it should not fall back to the default when we are hiding hundreds of applications at once.
- Keep delete available for cases where the record genuinely has to go.

How it is structured is up to whoever picks it up. Options include generalising `BulkDeleteApplicationsService` into something like `BulkModerateApplicationsService`, or adding a sibling service and a `planningalerts:bulk_hide_applications` task. Either way the two paths should share the matching and reporting logic rather than drifting apart.

- Deleting, as the task does today. It works, but it is irreversible and loses records we may want to restore.
- Hiding each application through the admin interface. Fine for a handful, not for the volume described in #2155.

### Anything else?

Worth landing before we run the cleanup described in #2155, so those applications can be hidden rather than deleted.

Assisted-by: Claude:claude-opus-5

benrfairless:
## Description

Extends the bulk application task from #2155 so an operator can hide (and unhide) applications in bulk rather than only delete them.

- `BulkDeleteApplicationsService` is generalised into `BulkModerateApplicationsService` with delete, hide and unhide modes sharing the same matching and reporting logic, so the paths can't drift apart.
- New `planningalerts:bulk_hide_applications` and `planningalerts:bulk_unhide_applications` rake tasks sit alongside the existing `planningalerts:bulk_delete_applications` task, which keeps its exact behaviour and output. All three match on `authority_id` plus a partial `council_reference`, dry run by default, and report what was (or would be) changed.
- Hiding requires a reason, passed via the `REASON` environment variable so it can contain spaces and commas. It is required even for dry runs because it is published word-for-word on the public page for each hidden application - no silent fallback to the default explanation when hiding hundreds of applications at once.
- The delete-mode guards are dropped when hiding: applications with comments and applications that are the target of an `ApplicationRedirect` hide safely, and the report still shows how many comments are affected.
- Applications that are already hidden are skipped (and reported), so a reason set by hand in the admin is never overwritten. This also makes the task safe to re-run after a partial run.
- Unhiding reverses the hide and clears the reason.
- Each record is saved individually rather than with `update_all` so searchkick's callbacks run and hidden applications leave the Elasticsearch index (`Application#should_index?` returns false when hidden). A bare `update_all` would leave hidden applications still showing up in full text search.

Example usage:

```sh
REASON="Duplicate record" rake planningalerts:bulk_hide_applications[123,PA1]          # dry run
REASON="Duplicate record" rake planningalerts:bulk_hide_applications[123,PA1,execute]  # hide
rake planningalerts:bulk_unhide_applications[123,PA1,execute]                          # unhide
```

## Motivation and Context

Resolves #2191. As flagged in #2155, hiding is the preferred outcome for the PA1/PA2/PA3 cleanup - deleting is irreversible and throws away records we may want back, and #2140 made hiding possible but only one application at a time through the admin. Worth landing before running the cleanup described in #2155.

## How Has This Been Tested?

All via Docker Compose on the dev environment:

- Full spec suite (`bin/rake`): 681 examples, 0 failures. `bin/rake ci:type` and `bin/rake ci:lint` clean.
- Service specs extended to cover hide/unhide: blank-reason rejection, comments and redirect targets not skipped, already-hidden applications skipped with their reason preserved, re-run idempotency, and a regression guard that re-enables searchkick callbacks (with the reindex job stubbed) to assert a reindex is enqueued for every changed record and never in a dry run - guarding against a future switch to `update_all`.
- New rake task specs (`spec/lib/tasks`) cover argument parsing, `REASON` validation, dry-run defaults and the operator-facing output for all three tasks.
- Manual end-to-end smoke test in dev against real Elasticsearch: reindexed the seed application, ran the hide task with execute, drained the sidekiq queue, and confirmed the application was hidden with the reason stored and gone from full text search; then ran unhide and confirmed it returned to search with the reason cleared.

## Types of Changes

- [ ] Documentation

## Checklist:

---

AI assistance disclosure: written with OpenCode using anthropic.claude-fable-5, reviewed and tested by me.


ianheggie-oaf (approved):
Approved - LGTM, motivation and design clear at a quick glance through, has spec to test and will have a final smoke test when used

Issue #2195, opened 2026-08-24
Feature: pin or self-host the five third-party scripts we load from CDNs


benrfairless:
### Which service is this about?

Planning Alerts

### What problem are you trying to solve?

Every page executes third-party JavaScript we do not control and cannot verify:

- `app/views/application/_html_head.html.erb:23` floating-ui core 1.6.1
- `:24` floating-ui dom 1.6.5
- `:25` tsparticles confetti 3.0.3
- `:26` **alpinejs `3.x.x`**, a floating tag, so the code executing on our pages changes without any change here
- `app/views/authorities/show.html.erb:117` d3 3.5.12 from cdnjs

None carry an `integrity` attribute, so a compromised or substituted file would execute unnoticed. The floating Alpine tag is the sharpest edge: Alpine drives interactive behaviour across the whole site, and we would inherit a breaking 3.x change on their release schedule, not ours.

Separately, splitting Alpine and our own `application.js` across two origins is what makes the #2193 failure mode possible at all: one can succeed while the other fails.

### What would you like to happen?

Decide, per script, between pinning with subresource integrity and vendoring into the asset pipeline. Self-hosting removes the cross-origin split and the third-party trust, at the cost of owning updates by hand. Worth recording as `doc/adr/0001-*.md` once decided, since a future reader will otherwise assume the CDN was simply never revisited.

### Alternatives you've considered

Leaving it and relying on per-call-site guards, which is what #2193 does. That contains the symptom for people but does not reduce the trust we place in five third parties, and it does nothing about the floating version tag.

### Anything else?

Note that self-hosting alone would not have fixed #2193: if Alpine is missing, `x-cloak` keeps the fallback hidden regardless. See the sibling issue on that.

---

Raised with AI assistance (OpenCode, model `anthropic/claude-opus-5`).

Issue #2194, opened 2026-08-24
Bug: map fallback stays hidden when Alpine itself fails to load


benrfairless:
### Which service is this about?

Planning Alerts

### What happened?

The "We couldn't load the map" fallback is hidden by `x-cloak`, and `[x-cloak] { display: none !important }` lives in `app/assets/stylesheets/application.tailwind.css:5`, a static stylesheet with no JavaScript dependency. Alpine is what removes the attribute. So when Alpine itself fails to load from jsDelivr, nothing removes it, the fallback can never appear, and the person gets the blank green box the fallback exists to prevent. The same applies to the streetview and the alert radius map.

This is the remaining half of #2193. That issue fixed the case where Alpine loads and our own JavaScript does not. This is the case where Alpine is the thing missing.

### Steps to reproduce

1. Open any `/applications/:id` with DevTools request blocking on `cdn.jsdelivr.net`
2. Blank green box, no message, nothing in the console

### What did you expect to happen?

A fallback message. Fixing it means inverting the default so the fallback renders visible and Alpine hides it on success, which risks a flash of "We couldn't load the map" on every normal page load before Alpine boots. Mitigating that flash, with a CSS delay or by moving the reveal into the head, is the actual design work here.

### Anything else?

Related: #2185 (Google Maps API unreachable), #2193 (our own bundle unreachable). Also worth folding in the empty pink chart box on `/authorities/:id`, which is the same unexplained-blank-box problem and was deliberately left alone in #2193 to avoid making the "Authority" Percy snapshot depend on cdnjs being reachable from CI.

---

Raised with AI assistance (OpenCode, model `anthropic/claude-opus-5`).

Issue #2185, opened 2026-08-19
Bug: Google Maps fails to load for some people, mostly on Mobile Safari


benrfairless:
### Which service is this about?

Planning Alerts

### What happened?

Sentry browser monitoring went live on 18 August (718cb0ca3) and immediately recorded several issues where the Google Maps JavaScript API, or one of its libraries, fails to load in a real person's browser. When that happens the person gets a blank space where the map should be.

Setting aside crawler traffic, the affected events over seven days:

| Sentry issue | Error | Events from real browsers | Browsers |
| --- | --- | --- | --- |
| [PLANNING-ALERTS-8](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-8) | The Google Maps JavaScript API could not load. | ~42 | 20 Chrome, 18 Mobile Safari, 2 Edge, 1 Safari |
| [PLANNING-ALERTS-7](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-7) | The Google Maps JavaScript API could not load. | 17 | 13 Mobile Safari, 4 Safari |
| [PLANNING-ALERTS-A](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-A) | Could not load "streetview". | 12 | 11 Mobile Safari, 1 Safari |
| [PLANNING-ALERTS-C](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-C) | Could not load "marker". | 4 | 3 Mobile Safari, 1 Chrome |

PLANNING-ALERTS-8 looks much larger than it is: its other 385 events are all Baiduspider, which is being dealt with separately as noise.

The skew towards Mobile Safari is the interesting part, and it is worth establishing the cause before spending much effort here. It is consistent with content blockers, Lockdown Mode, or flaky mobile connections cutting off the request to `maps.googleapis.com`, rather than with anything wrong in our own code. If that is what it is, there is no fix available to us, but there is still a question about what the person sees.

### Steps to reproduce

Not reproduced by hand. Reconstructed from the recorded events. A content blocker set to block `maps.googleapis.com` on Mobile Safari is the most likely way to see it deliberately.

### What did you expect to happen?

Two things worth deciding, whatever the cause turns out to be:

- The page should not leave an unexplained blank space where the map was going to be. Either say the map could not be loaded, or fall back to something useful such as a link to the address on Google Maps.
[...]

PR #2196, opened 2026-08-24
Reach the map fallback when our own javascript is missing

### LINKED ISSUE #2193: Bug: Alpine x-init throws ReferenceError when app JS fails to load, bypassing the map fallback

Planning Alerts

### What happened?

Sentry is recording `ReferenceError`s from Alpine `x-init` expressions calling our own global functions that don't exist at the time Alpine starts:

| Sentry issue | Error | Where | Events / users |
| --- | --- | --- | --- |
| [PLANNING-ALERTS-1D](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-1D) | `initialisePano is not defined` / `initialiseBasicMapWithMarker is not defined` | `/applications/:id` | 4 / 2 (real Chrome users on Windows) |
| [PLANNING-ALERTS-11](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-11) | `barGraph is not defined` | `/authorities/:id` | 2 / 1 (Sogou spider hitting web1.blue directly) |

These functions definitely exist in the codebase, so this is a load-ordering/failure race, not a missing function:

- `initialisePano` and `initialiseBasicMapWithMarker` are defined in `app/assets/javascripts/maps.js`, bundled into `application.js` (`//= require maps`), included **without** `defer` in `app/views/application/_html_head.html.erb`
- `barGraph` is defined in `app/assets/javascripts/bar_graph.js`, included at the bottom of `app/views/authorities/show.html.erb`
- Alpine itself is a **separate** script from a different origin: `https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js`, loaded with `defer`

So whenever the jsDelivr Alpine script loads successfully but our `application.js` / `bar_graph.js` (or the CDN d3 that `bar_graph.js` needs) fails to load or is blocked, Alpine starts, evaluates `x-init`, and throws `ReferenceError`.

The nasty part: the map components are designed to degrade gracefully —

```erb
"x-init" => "initialiseBasicMapWithMarker($el, {...}).catch(() => mapFailed = true)"
```

— but that `.catch` only handles a **rejected promise**. When the function itself is undefined the expression throws synchronously, `mapFailed` is never set, and the person gets the unexplained blank green box that the fallback ("We couldn't load the map. View this location on Google Maps") was built to avoid.

### Steps to reproduce

1. Open a page with a map (e.g. any `/applications/:id`) with DevTools request blocking enabled for `/assets/application-*.js` (or block jsDelivr's d3/`bar_graph` on an authority page while letting Alpine through)
2. Alpine boots, `x-init` throws `ReferenceError: initialiseBasicMapWithMarker is not defined`
3. Blank map area, no fallback message

### What did you expect to happen?

Either the fallback should engage, or the race should be impossible. Options, cheapest first:

- Make the `x-init` expressions defensive so the existing fallback catches this case, e.g. `typeof initialiseBasicMapWithMarker === 'function' ? initialiseBasicMapWithMarker(...) .catch(...) : mapFailed = true`
- Or eliminate the two-origin race by serving Alpine from our own asset pipeline alongside `application.js`, so they fail or succeed together

### Browser, device, or OS (if relevant)

Chrome on Windows for the map errors; the `barGraph` events so far are only a crawler, but the mechanism is identical for real browsers.

### Screenshots or error messages

```
ReferenceError: initialisePano is not defined
ReferenceError: initialiseBasicMapWithMarker is not defined
ReferenceError: barGraph is not defined
```

### Anything else?

Related to #2185: that issue covers the Google Maps CDN itself failing to load (nothing we can fix), while this one is the failure mode we *can* fix — our fallback UI not engaging when scripts go missing. Fixing this one properly would also give people a useful fallback in most of the #2185 scenarios.

---

Raised from Sentry with AI assistance (OpenCode, model `anthropic.claude-fable-5`).

benrfairless:
## Description

Adds `runWithFallback`, an inline unbundled helper in the head, and routes every Alpine `x-init` expression that depends on a separately loaded script through it. Passing the call as a thunk means a missing function throws inside the helper, where it can be caught, rather than while Alpine evaluates the expression, where it cannot.

Two commits:

1. The six map and graph expressions, which is #2193 proper.
2. `window.confetti` and `window.FloatingUIDOM`, which fail the same way through `window.` and so throw `TypeError` rather than `ReferenceError`. Not in #2193's Sentry data, folded in because splitting them would produce a near-identical second review.

Only a synchronous throw is reported to Sentry, once per page load, at warning level. A rejected promise means the script loaded and then failed at runtime, which in practice is `maps.googleapis.com` behind a content blocker: unfixable by us, and reporting it would restore the noise #2185 removed.

## Motivation and Context

Fixes #2193. Also closes Sentry issues `PLANNING-ALERTS-1D` and `PLANNING-ALERTS-11`.

The fallback added in #2185 hung off `.catch` on the returned promise, so it could only ever catch a rejection. When `application.js` or `bar_graph.js` fails to arrive, the function itself is undefined and the expression throws synchronously, `mapFailed` is never set, and the person gets the blank green box the fallback exists to prevent. The map error was recorded against a release that already had the #2185 fix, from a real Chrome user on Windows.

Two things this deliberately does not do, both raised separately:

- The empty pink chart box on the authority page is left alone. Hiding it is a visual design change and would make the "Authority" Percy snapshot depend on cdnjs being reachable from CI. Noted in #2194.
- The blank box that persists when Alpine itself never loads is untouched: `[x-cloak]` is static CSS, so the fallback stays hidden with no JavaScript to reveal it. That is #2194, and the underlying CDN exposure is #2195.

## How Has This Been Tested?

Full suite, Sorbet and lint all pass locally in Docker: `bin/rake` (708 examples, 0 failures), `bin/rake ci:type`, `bin/rake ci:lint`.

New coverage:

- Component specs for the three map components assert the expression reaches the fallback when the function is missing entirely.
- `spec/features/script_fallback_spec.rb` drives the helper in a real browser: missing function, runtime rejection, successful call left alone, and resolving to the fallback's return value so `MapWithRadiusComponent` can await it. `window.Sentry` is absent throughout in test, so these also cover the helper working without it.
- The authority page spec asserts both graphs route through the helper.

The real failure, a blocked asset request, is not reproducible here: the javascript driver is headless Firefox, which gives no way to block a request. The specs drive the helper directly instead. Nothing proves Alpine and the helper are wired together end to end, only that each half behaves.

## Types of Changes

- [ ] Documentation

## Checklist:

---

Written with AI assistance (OpenCode, model `anthropic/claude-opus-5`). Every commit carries an `Assisted-by` trailer.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -17,6 +17,8 @@ require "capistrano/honeybadger"
 require "capistrano/aws"
 require "capistrano/puma"
 require "capistrano/tagging3"
+require "net/ssh/proxy/command"
+
 install_plugin Capistrano::Puma
 install_plugin Capistrano::Puma::Systemd
 
--- a/README.md
+++ b/README.md
@@ -204,6 +204,16 @@ Credentials are separate from that file and must never be committed. v3 reads yo
 
 If the Sentry CLI is missing or unauthenticated the deploy still succeeds — the hook prints a warning and skips recording the release, so set it up before your next production deploy.
 
+### whatismyip trust-boundary check
+
+`GET /whatismyip` reports the IP Rails sees for the request, to aid checking Cloudflare/ALB proxying with a one-line curl rather than a real sign-in or a log dig. It returns the IP, or the IP plus ` FAIL` if that IP falls within Cloudflare's own published ranges, a sign the trust boundary isn't rewriting it to the real visitor IP. If Cloudflare's published ranges can't be fetched or look implausibly short, it returns the IP plus ` UNABLE TO CHECK` rather than risk reporting a false pass.
+
+Off by default; turn it on for a server with the `provide_whatismyip` Flipper feature flag.
+
+```sh
+curl https://www.planningalerts.org.au/whatismyip
+```
+
 ## Upgrading Ruby in production
 
 Upgrading Ruby in production is an unbelievably painful process right now. I'm sorry. Let's make it simpler
--- a/app/components/authority_map_component.html.erb
+++ b/app/components/authority_map_component.html.erb
@@ -1,0 +1,7 @@
+<%= tag.div class: "w-full my-8 h-80 bg-google-maps-green",
+            "x-data" => "{ mapFailed: false }",
+            "x-init" => "runWithFallback(() => initialiseAuthorityMap($el, #{map_params_json}), () => mapFailed = true)" do %>
+  <div x-show="mapFailed" x-cloak class="flex h-full w-full items-center justify-center p-8 text-center text-xl text-navy">
+    <p>We couldn't load the map. <%= helpers.pa_link_to "View this area on Google Maps", google_maps_url %></p>
+  </div>
+<% end %>
--- a/app/components/authority_map_component.rb
+++ b/app/components/authority_map_component.rb
@@ -1,0 +1,32 @@
+# typed: strict
+# frozen_string_literal: true
+
+class AuthorityMapComponent < ViewComponent::Base
+  extend T::Sig
+
+  sig { params(authority: Authority).void }
+  def initialize(authority:)
+    super()
+    @authority = authority
+  end
+
+  # Used as a fallback when the map javascript can't be loaded, for example
+  # because a content blocker is blocking maps.googleapis.com. An authority is
+  # a boundary rather than a point, so search Google Maps by name instead of
+  # linking to a lat/lng.
+  sig { returns(String) }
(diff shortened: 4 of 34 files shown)


ianheggie-oaf (approved):
Approved - LGTM at a quick read of description and a glance at code

PR #2189, opened 2026-08-19
Show a fallback when Google Maps fails to load

benrfairless:
## Description

Sentry browser monitoring (added in #2049) showed the Google Maps javascript failing to load for some people, mostly on Mobile Safari, which is consistent with content blockers, Lockdown Mode or flaky connections cutting off `maps.googleapis.com`. When that happened the person got an unexplained blank space where the map should have been, and the unhandled promise rejection was reported to Sentry.

This catches the failure everywhere a map is initialised:

- The map and streetview on the application page and the map on the alert edit page now explain that the map couldn't be loaded and link to the location on Google Maps instead (a plain link, so it works even when the javascript API is blocked).
- The alert radius watcher only updates the circle when the map actually loaded, so it can't raise a follow-on `TypeError` (likely the source of [PLANNING-ALERTS-N](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-N), [PLANNING-ALERTS-G](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-G) and [PLANNING-ALERTS-E](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-E)).
- The address autocomplete, the authority map and the admin geocoding map catch the failure quietly. The address field still works as a plain text field without autocomplete.

Part of #2185. The remaining step, once this is deployed, is archiving the Sentry issues ([PLANNING-ALERTS-8](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-8), [PLANNING-ALERTS-7](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-7), [PLANNING-ALERTS-A](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-A), [PLANNING-ALERTS-C](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-C)) so they stop competing for attention.

## Motivation and Context

Fixes the first half of #2185: the page should not leave an unexplained blank space where the map was going to be.

## How Has This Been Tested?

- [x] Checked affected area manually on my own / staging system
- [x] Ran automated tests on my own system
- [ ] Confirmed it passed the GitHub actions tests

Written test-first: new component specs for `MapComponent`, `StreetviewComponent` and `MapWithRadiusComponent` cover the fallback message, the Google Maps link and the fallback staying hidden until the map fails.

[...]

Issue #2187, opened 2026-08-19
Bug: Sentry reports crawler and browser extension errors as if they were ours


benrfairless:
### Which service is this about?

Planning Alerts

### What happened?

Browser monitoring went live on 18 August (718cb0ca3) and the Sentry project immediately filled with frontend errors that our code does not cause. All 15 unresolved issues were first seen within 20 hours, and most of them are not actionable.

The loudest example: [PLANNING-ALERTS-8](https://oaf-org-au.sentry.io/issues/PLANNING-ALERTS-8) shows 428 events across "183 users". 385 of those events are **Baiduspider-render**, a search engine crawler running our JavaScript and tripping over a Google Maps failure. The "183 users" are crawler IPs.

Sorting the 15 unresolved issues by what is actually causing them:

| Cause | Sentry issues |
| --- | --- |
| Crawlers running our JavaScript | -8 (90% of its volume), -D |
| Browser extensions and in-app webview bridges | -M (MetaMask), -J (contrast extension), -F (`webkit.messageHandlers`), -K (Android webview bridge) |
| Microsoft Outlook scanning links in our alert emails | -9 |
| Unactionable, no usable detail | -B (titled `&lt;unknown&gt;`), -G, -E, -N (minified third-party frames) |
| **Genuine, and worth fixing** | -7, -A, -C and the real-browser tail of -8 → #2185; -H → #2186 |

So of 15 issues competing for attention, two things were worth knowing about, and both were buried.

The cause is straightforward: `Sentry.init` in `app/views/application/_sentry.html.erb` sets no `ignoreErrors` and no `denyUrls`, and the loader script is served to every request including crawlers.

### Steps to reproduce

1. Open https://oaf-org-au.sentry.io/issues/?project=planning-alerts&query=is%3Aunresolved
2. Note that the top issue by volume is a search engine crawler, and that most of the rest are browser extensions

### What did you expect to happen?

The unresolved list should only contain errors we can act on. Specifically:

- Don't load the browser SDK at all for crawlers. They cannot be helped and they outnumber real people by roughly ten to one on these errors.
- Filter the known third-party patterns via `ignoreErrors`, and extension schemes via `denyUrls`.
- Filter *only* other people's code. Anything our own JavaScript causes belongs in a fix, not in an ignore list. In particular the Google Maps failures (#2185) must stay visible, because real people are affected, and the clipboard error (#2186) is our own bug.

### Browser, device, or OS (if relevant)

[...]

PR #2198, opened 2026-08-24
Filter Google Maps chunk noise from Sentry

benrfairless:
## Description

Adds two entries to the browser Sentry `ignoreErrors` list in `_sentry.html.erb`:

- `/^Could not load "\w+"\.$/` - chunk failures inside Google Maps' own loader ("onion", "map", "util", "streetview", "marker"): the main script arrived but a content blocker or flaky connection cut off an internal module. These reject promises created inside Google's code that we can never attach a handler to; our own `importLibrary` calls already catch their side of the failure and show people a fallback.
- `"this.dj.close is not a function"` - Google Maps Street View calling `HTMLDialogElement.close()` on browsers without `<dialog>` support (old Firefox/Safari).

The comment block now also records the policy behind two deliberate non-entries: no `denyUrls` for the Maps host (it would also hide errors caused by our own misuse of the Maps API), and no WebKit `"Load failed"` (that is the message for any failed fetch, including our own boundary GeoJSON requests). A caveat worth knowing: the `dj` identifier comes from Google's minifier and will rotate in a future Maps build, at which point this entry fails open - the errors reappear rather than being silently hidden, which is the right failure direction for a filter.

## Motivation and Context

Part of #2185. Since #2189 deployed, the loader-level failures ("The Google Maps JavaScript API could not load.") have stopped appearing in Sentry, but the chunk failures still account for the bulk of unresolved browser events (PLANNING-ALERTS-V, -C, -W, -A) despite there being no action available to us. Filtering them follows the approach agreed in #2187/#2188: only other people's code is filtered; anything our own JavaScript causes belongs in a fix. Once this deploys, the existing Sentry issues in this cluster will be archived.

## How Has This Been Tested?

The partial only renders in production for non-crawler user agents, so this is not exercisable locally; the regex was checked against the exact messages recorded in Sentry. Full local run of `bin/rake`, `bin/rake ci:type` and `bin/rake ci:lint` via Docker Compose: 704 examples, 0 failures; all linters clean.

## Types of Changes

- [ ] Documentation

## Checklist:

---

This change was made with AI assistance (OpenCode, model `anthropic.claude-fable-5`), as disclosed in the commit trailers.

### DIFF
--- a/app/views/application/_sentry.html.erb
+++ b/app/views/application/_sentry.html.erb
@@ -16,6 +16,12 @@
         // link scanners rather than by our own code. We cannot fix any of these
         // and they drown out real problems. Anything our own JavaScript causes
         // belongs in a fix, not in this list.
+        //
+        // Google Maps noise is filtered here by message, deliberately not by a
+        // denyUrls entry for maps.googleapis.com, which would also hide errors
+        // caused by our own misuse of the Maps API. Don't be tempted to add
+        // WebKit's generic "Load failed" either: it is the message for any
+        // failed fetch, including our own boundary GeoJSON requests.
         ignoreErrors: [
           // Microsoft Outlook and Office scanning links in our alert emails
           "Object Not Found Matching Id",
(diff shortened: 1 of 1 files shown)


ianheggie-oaf (approved):
Approved - makes sense to me

Issue #2159, opened 2026-08-17
TODO audit: extract shared components and deduplicate partials


benrfairless:
### Which service is this about?

Planning Alerts

### What problem are you trying to solve?

The audit found 33 TODOs about repeated markup/styling that should be extracted into shared ViewComponents or partials (plus closely-related component API questions). The duplication makes styling drift likely.

### What would you like to happen?

Extract the repeated patterns below into shared components/partials and resolve the component API questions.

- [ ] `app/components/heading_component.rb:15` — Perhaps we should allow the size override to work by saying we want an h1 heading but with the styling of an h2?
  <sub>initialize still takes raw size/weight/color/font strings (lines 16-37); no tag-style-of-tag API exists</sub>
- [ ] `app/components/info_popup_component.html.erb:49` — Remove duplication between here and pa_link_classes view helper
  <sub>Focus/hover classes on button (line 50) duplicate pa_link_classes (app/helpers/application_helper.rb:15-24)</sub>
- [ ] `app/components/link_block_component.rb:18` — Remove duplication between this and the link styles defined in pa_link_to
  <sub>linkify hardcodes "font-bold underline text-fuchsia group-hover:text-fuchsia-darker" (line 19), duplicating pa_link_classes/pa_link_to in app/helpers/application_helper.rb:15-31</sub>
- [ ] `app/controllers/alerts/registrations_controller.rb:40` — Get rid of duplication
  <sub>after_inactive_sign_up_path_for identically duplicated in users/registrations_controller.rb:82-85</sub>
- [ ] `app/controllers/api_keys_controller.rb:25` — Extract this
  <sub>Trial-key creation with limits/expiry inlined in controller (api_keys_controller.rb:26); no model/service method exists (grepped: only creation site in app/)</sub>
- [ ] `app/controllers/comments_controller.rb:59` — This seems to have a lot repeated from Application#show
  <sub>create failure path rebuilds @alert/@comments and renders "applications/show" (comments_controller.rb:59-66), duplicating applications_controller#show:117-136</sub>
- [ ] `app/controllers/documentation_controller.rb:31` — Move this action to ContactMessagesController#new
  <sub>contact action still in DocumentationController (documentation_controller.rb:33-37); contact_messages routes only have create/thank_you (routes.rb:197-201); ~7 views still link documentation_contact_path</sub>
- [ ] `app/helpers/application_helper.rb:24` — Generalise to support all the variants
[...]

PR #2204, opened 2026-08-31
Record the IP people sign up from and show it in the admin

### LINKED ISSUE #1682: Check if spam accounts being created (and stop them)
There have recently at least been a spate of unconfirmed user accounts created with email addresses that do not have any associated activity on the site.  Are they spam accounts? 

I did a little poking around but got stuck.

If the signups aren't genuine then at the least there are people receiving confirmation emails that they don't want. We could be in for more spammy comments. It's hard for me to easily explore the scale of the issue from my level of access and knowledge, so I don't really know what to do next for the best. Do you have any thoughts, guidance, or pointers for me to be able to investigate further @mlandauer?

benrfairless:
## Description

Adds `alerts.signup_ip` and `users.signup_ip`, recorded from `request.remote_ip` on every path that creates an alert or an account, shown in the admin backend, and cleared automatically after 90 days.

Five places record it. Two are Devise registrations controllers, where the IP is set by overriding `build_resource`: Devise's `create` is `build_resource(sign_up_params)`, then `resource.save`, then `yield resource`, so the `super do ... end` blocks these controllers already use run *after* the save and are the wrong hook. `build_resource` runs before the insert, so no second write is needed. Three are the `Alert.new` calls that actually save a record, in `AlertsController#create`, `Alerts::RegistrationsController#create` and `Alerts::SessionsController#create`. The many other `Alert.new` calls build unsaved objects to back a form or preview an email and are left alone.

The `build_resource` override is deliberately duplicated across the two registrations controllers rather than extracted into a concern. This app has no `app/controllers/concerns/` directory and never uses Sorbet's `requires_ancestor`, so a shared module would mean introducing two new patterns for six lines, and putting `super` inside a module whose ancestors Sorbet cannot see. Those two controllers already duplicate `after_inactive_sign_up_path_for` under a `TODO: #2159 Get rid of duplication` comment, so this copies that comment and gets swept up when #2159 is done properly. Setting the IP explicitly at each call site, rather than through an `ActiveSupport::CurrentAttributes` global with model `before_create` hooks, is also deliberate: the thing the implicit version buys you is a future collection point picking the IP up automatically, and for personal information that is a liability rather than a feature. Each collection point should be a line someone had to write, because each one needs covering by the APP 5 notice.

In the admin, `signup_ip: Field::String` in `ATTRIBUTE_TYPES` is what makes it searchable, because `Administrate::BaseDashboard#search_attributes` selects on `searchable?` rather than reading the index list. So `/admin/alerts?search=203.0.113` works with no extra code. It is also an alerts index column, since several signups from one place is a pattern you notice by scanning and sorting a list. It is **not** added to either dashboard's `FORM_ATTRIBUTES`: on `UserDashboard` that list doubles as administrate's strong-params allowlist on the live edit route, so adding it would let an admin overwrite the recorded value, and on `AlertDashboard` the list is currently dead code but would become editable if the edit routes ever came back.

The column is a `string` rather than a Postgres `inet`, to match `users.current_sign_in_ip` and `users.last_sign_in_ip`, which Devise already writes as strings. All three IP columns then read, query and render the same way. To be clear about what that trade is: `inet` would work fine with the admin search, because administrate builds its query as `LOWER(CAST(column AS CHAR(256))) LIKE ?` and the cast makes `inet` valid (I checked that against Postgres 15 rather than assuming). What `inet` would buy is subnet containment queries, which nothing needs yet, and it is an easy migration if that changes. There is no index, because `LIKE '%term%'` cannot use a btree index anyway and the only other query is the expiry job's, which after its first run matches a thin slice.

`ExpireSignupIpsJob` clears anything older than `RETENTION_PERIOD` (90 days) daily from `config/cron.yml`, at 4am Australia/Sydney, which is clear of every existing entry. It uses `update_all` in batches: batching bounds the row lock so a first run over a backlog cannot sit on a lock while people are signing up, and `update_all` skips callbacks and does not bump `updated_at`. Neither `Alert` nor `User` has `has_paper_trail`, so clearing the column genuinely removes the value rather than leaving a copy in `versions`.

One thing worth knowing when reading the recorded values: this app does not set `config.action_dispatch.trusted_proxies`, so treat the field as a signal for clustering rather than as evidence. That is written down in `doc/data_retention_policy.md`, along with the fact that database backups still hold the value for as long as we keep the backups, and the inconsistency that `users.current_sign_in_ip`, `users.last_sign_in_ip` and the IPs `LogApiCallJob` sends to Elasticsearch all predate this work and have no retention rule at all. Both are worth separate issues.

## Motivation and Context

Closes #1682, which I reopened, since it was closed by the stale bot rather than resolved. There was no way to tell a genuine signup from a bulk one from the admin backend, which is exactly what @katska hit: "it's hard for me to easily explore the scale of the issue from my level of access". The only IP addresses we store are `users.current_sign_in_ip` and `users.last_sign_in_ip`, which Devise `:trackable` writes on sign-in. Someone signing up does not sign in during that request, so those columns are empty for a brand new account, and nothing at all was recorded against a new alert. The data simply was not there to look at.

Because an IP address is personal information under the Australian Privacy Principles, the retention side is part of the change rather than a follow-up.

This only works if `request.remote_ip` is the real visitor address. PlanningAlerts is not behind the Cloudflare proxy (every web record in `terraform/planningalerts/dns.tf` in the infrastructure repo is `proxied = false`), so traffic goes straight to the load balancer and there is no Cloudflare edge address to unpick. That is why the `cloudflare_realip` Ansible role is wired into `righttoknow` and not `planningalerts`. The `/whatismyip` endpoint from #2203 is how to confirm it against production, and that check is unticked below because it needs a production host.

## How Has This Been Tested?

All three CI jobs (`test`, `type_check`, `lint`) are green on `2a98259`, and were also green on the first commit. That matters more than usual here: `spec/features` could not be run in the environment I was working in, because those specs call `percy_snapshot` and drive a real browser, both of which reach `cdn.jsdelivr.net` and `fonts.gstatic.com`, and outbound access to those hosts was blocked, so they hung rather than failed. GitHub Actions is what confirms them, and it has.

Locally, before pushing: `bin/rake ci:type` and `bin/rake ci:lint` both clean, including `tapioca dsl --verify --environment=test` and `check-shims` (the shim needed a `build_resource` entry) and Brakeman with no warnings. The regenerated DSL RBIs are committed, generated with `--environment=test` to match what CI verifies, so they only add the `signup_ip` accessors. The spec suite ran as `bundle exec rspec spec --exclude-pattern "features/**/*_spec.rb"`: 617 examples, 0 failures, 5 pending. The one new feature spec, `spec/features/admin_views_signup_ip_spec.rb`, is rack_test only with no Percy call, so it did run locally and passes.

Manually against a real PostGIS database, on top of the specs:

- Searching the admin alerts index for an IP returns the alert (`Administrate::Search` over real Postgres, not a stub), which is the check that matters for the column type choice above.
- `AlertDashboard#search_attributes` and `UserDashboard#search_attributes` both include `signup_ip`, and neither dashboard's `FORM_ATTRIBUTES` does, so it is searchable and not editable.
- Backdating a record's `created_at` past 90 days and running `ExpireSignupIpsJob.new.perform` clears only that record's IP, leaves a recent one alone, and leaves the rest of the expired row (address, email, subscription state) untouched.

Not done, and it should be done before this merges: the `/whatismyip` check against a production host, since that needs production access. Turn on the `provide_whatismyip` Flipper flag, `curl https://www.planningalerts.org.au/whatismyip`, confirm it returns a bare IP matching your own rather than one with ` FAIL` appended, and turn the flag back off.

None. `spec/features/admin_views_signup_ip_spec.rb` asserts the IP appears on both the alerts index and the alert show page, but the sandbox I was working in cannot run a real browser (see above), so I could not take one. Worth adding a shot of the alerts index column when someone runs this locally.

## Types of Changes

- [x] Documentation

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

---

**A note on SonarCloud:** its first run flagged 7 new issues. Five were duplicated string literals in the new specs (`ruby:S1192`) and are fixed in `2a98259`. The two that remain are `ruby:S1135` "complete the task associated to this TODO comment", on the `TODO: #2159` markers added to the two registrations controllers. Those are deliberate and I have left them: the same two controllers already carry that exact marker on the other method they duplicate, and keeping it is what makes the duplication findable when someone picks up #2159. Both are INFO severity and the quality gate passes.

**AI assistance:** written with Claude Code (model `claude-opus-5`). Each commit carries an `Assisted-by:` trailer. The commits deliberately carry **no `Signed-off-by`** trailer, because the DCO is a certification only a person can make and the org `AGENTS.md` says never to add one on an agent's behalf. This needs re-signing by a human before it can be merged.

## Summary by CodeRabbit

- **New Features**
  - Signup IP addresses are recorded for new accounts and alerts.
  - Administrators can view and search signup IP addresses in account and alert details.
  - Signup IP addresses are automatically removed after 90 days.

- **Documentation**
  - Added documentation covering signup IP collection, visibility, retention, and privacy considerations.

- **Tests**
  - Added coverage for recording, displaying, retaining, and expiring signup IP addresses.


Br3nda:
@benrfairless this has git conflicts -- but also looks like gen AI is warning you about something at the top of the description

benrfairless:
@Br3nda I can't see any git conflicts. I've removed the warning at the top - this will be addressed in an upcoming piece of work.

benrfairless:
@coderabbitai full review

Br3nda:
is this reflected in our privacy statements?

benrfairless:
> is this reflected in our privacy statements?

It will be

Br3nda (approved):
confirmed. works as described

PR #2205, opened 2026-09-01
Add ALTCHA to public forms behind Flipper gates

benrfairless:
PlanningAlerts has almost no anti-bot protection. reCAPTCHA is wired into
exactly one place, the contact form, and it short-circuits to nothing when
the credential is absent. Every other unauthenticated form that writes a
record or sends email is unprotected. See #1682.

ALTCHA is a self-hosted, MIT-licensed alternative to a captcha: the browser
does a small amount of arithmetic instead of a person identifying traffic
lights. Nothing is sent to a third party, which is the main reason to prefer
it to reCAPTCHA.

Protects sign up on both routes, password reset, resend confirmation,
account activation, and the contact form. Comments, alerts and abuse reports
already need somebody signed in, so they are out of scope.

Two flags, both off by default, so nothing changes until they are turned on:

  altcha         show the widget and check the answer
  altcha_enforce turn a failed check into a rejection

With altcha on and altcha_enforce off we are in monitor mode: the outcome is
counted in Sentry and the form goes through anyway. That is how we find out
what enforcing would cost before anybody is turned away. Enforcing locks out
everyone without JavaScript, which is a product decision to take with the
monitor numbers in hand, not one to make in a pull request.

Because these forms are only ever seen by logged-out people, current_user is
no use as a Flipper actor. Visitor gives them a stable flipper_id from the
session so that percentage_of_actors answers the same on the request that
renders a form and the request that checks it. percentage_of_time would
reject people whose form was rendered without a widget.

The contact form is the one form that already had a check, so ALTCHA and
reCAPTCHA are alternatives there rather than both at once, and a percentage
gate runs the two side by side on a sticky split.

Some details worth knowing:

- The HMAC secrets are derived from secret_key_base rather than added as a
  credential, so the check is not silently absent in test the way reCAPTCHA
  is.
- The gem has no replay protection, so AltchaSolution is ours: one row per
  accepted answer, unique index, swept daily.
- The widget is vendored rather than loaded from a CDN. Every other
  third-party script here degrades to something cosmetic; a missing ALTCHA
  widget in enforce mode stops somebody signing up or asking us for help.
  The UMD build, so it needs no type="module" and survives Terser.
- Flipper's admins group raised on any actor without roles, which a Visitor
  is. Fixed in the initializer.

doc/altcha.md covers the rollout order and what to check before enforcing.
The Rack::Attack throttle on the challenge endpoint cannot be spec'd, since
spec_helper disables Rack::Attack.

Assisted-by: Claude Code


Br3nda (approved):
lgtm

PR #2221, opened 2026-09-11
Throttle page views by IP to blunt scraper bursts

ianheggie-oaf:
## What and why

A scraper walking application ids pushed a t3.medium app server to ~95% CPU for almost two hours. Add a `pages/ip` Rack::Attack throttle (2 req/s per IP) so a single IP can't do that again, excluding the `api` subdomain's own per-key throttle, static assets, and the load balancer's health check.

Serve a branded 429 page for this throttle specifically (the others keep their original plain-text response, in case anything relies on it), and add `THROTTLE_DIVISOR` (documented in the README) so the limits are reachable by hand while testing in development.

- Related to openaustralia/infrastructure#806

image

## Type of change

- [x] New feature

## How this was checked

Rack::Attack is disabled in the spec suite (see the existing comment in `config/initializers/rack_attack.rb`), so this was verified directly against the real app via `bin/rails runner`: a host-header-spoof bypass a `/code-review` pass found is fixed (a request can no longer dodge the throttle by sending `Host: api.planningalerts.org.au` while hitting a non-API page route), genuine API calls stay exempt, the branded 429 page only replaces the response for `pages/ip` and not the other throttles, and `THROTTLE_DIVISOR` produces the expected limits (including clamping to 1, never 0, at an extreme value).

I ran the tests via guard, and relied on GitHub actions to run the full `bin/rake ci:type`/`ci:lint`/spec suite

- [x] Checked the affected area on my own or a staging system (THROTTLE_DIVISOR=10)
- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed concerns (and rechecked)
- [x] GitHub Actions tests: likewise fixed so they pass or linked a follow-up issue / replied why not important directly on each comment (reviewer will mark resolved as part of review)

## Summary by CodeRabbit

* **New Features**
  * Added configurable request throttling to help protect pages and API endpoints from excessive traffic.
  * Added a dedicated, responsive page for HTTP 429 rate-limit responses, with helpful navigation links.
  * Added support for passing the throttle setting through local container environments.

* **Documentation**
  * Documented how to test throttling locally, including an example configuration that makes rate limits easier to trigger.
  * Clarified that normal development behaviour remains unchanged unless the setting is enabled.


benrfairless (commented):
Ran this through CodeRabbit and had a look myself. One real gap and a couple of small tidy-ups, all inline as suggestions. Nice to see the branded 429 page and the `THROTTLE_DIVISOR` hook for local testing.

config/initializers/rack_attack.rb
```diff
@@ -35,8 +40,41 @@
...
+  next if req.path.start_with?("/assets/") ||
+          req.path == "/health_check"
+
+  # Exclude what is matched for the api controller throttle above
+  path_info = begin
+    Rails.application.routes.recognize_path(req.url)
+  rescue StandardError
+    nil
+  end
+  next if path_info && path_info[:controller] == "api"
```
benrfairless: Requests to the API without a `key` currently escape both throttles. The `limit api requests` discriminator (line 20) returns `request.params["key"]`, which is `nil` when no key is supplied, and Rack::Attack skips a request whose discriminator is nil. This exclusion then skips the same request for `pages/ip`, so something like `/applications.json?lat=..&lng=..` with no key is counted by neither. Each of those still boots Rails and runs `require_api_key` before the 401, so a scraper hitting the API unkeyed stays completely unthrottled.

Only deferring to the API throttle when a key is actually present closes it:

```suggestion
  next if path_info && path_info[:controller] == "api" && req.params["key"]
```

config/initializers/rack_attack.rb
```diff
@@ -1,8 +1,13 @@
+# divide_limit divides by THROTTLE_DIVISOR if set, returning a minimum of 1
```
benrfairless: Trailing whitespace. `config/*` is excluded from RuboCop so it won't fail lint, just tidying.

```suggestion
# divide_limit divides by THROTTLE_DIVISOR if set, returning a minimum of 1
```

config/initializers/rack_attack.rb
```diff
@@ -35,8 +40,41 @@
...
-Rack::Attack.throttled_response_retry_after_header = true
+# 1,000 rpm (/applications/N) per t3.medium instance maxes out our CPU (95%).
+# 40 requests per 20 seconds is roughly 12% of max and should be long enough for natural clumping
+Rack::Attack.throttle("pages/ip", limit: divide_limit.call(40), period: 20.seconds) do |req|
+  next if req.path.start_with?("/assets/") ||
+          req.path == "/health_check"
+
+  # Exclude what is matched for the api controller throttle above
+  path_info = begin
+    Rails.application.routes.recognize_path(req.url)
```
benrfairless: Non-blocking: `recognize_path` now runs twice on every request (here and in the API throttle above). Route recognition isn't free and this is on the hot path for every page hit. Fine for now, but if it shows up in profiling, stashing the result in `req.env` on first use and reading it back in the second block would halve it.

ianheggie-oaf:
@coderabbitai please review again

config/initializers/rack_attack.rb
ianheggie-oaf: I have updated code to address this

benrfairless (approved):

PR #2240, opened 2026-09-19
Allow docker compose ports to be overridden per worktree

### LINKED ISSUE #2227: Feature: Support running tests across multiple worktrees in parallel

Planning Alerts

Investigating several open PRs side by side in separate git worktrees is useful (e.g. checking a batch of Dependabot bumps in parallel), but `docker-compose.yml` hard-codes the host ports for `web` (3000), `postfix` (2525), `postgres` (15432) and `mailcatcher` (1080). Only one worktree's Docker stack can be up at a time; a second worktree's `docker compose up`/`docker compose run` collides on those ports.

Parameterise those four host ports in `docker-compose.yml` with env-var defaults, e.g. `"${WEB_PORT:-3000}:3000"`, so the existing single-worktree workflow is unchanged by default, but a worktree that wants to run alongside another can drop a `.env` file (already how Compose auto-loads per-project overrides) picking different numbers. Docker Compose already scopes named volumes and the container network per project (which defaults to the directory name), so gems, Postgres data, etc. need no extra isolation once the ports are free to vary. Document the pattern briefly in the README.

- `docker-compose.override.yml`: doesn't work for this — Compose merges the `ports` list by concatenation, not replacement, so the original hard-coded port stays published alongside the override and the collision remains.
- Isolating the *host*-side gem install per worktree (`GEM_HOME`/`GEM_PATH` layering, or a `.bundle/config` path override): turned out to be unnecessary. The documented test workflow (`docker compose run web bin/rake ci:type` etc., per AGENTS.md) never runs `bundle install` on the host at all — only inside the `web` container, whose gems already live in a Compose-project-scoped volume. This only matters for someone who *also* runs host-side Ruby tooling (an IDE's RuboCop/Sorbet integration) against a worktree, which is a personal convenience rather than something the shared setup needs to solve.
- direnv-based `.envrc` per worktree: not viable as a team-wide mechanism since not everyone uses direnv, and it doesn't help on its own without an explicit per-worktree config file anyway (direnv only finds the nearest ancestor `.envrc`, so a shared one can't vary per worktree).

### Anything else?

Found while test-driving six worktrees for the currently-open Dependabot PRs (#2050, #2139, #2207, #2208, #2218, #2219) in parallel. As a side effect that investigation turned up a real, unrelated finding worth its own issue/comment: bumping `devise` to 5.0.4 (#2050) transitively bumps `connection_pool` to 3.0.2, which breaks Rails boot entirely (`ArgumentError` in `config/initializers/rack_attack.rb` via `ActiveSupport::Cache::RedisCacheStore`), independent of a separate Sorbet/RDoc payload-conflict issue from the same bump.

ianheggie-oaf:
## What and why

- Running a second worktree's `docker compose` stack fails with "port is already allocated", because `web`, `postfix`, `postgres` and `mailcatcher` each bind a fixed host port. Those four are now taken from `WEB_PORT`, `POSTFIX_PORT`, `POSTGRES_PORT` and `MAILCATCHER_PORT`, so a worktree can put them somewhere else in a `.env` file.
- Only the host side of each mapping changes. Inside the compose network the ports are as before, so no in-container configuration needs a per-worktree value.
- `WEB_PORT` also feeds `config/environments/development.rb` and `api_port` in `app/helpers/applications_helper.rb`, which both hardcoded `localhost:3000`. Without that, a confirmation link from the worktree on 3001 opens the worktree on 3000, where the token does not exist, and the API example URLs point at the wrong stack.
- `Alternative web port` in `README.md` used a `docker-compose.override.yml` `ports` block for a related purpose. Compose appends those mappings rather than replacing them, so it left 3000 published and collided across worktrees anyway. It now uses `WEB_PORT` instead.
- Fixes #2227

## Type of change

- [x] Chore
- [x] Documentation

## How this was checked

Ran the full suite and both static gates in a container against this branch. `docker compose run` does not publish ports, so the collision only ever affected `docker compose up`.

- [ ] Checked the affected area on my own or a staging system
- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues above, or noted why they are not important above

Results: `bin/rake` 849 examples, 0 failures, 6 pending (pre-existing accessibility pending specs). `bin/rake ci:lint` and `bin/rake ci:type` both clean. Two stacks have not yet been brought up side by side end to end, so that box is unticked.

The local test database needed `bin/rake db:test:prepare` first, because it predated the `signup_ip` migration from main. CI builds its database fresh, so that is a local environment matter only.

Assisted-by: Claude Code:claude-opus-5


ianheggie-oaf:
@benrfairless - this is not on the critical path, but was used for previous parallel checks I did, so I have wrapped it up in a PR for review as I found it very useful.

Br3nda (approved):

Issue #2239, opened 2026-09-18
Feature: Password-protect www-idle.planningalerts.org.au with HTTP basic auth


ianheggie-oaf:
### Which service is this about?

Planning Alerts

### What problem are you trying to solve?

Protect www-idle.planningalerts.org.au from bots and casual visitors while that colour is idle and awaiting cutover, as part of the work to get PlanningAlerts behind the Cloudflare proxy. api-idle must not be affected.

Originally tracked as openaustralia/infrastructure#838. That issue proposed a Cloudflare WAF rule; this issue replaces it with an app-side approach for the reasons below.

### What would you like to happen?

Add it as Rack middleware running inside Puma, in this repo, rather than as a Cloudflare WAF rule or at the ALB:

- An `IdleBasicAuth` middleware class alongside the existing `PlausibleProxy` in `config/application.rb` (that file is excluded from Sorbet/RuboCop, so it avoids the `# typed: strict` requirement that applies everywhere under `app/` and `lib/`).
- Registered with `config.middleware.insert_before 0`, not `.use`. Rails serves `public/` directly (`config.public_file_server.enabled = true`, no nginx in front), so appending to the end of the stack would let static files through unauthenticated.
- Gated on the Host header prefix (`www-idle`), so api-idle is untouched and the ALB health check (which hits `/health_check` with the target IP/port as Host) isn't challenged. It also follows the colour automatically at cutover, with no coordination needed.
- Credentials via `Rails.application.credentials.dig(:idle_basic_auth, ...)`, compared with `ActiveSupport::SecurityUtils.secure_compare`, matching the existing credentials-access style in the app.
- A request spec at `spec/requests/idle_basic_auth_spec.rb`, matching `spec/requests/plausible_proxy_spec.rb`.

### Alternatives you've considered

- **Cloudflare WAF rule** (the original approach, infrastructure#844): collides with the in-progress terraform import of the planningalerts zone's `http_request_firewall_custom` ruleset (PR #788). Cloudflare allows only one entrypoint ruleset per zone per phase, so a second one would fight it on every apply. That stack is also mid-migration from the v4 to v5 Cloudflare terraform provider, which changes the ruleset syntax. Separately, Cloudflare's WAF block action can't send a `WWW-Authenticate` header without an Enterprise plan, so it would need credentials stuffed into the URL rather than a normal browser login prompt.
- **ALB-level auth**: ruled out, as it means Cognito/OIDC, which costs extra.

### Anything else?

[...]

PR #2241, opened 2026-09-19
Fix RAILS_ENV=production crashing outside a Capistrano deploy

ianheggie-oaf:
## What and why

Running rails console for production fails when you are not in a release directory created by capistrano.

- `config/initializers/app_version.rb` assumed the `REVISION` file created by capistrano exists whenever `Rails.env.production?`.

Before:

```
$ RAILS_ENV=production bin/rails c
E, [2026-09-19T21:37:02.933561 #461503] ERROR -- : [SKYLIGHT] [6.0.4] could not load config file; msg=No such file or directory @ rb_sysopen - REVISION; disabling Skylight agent
.../config/initializers/app_version.rb:5:in `read': No such file or directory @ rb_sysopen - .../REVISION (Errno::ENOENT)
```

Notes:

- `config/skylight.yml` reads `REVISION` unconditionally too, and the line
  above shows it. That one is not fatal: Skylight catches it, logs, and
  disables its agent. Left alone, because a local console has no business
  reporting to Skylight anyway, and on a deployed server the file exists so
  the warning does not appear. Worth knowing it is there if anyone widens this
  later.
- Found while working on #2239, but not a fix for it, so this closes nothing.
- No spec. The behaviour only shows up under `RAILS_ENV=production` with
  initializers running, which the suite does not exercise, and the feature
  spec that touches the about page already stubs `APP_VERSION`.

## Type of change

- [x] Bug fix

## How this was checked

- [x] Checked the affected area on my own or a staging system
- [ ] Ran the automated tests
- [x] GitHub Actions tests: likewise fixed so they pass or linked a follow-up issue / replied why not important directly on each comment (reviewer will mark resolved as part of review)

Ran `RAILS_ENV=production bin/rails c` in a local checkout with no `REVISION`
file, directly rather than through Docker. It raised the error above before
this change and gets to a prompt after it. Draft until the CI gates have run.

Assisted-by: Claude Code:claude-opus-5 (this description only, the code change
is my own)


config/initializers/app_version.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 unless defined?(APP_VERSION)
-  APP_VERSION = if Rails.env.production?
-                  File.read(File.join(Rails.root, "REVISION"))[0..6]
+  revision_path = File.join(Rails.root, "REVISION")
+  APP_VERSION = if Rails.env.production? && File.size?(revision_path)
```
Br3nda, line 5: do we need to check there's no newline in the file? that'll end up in `APP_VERSION`

Br3nda (commented):
question about newline

config/initializers/app_version.rb
ianheggie-oaf: @Br3nda - I don't believe so. Do you think that is a likely cap failure case? And if it is, it fails the same way as it did before.

Br3nda: sometimes files have that, sometimes they dont'

Br3nda: okay, if you're sure that won't get in commands where the new line interferes with it

Br3nda (approved):

PR #2242, opened 2026-09-19
Password-protect www-idle with HTTP basic auth

### LINKED ISSUE #2239: Feature: Password-protect www-idle.planningalerts.org.au with HTTP basic auth

ianheggie-oaf:
## What and why

`www-idle.planningalerts.org.au` is whichever environment is idle and waiting for cutover. This puts HTTP basic auth in front of it so bots and casual visitors can't reach it, while leaving the live environment and api-idle alone.

- Resolves #2239. Replaces the Cloudflare WAF approach in openaustralia/infrastructure#844, which collides with the in-progress terraform import of that zone's `http_request_firewall_custom` ruleset. ALB auth was the other option and would mean paying for Cognito.
- Gated on the Host header rather than on anything machine-specific, so api-idle is untouched, the ALB health check isn't challenged (it sends the target's IP and port as the Host), and the protection follows the environment at cutover with no coordination needed.
- Registered with `insert_before 0`. There's no nginx in front of Rails here, so anything later in the stack would sit behind `ActionDispatch::Static` and serve `public/` unchallenged.
- Credentials come from `Rails.application.credentials`, falling back to `IDLE_BASIC_AUTH_USERNAME` and `IDLE_BASIC_AUTH_PASSWORD` so it can be exercised in development, where we have no credentials key. With neither set it challenges everything rather than failing open.

Not fixed here: Cloudflare can serve cached static assets without them reaching Puma, so CSS, JS and fonts can leak past the challenge. No personal data is in any of them. The class comment explains why, and sets out the two ways to close it if we decide to.

Stacked on #2241, which is stacked on #2240, so until those merge this diff also shows their changes. The files belonging to this PR are `app/middleware/idle_basic_auth.rb`, `config/initializers/idle_basic_auth.rb`, `config/application.rb`, `spec/requests/idle_basic_auth_spec.rb`, the README note and the production credentials.

## Type of change

- [x] New feature

## How this was checked

18 request specs, written before the middleware existed and confirmed failing against an empty stack first. They cover the challenge itself, wrong username and wrong password separately, a static file served through `ActionDispatch::Static`, a spoofed `X-Forwarded-Host`, host case and a host carrying a port, the live host and api-idle going untouched, an IP address host standing in for the ALB health check, the environment-variable fallback, and fail-closed behaviour with nothing configured.

Ran the full CI set locally: `bin/rake` (867 examples, 0 failures, 6 pre-existing pending), `bin/rake ci:type` and `bin/rake ci:lint` all clean. `bin/rails middleware` confirms `IdleBasicAuth` sits first in the stack, ahead of `ActionDispatch::HostAuthorization` and `ActionDispatch::Static`.

- [x] Checked the affected area on my own or a staging system - Deployed and tested on <https://www-idle.planningalerts.org.au>
- [x] Ran the automated tests

Assisted-by: Claude Code:claude-opus-5


app/middleware/idle_basic_auth.rb
```diff
@@ -0,0 +1,157 @@
...
+# Enterprise) and the response doesn't say public, s-maxage or must-revalidate.
+# Ours says public, so that protection doesn't apply, and signing in to load a
+# page can put the CSS, JS, fonts and images behind it into Cloudflare's cache.
+# Accepted rather than fixed: asset paths carry a content digest, so a bot or a
+# curious human has no way to know what to ask for if the file has changed, and
+# if it hasn't, the path they could guess from the live environment returns the
+# same bytes live already serves.
+#
+# Two ways to close it if we want to: a cache-bypass rule for www-idle in
+# Cloudflare, or set "Cache-Control: no-store" here on what we let through,
```
Br3nda, line 82: not sure we need so much commentary on this file.... 🤷‍♀️

Br3nda (approved):

app/middleware/idle_basic_auth.rb
ianheggie-oaf: Useful for future me if we start caring about it being like RTK staging is protected

PR #2245, opened 2026-09-19
Feature/2082 dependabot tapioca workflow

### LINKED ISSUE #2082: Tapioca and sorbet type checks take more effort to maintain than the benefit they give
**Describe the bug**

I have spent 6 hours to finally track down how to fix the sorbet errors whilst updating gems for security patches.

This is frustrating and wasteful of resources, way past the value we receive from using typing on our project.

I feel we would get better value from making sure we have good test coverage.

**To Reproduce**

1. Apply security updates from dependabot
2. Sometimes you will get non obvious errors
3. It will always take time to run up the docker environment to do all the type checking and updates

**Requested change:**

Technical spike to see if changing to RBS would be the better choice, or completely removing typing.

**Expected behavior**

We shouldn't have to fight the tooling each time we update gems for security patches.

Clobbering the gem cache and rebuilding web and then rerunning the type update  is a waste of time!

**Screenshots/Output**

```
+ docker compose run web bin/tapioca dsl
...
Loading Rails application... 
Tapioca attempted to load the Rails application after encountering a `config/application.rb` file, but it failed. If your application uses Rails please ensure it can be loaded correctly before generating RBIs. If your application does not use Rails and you wish to continue RBI generation please pass `--no-halt-upon-load-error` to the tapioca command in sorbet/tapioca/config.yml or in CLI.
You called sig twice without declaring a method in between
/usr/local/bundle/gems/sorbet-runtime-0.6.13308/lib/types/private/methods/_methods.rb:56:in `declare_sig'
/usr/local/bundle/gems/sorbet-runtime-0.6.13308/lib/types/sig.rb:99:in `sig'
/app/app/models/api_key.rb:23:in `<class:ApiKey>'
/app/app/models/api_key.rb:4:in `<compiled>'
/usr/local/bundle/gems/require-hooks-0.4.0/lib/require-hooks/mode/load_iseq.rb:18:in `eval'
/usr/local/bundle/gems/require-hooks-0.4.0/lib/require-hooks/mode/load_iseq.rb:18:in `block in load_iseq'
/usr/local/bundle/gems/require-hooks-0.4.0/lib/require-hooks/api.rb:50:in `run_around_load_callbacks'
/usr/local/bundle/gems/require-hooks-0.4.0/lib/require-hooks/mode/load_iseq.rb:15:in `load_iseq'
/usr/local/bundle/gems/tapioca-0.19.1/lib/tapioca/rbs/rewriter.rb:29:in `load_iseq'
/usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
/usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
```

**Additional context**

I persisted and got PR working after a lot of effort - this confirmed to me the effort required at times.
* https://github.com/openaustralia/planningalerts/pull/2076

Rails itself runs fine!

Ben had added a workflow to update the rdi files and then had to revert it due to problems, so rdi files have to be updated for each dependabot PR manually.

main branch is fine - its one of these dependabot security updates that triggers the issue:

Image

Br3nda:
## What and why

Adds `.github/workflows/dependabot-tapioca.yml`: on a Dependabot bundler PR, regenerates Sorbet's Tapioca RBI files and pushes the result back, but only once it's confirmed the regenerated RBI still type checks cleanly.

Dependabot bumps `Gemfile.lock` but never runs `tapioca`, so `ci:type` reliably fails on gem-bump PRs with a stale-RBI error unrelated to any real incompatibility - the base branch of this PR, #2139, is a live example. That failure currently goes unnoticed too, since no status check is required to merge.

- Closes #2082 once the manual setup below is done.
- The job itself lives in a new shared, reusable workflow in `openaustralia/pavlovabot` (not yet created - see below), authenticating as a dedicated GitHub App rather than a personal token, so it can be reused as other openaustralia repos adopt Sorbet.
- This re-attempts what #2043 tried and #2048 reverted. The revert reason was that #2043's auto-push used the default `GITHUB_TOKEN`, which GitHub does not let re-trigger other `pull_request` workflows - so `test`/`type_check`/`lint` silently never ran again against the final commit. This version pushes as a separate bot identity instead, so those checks genuinely re-run.
- On a real Sorbet error (not just a stale file), the job fails and pushes nothing, leaving the PR exactly as Dependabot created it for a human to look at.

**Not yet functional** - three manual, one-off steps are still needed before this actually runs, all documented in `openaustralia/pavlovabot`'s README:
1. Create the `openaustralia/pavlovabot` repo (blocked so far by a safety check on creating new public repos).
2. Create and install the PavlovaBot GitHub App under the openaustralia org.
3. Add `PAVLOVABOT_APP_ID` / `PAVLOVABOT_PRIVATE_KEY` as **organisation Dependabot secrets** (not Actions secrets - Dependabot-triggered runs can't see those at all), scoped to this repo.

## Security: can an outside contributor get the secrets?

No. Take someone who opens a PR from a fork, names their branch `dependabot/bundler/anything`, and edits this workflow.

- The job's `if:` requires `github.actor == 'dependabot[bot]'`, an exact match. GitHub sets the actor from the authenticated account, and names ending in `[bot]` are reserved for GitHub Apps, so a person can't take that name.
- The `if:` also requires `head.repo.full_name == github.repository`, which excludes forks.
- The branch-name check is attacker-controlled and gives no protection on its own. It only skips Dependabot's non-bundler PRs.
- GitHub gives fork PR runs on a public repo no secrets at all, whatever the workflow file says.
- The App credentials are Dependabot secrets, which should not reach runs triggered by a person. This is from GitHub's docs and hasn't been tested yet, so confirm it on the first real run.
- The remaining risk is anyone who can merge to `main` here or in `openaustralia/pavlovabot`, because the caller uses `@main`. `pavlovabot` now has a `CODEOWNERS` file, and "Require review from Code Owners" still needs turning on in its branch protection.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Chore
- [ ] Already deployed
- [ ] Breaking change
- [ ] Documentation

## How this was checked

- Both workflow files were validated as syntactically correct YAML.
- Not yet exercised end-to-end against a real Dependabot PR - can't be, until the setup steps above are done.

- [ ] Checked the affected area on my own or a staging system
- [x] Ran the automated tests
- [x] Documentation updated, or not needed - see `openaustralia/pavlovabot`'s README for the full design

Note this PR currently targets `dependabot/bundler/json-2.21.2` (per request), not `main` - worth deciding whether to retarget before merge, depending on what happens to #2139 first.

Assisted-by: Claude Code:claude-sonnet-5


ianheggie-oaf (approved):
Approved - LGTM at a glance

